Issue #8, opened 2023-08-16
Set up publishing flow


matejchalk:
- [ ] use [@jscutlery/semver](https://github.com/jscutlery/semver#readme) and [ngx-deploy-npm](https://github.com/bikecoders/ngx-deploy-npm#readme)
- [ ] use sync mode for all packages
- [x] change NPM package scope to `@code-pushup`
- [ ] automate release with GitHub Actions

Issue #6, opened 2023-08-16
Set up Yargs and basic structure


matejchalk:
Implement:
- yargs in a testable way
- export bin & index
- e2e test config parsing

PR #12, opened 2023-08-22
Packages setup

### LINKED ISSUE #13: Set up package builds and test, config imports

matejchalk:
Closes #13 

Set up `cli`, `plugin-eslint` and `plugin-lighthouse` packages.

**Edit:** Went with [esbuild](https://esbuild.github.io/) (for bundling library) and [bundle-require](https://github.com/egoist/bundle-require#readme) (for importing config files) in the end, and we only support config files in ESM.

## Testing

Make sure to run `npm install`.

### Execute tests with Vitest

```sh
npx nx test cli
````

### Execute built CLI manually

```sh
npx nx run cli:exec ./packages/cli/src/lib/config.mock.js
```

```sh
npx nx run cli:exec ./packages/cli/src/lib/config.mock.mjs
```

```sh
npx nx run cli:exec ./packages/cli/src/lib/config.mock.ts
```


vmasek (approved):

vmasek (dismissed):

esbuild.config.js
```diff
@@ -0,0 +1,44 @@
...
+esbuild.build({
+  plugins: [
+    {
+      name: 'TypeScriptDeclarations',
+      setup(build) {
+        build.onEnd(result => {
+          if (result.errors.length > 0) return;
+
+          execSync(
+            `tsc --emitDeclarationOnly --project packages/${project}/tsconfig.lib.json --outDir dist`,
```
ChristopherPHolder, line 16: I assume you have tested that its the correct output location :)

packages/cli/.eslintrc.json
```diff
@@ -0,0 +1,35 @@
...
+      "parser": "jsonc-eslint-parser",
+      "rules": {
+        "@nx/dependency-checks": [
+          "error",
+          {
+            "ignoredDependencies": [
+              "vite",
+              "@nx/vite",
+              "@quality-metrics/eslint-plugin",
+              "@quality-metrics/lighthouse-plugin"
```
ChristopherPHolder, line 28: Not sure what the purpose of this is. Don't you usually ignore all deps for linting?

packages/cli/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@quality-metrics/cli",
+  "version": "0.0.1",
+  "dependencies": {
+    "bundle-require": "^4.0.1"
```
ChristopherPHolder, line 5: Though this was added automagically by nx

packages/cli/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import { cli } from './lib/cli';
+
+cli(process.argv[process.argv.length - 1]);
```
ChristopherPHolder, line 3: main, and index are 3 different this things. is index being used as a executable ? 

Because it probably should only import and export things.

packages/cli/src/lib/cli.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { execSync } from 'child_process';
+import { cli } from './cli';
+
+describe('cli', () => {
+  beforeAll(() => {
+    // symlink NPM workspaces
+    execSync('npm install');
```
ChristopherPHolder, line 7: Dont get this, why are you syncing the npm in the test

packages/cli/tsconfig.spec.json
```diff
@@ -0,0 +1,19 @@
...
+    "src/**/*.test.ts",
+    "src/**/*.spec.ts",
+    "src/**/*.test.tsx",
+    "src/**/*.spec.tsx",
+    "src/**/*.test.js",
+    "src/**/*.spec.js",
+    "src/**/*.test.jsx",
+    "src/**/*.spec.jsx",
+    "src/**/*.d.ts"
+  ]
```
ChristopherPHolder, line 18: Do mocks need to be included here?

packages/plugin-eslint/tsconfig.spec.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../dist/out-tsc",
+    "types": ["vitest/globals", "vitest/importMeta", "vite/client", "node"]
+  },
+  "include": [
```
ChristopherPHolder, line 7: Should this include mocks ?

packages/plugin-lighthouse/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@quality-metrics/lighthouse-plugin",
+  "version": "0.0.1",
+  "dependencies": {
+    "lighthouse": "^11.0.0"
```
ChristopherPHolder, line 5: Should this not happen automatically by nx

packages/plugin-lighthouse/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "module": "commonjs",
```
ChristopherPHolder, line 4: I guess this should not be commonjs

packages/plugin-lighthouse/tsconfig.lib.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../dist/out-tsc",
+    "declaration": true,
+    "types": ["node"]
+  },
+  "include": ["src/**/*.ts"],
+  "exclude": [
+    "jest.config.ts",
```
ChristopherPHolder, line 10: I guess there is no jest

tools/scripts/publish.mjs
ChristopherPHolder, line 1: are we using our own tool for publishing and versioning? why not use something like the thing Eduard created for nx

ChristopherPHolder (changes requested):
Most of the comments are questions, and the comments repeat themselves a lot.

.eslintrc.json
```diff
@@ -0,0 +1,40 @@
+{
+  "root": true,
+  "ignorePatterns": ["**/*"],
```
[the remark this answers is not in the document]
vmasek: It is correct, you first ignore everything and then in the `overrides` array `files` property specify to what file pattern extensions specific rules apply.

.eslintrc.json
```diff
@@ -0,0 +1,40 @@
...
+  "plugins": ["@nx"],
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+      "rules": {
+        "@nx/enforce-module-boundaries": [
```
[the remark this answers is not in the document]
matejchalk: It's generated by Nx this way. It does already check some things without an additional config, e.g. importing between projects only via path aliases, and no circular dependencies

I would like to add more configuration, so that we also include some tag-based rules, but I think it makes sense to way until @BioPhoton's PR for #6 lands (he needs this to be merged first), since he's also adding more projects (e.g. `core` and `cli`), and we'll also need to separate the e2e project #19 (`cli` imports from `plugin-eslint` and `plugin-lighthouse` in test files, which would break our lint rule).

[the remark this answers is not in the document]
matejchalk: I've created an issue for this here: #28

.eslintrc.json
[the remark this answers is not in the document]
matejchalk: Nx creates it this way. The base ESLint config ignores everything, while each project config overrides to include its own files. This way each project can be linted separately.

I actually didn't touch the ESLint configs at all yet, so this all from Nx generators.

packages/cli/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "module": "commonjs",
```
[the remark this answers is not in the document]
matejchalk: Nx generated it like this by default, but its actually ignored in the `build` because we're using `esbuild` instead of `tsc`.

But I think I'll change these to `ESNext` anyway, otherwise I think TypeScript would complain if we used some ESM-only syntax (like `import.meta.url`).

packages/plugin-eslint/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@quality-metrics/eslint-plugin",
+  "version": "0.0.1",
+  "dependencies": {
+    "eslint": "~8.46.0"
```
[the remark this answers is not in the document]
matejchalk: It's manually added. Nx doesn't add it automatically, but it does check that each imported dependency is listed in the project's `package.json` using the `@nx/dependency-checks` ESLint rule: https://nx.dev/packages/eslint-plugin/documents/dependency-checks

esbuild.config.js
matejchalk: Yes, I repeatedly ran the `build`s, checked the `dist` folder and checked that I get type information when import the built package.

image

image

packages/cli/package.json
matejchalk: As mentioned above, Nx has a lint rule for this now. It seems like they were generating it before (e.g. there's [a deprecated `generatePackageJson` executor option](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/esbuild.impl.ts#L77C5-L80)), but I guess they had a reason for moving away from that.

packages/cli/.eslintrc.json
matejchalk: The rule checks that every imported package is listed in the `package.json`, which is pretty cool. E.g. I would get a lint error if I don't have `bundle-require` in the `cli`'s `package.json`.

However, I had to add some exceptions to make the linter pass:
- Adding the `@quality-metrics/eslint-plugin` and `@quality-metrics/lighthouse-plugin` is temporary and won't be needed once we move out the test files which import them in #19.
- For some reason, the Nx rule gets tripped up by the imports in `vite.config.ts`. There's [a related issue for it in Nx](https://github.com/nrwl/nx/issues/17905) and even [a merged PR](https://github.com/nrwl/nx/pull/17930), but it doesn't seem to be in the latest stable version yet.

packages/cli/src/index.ts
matejchalk: This will be restructured later by @BioPhoton in #6, I only included the bare minimum here for testing the config imports work end-to-end.

packages/cli/src/lib/cli.spec.ts
matejchalk: The config files used in the test have imports from the NPM package names (`@quality-metrics/eslint-plugin` and `@quality-metrics/lighthouse-plugin`). This works via symlinks which NPM creates automatically when installing thanks to the `"workspaces": "dist/packages/*"` in `package.json`.

The problem is that, in the CI job, `npm install` runs before the project `build`s, so the symlinks don't get created because `dist` is empty at that time (less common locally), so [the test fails](https://github.com/flowup/quality-metrics-cli/actions/runs/6011150937/job/16303950914). So I re-run `npm install` here just so that NPM creates the symlinks and the test works.

I'd like address this in a better way once we move out these E2E tests in #19, but for now I just wanted a passing CI and not to do too much in one PR.

packages/plugin-lighthouse/package.json
matejchalk: Addressed above.

tools/scripts/publish.mjs
matejchalk: This was generated by Nx when creating a publishable library (same as the Verdaccio stuff), I didn't modify it. 

I've no idea if we want to use this in the end, but I think it's out of scope for this PR, we can look into customizing the publishing flow later in #8.

packages/plugin-eslint/package.json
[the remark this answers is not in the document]
ChristopherPHolder: Interesting, in the other projects i have i see this being added automatically. Anyway, was just a question :)

packages/cli/.eslintrc.json
ChristopherPHolder: Super cool

BioPhoton (approved):

Issue #33, opened 2023-08-31
Use Verdaccio local registry for E2E tests


matejchalk:
- currently we use NPM workspaces to e2e test configs with plugin imports
  - requires running `npm install` after first package builds
- Nx supports Verdaccio for publishing project's to local registry
- [@nx/plugin:e2e-project](https://nx.dev/packages/plugin/generators/e2e-project) generator sets up local registry setup and teardown for E2E testing
  - it only supports Jest (and CJS because of `require.resolve('nx')`), but we should be able to replace it with [Vitest using `globalSetup` option](https://vitest.dev/config/#globalsetup)
- replace imports in `e2e/cli-e2e/mocks/code-pushup.config.(mjs|js|cjs|ts)`

PR #34, opened 2023-08-31
E2E setup

### LINKED ISSUE #19: E2E tests setup for CLI
The `cli` project's `test` currently relies on `build` of other projects, because it's testing if config files with plugin imports can be loaded. This also relies on running `npm install` after build to create symlinks for NPM workspaces.

Testing this part is valuable, but it would be better to move it to an E2E tests suite, so that unit tests can be build independent.

- [ ] move config loading tests under `e2e` target
- [ ] make CLI unit tests under `test` target independent of `build`
- [ ] use separate `cli-e2e` project

matejchalk:
Closes #19 

- moved E2E tests (loading config with imports from symlinked NPM packages) into separate `cli-e2e` project with an `e2e` target
- project structure inspired by [@nx/plugin:e2e-project](https://nx.dev/packages/plugin/generators/e2e-project) generator
- spent a little time to see if we can use Verdaccio local-registry instead of NPM workspaces, created a follow-up issue in case we want to revisit it: #33


vmasek (approved):

BioPhoton (commented):
Like the workspace layout. 

we could move the e2e config partially as target defaults in `nx.json`

```typescript
, 
    "targetDefaults": {
     // ...
     "e2e": {
      "dependsOn": ["^build"],
      "inputs": ["default", "^production"]
    }
  }
```

vmasek (approved):

PR #39, opened 2023-09-06
Add package `utils` for cli independent logic

BioPhoton:
This PR focuses on the CLI independent logic for the collect command.

It includes:
- source code + JsDocs
- unit tests
- test helper
- mock data
- VFS setup for tests that produce files
- adoptions for libs (make them buildable)


packages/utils/src/lib/collect/implementation/mock/process-helper.mock.ts
```diff
@@ -0,0 +1,76 @@
...
+    runs?: number;
+    outputPath?: string;
+  } = { throwError: false },
+) {
+  const outputPath = cfg?.outputPath || './out-async-runner.json';
+  const args = [asyncProcessPath];
+  cfg?.interval ? args.push(cfg?.interval + '') : args.push('10');
+  cfg?.runs ? args.push(cfg?.runs + '') : args.push('4');
+  cfg?.throwError ? args.push('1') : args.push('0');
+  args.push(outputPath);
```
matejchalk, line 25: I really don't like how the ternary operator is being used here. It's supposed to return different expressions based on a condition (FP-style immutable logic). But you're using it here to conditionally execute statements, you're not assigning the result because the point is the side-effect of mutating the array. An `if`/`else` would fit better with such an imperative approach.

But this could also be easily refactored to use `? :` in a more semantic way:

```suggestion
  const args = [asyncProcessPath];
  args.push(cfg?.interval ? cfg.interval + '' : '10');
  args.push(cfg?.runs ? cfg.runs + '' : '4');
  args.push(cfg?.throwError ? '1' : '0');
  args.push(outputPath);
```

And can in fact be further simplified into a completely immutable version:

```suggestion
  const args = [
    asyncProcessPath,
    cfg?.interval ? cfg.interval + '' : '10',
    cfg?.runs ? cfg.runs + '' : '4',
    cfg?.throwError ? '1' : '0',
    outputPath,
  ];
```

packages/utils/src/lib/collect/index.ts
```diff
@@ -0,0 +1,110 @@
...
+ *   console.log('Valid config:', validationResult.data);
+ * } else {
+ *   console.error('Invalid config:', validationResult.error);
+ * }
+ *
+ */
+export const runAndCollectOutputSchema = runnerOutputSchema.merge(
+  z.object(
+    {
+      date: z.string({ description: 'Date of the run' }),
```
matejchalk, line 74: It wasn't clear to me until I read this implementation that the `date` is when it finished, not when it started. It would be helpful to specify this in the description:

```suggestion
      date: z.string({ description: 'Date and time when the run finished' }),
```

packages/utils/src/lib/collect/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    // @TODO better error message e.g. `Plugin ${cfg.slug} produced wrong output under ${processOutputPath}.` + zod trace
+    const processResult = runAndCollectOutputSchema.parse({
+      ...runnerOutputs,
+      date: new Date().toISOString(),
+      duration: performance.measure('startExecutePlugins', 'stopExecutePlugins')
+        .duration,
+    });
+    return Promise.resolve(processResult);
+  } else {
+    return Promise.reject(new Error('No plugins registered'));
```
matejchalk, line 108: When the function is declared as `async`, then the return value and errors are automatically wrapped in promises, so it's unnecessary to wrap them yourself:

```suggestion
    return processResult;
  } else {
    throw new Error('No plugins registered');
```

matejchalk, line 108: To avoid nesting code, it's generally a good idea to use an early `return` or `throw` for the edge case at the top of the function:

```ts
  if (!plugins?.length) {
    throw new Error('No plugins registered');
  }

  // ... main code no longer need be nested in an if/else-block ...
```

packages/models/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@quality-metrics/models",
+  "version": "0.0.1",
+  "dependencies": {
+    "bundle-require": "^4.0.1"
```
matejchalk, line 5: The dependencies are inaccurate, `bundle-require` isn't used here, but `zod` is for example.

packages/utils/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@quality-metrics/utils",
+  "version": "0.0.1",
+  "dependencies": {
+    "bundle-require": "^4.0.1"
```
matejchalk, line 5: Same issue as for `models` package, these aren't the correct dependencies.

packages/utils/src/lib/collect/implementation/execute-plugin.ts
```diff
@@ -0,0 +1,113 @@
...
+ * console.log(output.stdout); // hello world
+ * console.log(output.stderr); // ''
+ * console.log(output.code); // 0
+ *
+ * @example
+ * // error handling
+ * try {
+ * await executeProcess(cfg);
+ * } catch (e) {
+ * console.log(e.message); // Plugin output is invalid
```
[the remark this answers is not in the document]
BioPhoton: fixed

packages/utils/src/lib/collect/implementation/execute-plugin.ts
```diff
@@ -0,0 +1,113 @@
...
+export async function executePlugins(
+  plugins: PluginConfigSchema[],
+): Promise<RunnerOutputSchema> {
+  return await plugins.reduce(async (acc, pluginCfg) => {
+    const outputs = await acc;
+    const runnerOutput = await executePlugin(pluginCfg);
+    outputs.audits.concat(runnerOutput.audits);
+    return outputs;
+  }, Promise.resolve({ audits: [] } as RunnerOutputSchema));
+}
```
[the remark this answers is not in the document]
BioPhoton: I did it on purpose for now as we don't have a way to decide is the plugin can run in parallel to others. lighthouse for example should not run next to others. Any static analysis can.

packages/utils/src/lib/collect/index.ts
BioPhoton: I made it the start date. this is more align with other time series data like a e.g. network request

packages/utils/src/lib/collect/index.ts
BioPhoton: thx! nice one!

packages/models/package.json
BioPhoton: I guess we can use the build process to generate the deps automatically... Let me try.

packages/utils/src/lib/collect/implementation/execute-plugin.ts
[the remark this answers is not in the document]
matejchalk: Ah yes, good point. OK, then let's keep it always sequential for now :+1:

packages/utils/src/lib/collect/implementation/execute-process.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+} from './mock/process-helper.mock';
+import { join } from 'path';
+
+const outFolder = '';
+const outName = 'out-async-runner.json';
+const outputPath = join(outFolder, outName);
+
+describe('executeProcess', () => {
+  it('should work with shell command `ls`', async () => {
+    const cfg = mockProcessConfig({ command: `ls`, args: ['-a'] });
```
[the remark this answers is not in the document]
BioPhoton: fixed

packages/models/package.json
BioPhoton: Hmmm the [linting rule](https://nx.dev/packages/eslint-plugin/documents/dependency-checks) does not work on my end...

packages/utils/package.json
BioPhoton: fixed

packages/utils/src/lib/collect/implementation/mock/process-helper.mock.ts
BioPhoton: Thx! way cleaner now.

packages/models/package.json
matejchalk: Yeah, the Nx rule should catch it, I believe. But it has to be in each project's `.eslintrc.json` file, so the fix should be to add that rule to `models` and `utils` projects (Nx only generates it there when you set it to publishable).

packages/utils/src/lib/collect/index.ts
matejchalk: Agreed, start date is more intuitive for me also :+1:

packages/models/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@quality-metrics/models",
+  "version": "0.0.1",
+  "type": "module",
+  "main": "./index.js",
+  "types": "./src/index.d.ts",
```
matejchalk, line 6: These shouldn't be necesary, the build automatically overwrites them anyway (see [esbuild.config.js](https://github.com/flowup/quality-metrics-cli/blob/ae63670d5b87ca5d80b627ccda481e835e7d3ef2/esbuild.config.js#L32-L34)).

```suggestion
```

packages/cli/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@quality-metrics/cli",
   "version": "0.0.1",
+  "type": "module",
```
matejchalk, line 4: Also not necessary (see above):

```suggestion
```

matejchalk (dismissed):

matejchalk (approved):

PR #46, opened 2023-09-08
refactor(chore): improvements testing

BioPhoton:
This PR includes:
- a refactoring for the test helper to mock schemas
- updated imports
- adoptions to the schemas (removes `label`)


packages/models/src/lib/plugin-config.ts
```diff
@@ -183,7 +162,7 @@ const sourceFileLocationSchema = z.object(
 export const issueSchema = z.object(
   {
     message: z.string({ description: 'Descriptive error message' }).max(128),
-    severity: z.enum(['info', 'warning', 'error'], {
+    severity: z.enum(['success', 'warning', 'error', 'info'], {
```
matejchalk, line 165: Changing the `severity` enum is out of scope for this PR, I think.

I'm not sure it makes sense to add `'success'` here anyway, because it diverges from how I intended this to be applied. The severity is inspired by typically used log levels (so extending it to include `'debug'` or `'critical'` would be more inline, but I didn't want to add too much too soon), because it can also apply to ESLint rule levels (`'warn'` or `'error'`), [NPM audit vulnerability levels](https://docs.npmjs.com/cli/v8/commands/npm-audit#audit-level), etc. There's a clear hierarchy with such levels, so setting a minimum severity (e.g. "I only care about warnings and errors, but nothing below that, so I'll set my minimum severity to 'warning'"). I'd struggle to place a "success severity" in such a hierarchy.

tsconfig.base.json
```diff
@@ -23,6 +23,7 @@
         "packages/plugin-lighthouse/src/index.ts"
       ],
       "@quality-metrics/models": ["packages/models/src/index.ts"],
+      "@quality-metrics/models/testing": ["packages/models/test/index.ts"],
```
matejchalk, line 26: I really like this separation :+1: 

Shouldn't we also exclude the `test` directory from `tsconfig.lib.json`?

packages/models/src/lib/plugin-config.ts
BioPhoton: ok! lets remove it

tsconfig.base.json
BioPhoton: done

matejchalk (approved):

PR #47, opened 2023-09-09
feat(cli): persist login and formatting options

### LINKED ISSUE #20: Implement collect command
- [x] runs runner for each plugin
- [x] reads outputs for each plugin and validates them
- [ ] prints summary to terminal

BioPhoton:
This PR implements:
- a format option on the collect logic
- a `persistReport` function
- formatters for the following report formats:
  - `stdout`
  - `json`
  - `md` 
    - helpers to transform data to markdown 
 - nearly all tests
 - a clumsy way to mock `console.log`'s

to discuss :
- We should probably consider using header for Plugins and audit details and then place results in tables ?
- and probably even the details.issues (for ESLint audits - at least a summary of how many errors, warnings, etc. were found).

closes #20


ChristopherPHolder:
@BioPhoton Does this run locally?

I get typing errors when trying it locally.

packages/utils/src/lib/collect/index.ts
```diff
@@ -76,3 +83,29 @@ export async function collect(options: CollectOptions): Promise<Report> {
     }),
   };
 }
+
+export class ReadPackageJsonError extends Error {
+  constructor(outputPath: string) {
+    super(`outPath: ${outputPath} is no directory`);
+  }
```
matejchalk, line 90: This is copy-pasted from elsewhere, it seems, doesn't fit here. I guess it should be something like this:

```suggestion
  constructor(message: string) {
    super(`error reading package.json: ${message}`);
  }
```

matejchalk (changes requested):

packages/utils/docs/report.md
```diff
@@ -0,0 +1,93 @@
...
+  </details>
+- <details>
+  <summary>Audit 2</summary>
+  </details>
+
+<!-- STOP: Details Section -->
+
+<!-- START: Footer Section -->
+
+Code Pushup Cloud ID: [123abc456def]()
```
matejchalk, line 91: I would skip this for now, it'll be some quite time before we can support such a flow, I think.

packages/utils/docs/report.md
```diff
@@ -0,0 +1,93 @@
...
+<!-- START: Details Section -->
+
+**Performance: 60** (3/4)
+
+- <details>
+  <summary>Largest Contentful Paint</summary>
+  Largest Contentful Paint (LCP) is an important, stable Core Web Vital metric for measuring [perceived load speed](https://web.dev/user-centric-performance-metrics/#types-of-metrics) because it marks the point in the page load timeline when the page's main content has likely loaded—a fast LCP helps reassure the user that the page is useful.
+  </details>
+- <details>
+  <summary>Cumulative Layout Shift</summary>
```
matejchalk, line 53: I'm missing the actual results here. The `description` is a nice-to-have for me (we could also just add a link with the `docsUrl` instead), the main is the `formattedValue` (mainly for Lighthouse audits), and probably even the `details.issues` (for ESLint audits - at least a summary of how many errors, warnings, etc. were found).

packages/utils/src/lib/collect/implementation/persist.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    /* expect(console.log).toHaveBeenCalledTimes(configReportLogNum);
+     expect(
+       logs.find(msg => msg.match(/(erformance)*(11)*(EO)*(validators))),
+     ).toBeTruthy(); */
+  });
+
+  it('should persist some formats`', async () => {
+    await persistReport(dummyReport, {
+      ...dummyConfig,
+      persist: mockPersistConfig({ outputPath, format: ['md', 'stdout'] }),
```
matejchalk, line 137: I'm wondering whether we should allow omitting `json` output :thinking:  The `md` and `stdout` outputs are only for the user, but for uploading to the portal we need the `json` output.

It would be quite unpleasant to run a long `collect` command, only to find out when you run `upload` that you should've specified the `json` format. In general, I would say that when both `upload` config is specified and `persist.format` doesn't include `'json'`, that's almost certainly a mistake. We should either warn the user about it as soon as possible, or just create the `json` file regardless.

packages/utils/docs/report.md
```diff
@@ -0,0 +1,93 @@
...
+| ----------- | ---- | ---------- |
+| 60          | 34   | 80         |
+
+<!-- Overview Version 2 -->
+
+| Category    | Score | Audits |
+| ----------- | ----- | ------ |
+| Performance | 60    | (3/4)  |
+| A11y        | 34    | (7/24) |
+| NxAdoption  | 80    | (7/12) |
```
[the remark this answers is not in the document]
ChristopherPHolder: I do not get how you can concatinate the NxAdo... report and the LH report. I say this because of the multiple steps and flow. 

It seems like we will need to generate a different one for each Auditor/Plugin (eslint, nx-valiator, user-flow). 

Additionally the question we should be asking is about headers, column and row headers. 

NxApdoption will probably end up worth and Categories too. 

As we got essentially unlimited vertical space then I guess we need to go in this direction, 

I also would strongly suggest we do what we did with user-flow where we just reduce the report and convert that into a table.

packages/utils/docs/report.md
```diff
@@ -0,0 +1,93 @@
...
+---
+
+<!-- STOP: Overview Section -->
+
+<!-- START: Details Section -->
+
+**Performance: 60** (3/4)
+
+- <details>
+  <summary>Largest Contentful Paint</summary>
```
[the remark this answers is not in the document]
ChristopherPHolder: We should probably consider using header for Plugins and audit details and then place results in tables ?

packages/models/src/lib/report.ts
```diff
@@ -1,20 +1,129 @@
...
+ *
+ * // Validate the data against the schema
+ * const validationResult = runnerOutputSchema.safeParse(runnerOutputData);
+ *
+ * if (validationResult.success) {
+ *   console.log('Valid runner output:', validationResult.data);
+ * } else {
+ *   console.error('Invalid runner output:', validationResult.error);
+ * }
+ */
```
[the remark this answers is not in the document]
BioPhoton: yes agree!

packages/utils/docs/report.md
```diff
@@ -0,0 +1,93 @@
+# Code Pushup Report
+
+<!-- START: Meta Section -->
+
+---
+
+_Version: 0.0.1_  
+_Date: 9/11/2023 12:00:00am (430ms)_
```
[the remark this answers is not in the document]
BioPhoton: ok! I honestly have no expertise in date strings :D

packages/utils/src/lib/collect/implementation/persist.ts
```diff
@@ -0,0 +1,60 @@
...
+  const { persist } = config;
+  const outputPath = persist.outputPath;
+  let { format } = persist;
+  format = format && format.length !== 0 ? format : ['stdout'];
+
+  if (!existsSync(outputPath)) {
+    try {
+      mkdirSync(outputPath, { recursive: true });
+    } catch (e) {
+      throw new PersistDirError(outputPath);
```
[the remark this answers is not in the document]
BioPhoton: I will also add the stack to the new error! This is a good catch!

packages/utils/src/lib/collect/implementation/report.ts
```diff
@@ -0,0 +1,8 @@
+import { CategoryConfig } from '@quality-metrics/models';
+
+export const reportHeadlineText = 'Code Pushup Report';
```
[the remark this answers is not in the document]
BioPhoton: I would say we clarify this with Albert in slack?
WDYS?

packages/utils/src/lib/collect/implementation/persist.spec.ts
BioPhoton: > The md and stdout outputs are only for the user
The MD is also important for CI pipeline actions and the automated comments in PR's

Yes warning and also Ill see if I can just automatically create it in cases it makes sense

packages/models/src/lib/report.ts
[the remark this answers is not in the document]
BioPhoton: fixed

packages/utils/docs/report.md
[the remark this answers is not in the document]
BioPhoton: done

packages/utils/docs/report.md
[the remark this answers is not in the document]
BioPhoton: plugin name is visible

packages/utils/src/lib/collect/implementation/md/list.ts
```diff
@@ -0,0 +1,11 @@
+export type Order = 'unordered' | 'checkbox';
+
+/**
+ * \# {text} // hierarchy set to 1
+ *
+ * \## {text} // hierarchy set to 2
+ */
```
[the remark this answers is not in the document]
BioPhoton: fixed

packages/utils/src/lib/collect/implementation/persist.ts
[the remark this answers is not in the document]
BioPhoton: done

packages/utils/docs/report.md
BioPhoton: lets focus on the `details.issues` when we refined the data structure

packages/utils/src/lib/collect/implementation/persist.spec.ts
BioPhoton: So now it will get always generated

packages/utils/src/lib/collect/implementation/md/font-style.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * **{text}** // default is bold
+ *
+ * *{text}* // italic - styles set to `['i']`
+ *
+ * ~**{text}**~ // bold & stroke-through - styles set to `['b','s']`
+ */
+export function style(text: string, styles: FontStyle[] = ['b']): string {
+  return styles.reduce((t, s) => `${stylesMap[s]}${t}${stylesMap[s]}`, text);
+}
```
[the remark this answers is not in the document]
BioPhoton: fixed

packages/utils/src/lib/collect/implementation/report.ts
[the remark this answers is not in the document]
matejchalk: I like `Code PushUp` best. We can raise it on Slack on Monday, though.

packages/cli/project.json
```diff
@@ -2,6 +2,7 @@
   "name": "cli",
   "$schema": "../../node_modules/nx/schemas/project-schema.json",
   "sourceRoot": "packages/cli/src",
+  "bin": "src/cli",
```
matejchalk, line 5: Does this do anything? I can't find `bin` anywhere in the [`project.json` type definition](https://github.com/nrwl/nx/blob/master/packages/nx/src/config/workspace-json-project-json.ts#L42-L101).

packages/models/src/lib/report.ts
```diff
@@ -1,20 +1,108 @@
...
+/**
+ * Define Zod schema for the RunnerOutput type.
+ */
+export const runnerOutputSchema = z.object(
+  {
+    audits: z
+      .array(auditResultSchema, { description: 'List of audits' })
+      // audit slugs are unique
+      .refine(
+        audits => !getDuplicateSlugsInAudits(audits as never), //@TODO use hackfix types instead of never same as in plugin config
```
matejchalk, line 95: Would be great to get rid of these `never`s :pray:

packages/utils/src/lib/collect/implementation/md/font-style.ts
```diff
@@ -0,0 +1,18 @@
...
+  b: '**', // bold
+  s: '~', // strike through
+} as const;
+
+export type FontStyle = keyof typeof stylesMap;
+
+/**
+ * **{text}** // default is bold
+ *
+ * *{text}* // italic - styles set to `['i']`
```
matejchalk, line 12:
```suggestion
 * _{text}_ // italic - styles set to `['i']`
```

packages/utils/src/lib/collect/implementation/md/list.ts
```diff
@@ -0,0 +1,9 @@
+export type Order = 'unordered' | 'checkbox';
+
+/**
+ * - {text}
+ */
+export function li(text: string, order: Order = 'unordered'): string {
+  const style = order === 'unordered' ? '-' : '- [ ]';
+  return `${style}  ${text}`;
```
matejchalk, line 8: Produces unnecessary extra space:

```suggestion
  return `${style} ${text}`;
```

packages/utils/src/lib/collect/implementation/md/list.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+  it('should handle empty text with checkbox order', () => {
+    const result = li('', 'checkbox');
+    expect(result).toBe('- [ ]  ');
+  });
+
+  it('should correctly handle special characters', () => {
+    const result = li('Hello *World*', 'checkbox');
+    expect(result).toBe('- [ ]  Hello *World*');
+  });
+});
```
matejchalk, line 28: (See above.)

```suggestion
describe('li', () => {
  it('should return an unordered list item by default', () => {
    const result = li('Hello World');
    expect(result).toBe('- Hello World');
  });

  it('should return a checkbox list item when order set to "checkbox"', () => {
    const result = li('Hello World', 'checkbox');
    expect(result).toBe('- [ ] Hello World');
  });

  it('should handle empty text', () => {
    const result = li('');
    expect(result).toBe('- ');
  });

  it('should handle empty text with checkbox order', () => {
    const result = li('', 'checkbox');
    expect(result).toBe('- [ ] ');
  });

  it('should correctly handle special characters', () => {
    const result = li('Hello *World*', 'checkbox');
    expect(result).toBe('- [ ] Hello *World*');
  });
});
```

packages/cli/project.json
BioPhoton: I found a description in npm. 

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin

You even helped me back then to fix a path problem with it in user-flow CLI :D

matejchalk: Yeah, but that's for the `package.json`, not `project.json` :grin:

packages/models/src/lib/report.ts
BioPhoton: done

packages/cli/project.json
BioPhoton: oh my!!! haha.
Good you saw it!

fixed

matejchalk (approved):

PR #51, opened 2023-09-17
feat(nx-plugin): initial setup and init command

BioPhoton:
This PR includes:
- initial setup
- init command implementation, basic tests init generator

Related to #9


BioPhoton:
@matejchalk 
in the `package.json`of `@quality-metrics/nx-plugin` soe packages like `vitest` or `@nx/vitest` are required by the linting rule, but it is not used in the code base... 

So far I could not understand where this problem is coming from.

Another recurring problem is the package `@nx/nx-darwin-x64` is missing and I could not figure out the exact issue... multiple times deleting and installing solves it but... magic..

Imho it should not even be used on a Mac...

Maybe we can discuss it in the daily.

BioPhoton:
ready for review @matejchalk

vmasek (approved):

PR #53, opened 2023-09-18
refactor(cli): add entry point for bin

ChristopherPHolder:
This MR addresses the issue from MR #49

image

### DIFF
--- a/packages/cli/project.json
+++ b/packages/cli/project.json
@@ -10,9 +10,8 @@
       "options": {
         "outputPath": "dist/packages/cli",
         "main": "packages/cli/src/index.ts",
-        "additionalEntryPoints": ["packages/cli/src/bin.ts"],
         "tsConfig": "packages/cli/tsconfig.lib.json",
-        "assets": ["packages/cli/*.md"],
+        "assets": ["packages/cli/*.md", "packages/cli/src/bin.js"],
         "esbuildConfig": "esbuild.config.js"
       }
     },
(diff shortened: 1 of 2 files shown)


BioPhoton:
@ChristopherPHolder thx for having a look at this!

Did you try to execute the build artefact?
`npx dist/packages/cli collect`

BioPhoton (changes requested):
Would a combination of bundle false + the additional entry point work for us?

ChristopherPHolder:
> Would a combination of bundle false + the additional entry point work for us?

It is worth testing it.

matejchalk:
I had to change the `bin.js` path, but otherwise this seems to work just fine :slightly_smiling_face: 

image

BioPhoton:
Here another 🧩 of the puzzle:
https://github.com/favware/esbuild-plugin-file-path-extensions

ChristopherPHolder:
So it seems it is really not straight forward. Just to getting running for the moment i have converted the bin to a .js file and passed as an asset. 

I know this is not optimal and i am looking into the configuration for getting the bundling correct without the issue of import file extensions.

ChristopherPHolder:
Relevent: https://github.com/evanw/esbuild/issues/622

matejchalk:
I'm OK with the `bin.js` asset as a workaround :man_shrugging: 

Although as mentioned in #73, maybe we don't need two entry points for the `cli` anyway, `index.ts` could just execute instead of only exporting a function. I'm not sure we need to support importing from `cli`, users that want programmatic usage should be able to import everything they need from `core` (or `models`, `utils`, etc.).

BioPhoton:
I'm in for the core package!
lets for now have only the `bin.ts` file present and merge the discussed approach for the cli package.
I'll setup a core package later.

BioPhoton:
closed as implemented in #66

PR #54, opened 2023-09-18
build: fix dependency-checks rule - ignore vite.config.ts and *.mocks.ts

matejchalk:
Workaround with `ignoredDependencies` for `@nx/dependency-checks` ESLint rule is now no longer needed :slightly_smiling_face:


vmasek (approved):

BioPhoton (commented):
lgtm

PR #58, opened 2023-09-22
test: reduce unit test flakiness

### LINKED ISSUE #71: Reduce unit test flakiness - part 1
- [x] fix reading `package.json` error
- [x] fix reliance on test execution order
- [x] don't pollute repo with files created in tests

Follow-up:
- #72

matejchalk:
Closes #71 

Unit tests for `cli` and `utils` packages were repeatedly failing locally, and some tests were even failing or passing seemingly at random. So I identified some sources of flakiness which I was able to eliminate to make unit tests pass repeatedly, even when run in parallel.

- reading `package.json` via `fs` made it unreliable because of relative path mismatch between `src` and `dist`
  - replaced with a static `import`
- the same `out-execute-plugin.json` file was being created by many different tests, so some test results depended on which test ran before
  - randomized the file name so that each test works with a different file
  - also moved files into `.gitignore`d `tmp` directory so that tests don't generate untracked files
- test results also differed based on if project `test` target was run in isolation or in parallel with another unit test
  - ensured `tmp` directory is cleaned out in global setup
- mock Lighthouse data contained a `title` that exceeded 128 characters
  - increased limit to 256 characters

Unit tests now pass on my machine (Linux). Let me know if you still have problems on others machines (Windows or Mac) :pray: 

image

Next steps:
- #72

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -4,7 +4,6 @@
   "dependencies": {
     "@quality-metrics/models": "^0.0.1",
     "chalk": "^5.3.0",
-    "cliui": "^8.0.1",
-    "type-fest": "^4.3.1"
+    "cliui": "^8.0.1"
   }
 }
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -2,8 +2,9 @@
   "name": "@quality-metrics/utils",
   "version": "0.0.1",
   "dependencies": {
+    "@quality-metrics/models": "^0.0.1",
     "chalk": "^5.3.0",
     "cliui": "^8.0.1",
-    "@quality-metrics/models": "^0.0.1"
+    "type-fest": "^4.3.1"
   }
 }
--- a/packages/utils/src/lib/collect/index.ts
+++ b/packages/utils/src/lib/collect/index.ts
@@ -1,6 +1,7 @@
 import { CoreConfig, GlobalCliArgs, Report } from '@quality-metrics/models';
+import type { PackageJson } from 'type-fest';
 import { executePlugins } from './implementation/execute-plugin';
-import { calcDuration, readPackageJson } from './implementation/utils';
+import { calcDuration } from './implementation/utils';
 
 /**
  * Error thrown when collect output is invalid.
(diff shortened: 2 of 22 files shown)


vmasek:
would it make sense to prefix by run timestamp instead of random number?

packages/utils/package.json
```diff
@@ -2,8 +2,9 @@
   "name": "@quality-metrics/utils",
   "version": "0.0.1",
   "dependencies": {
+    "@quality-metrics/models": "^0.0.1",
     "chalk": "^5.3.0",
     "cliui": "^8.0.1",
-    "@quality-metrics/models": "^0.0.1"
+    "type-fest": "^4.3.1"
```
BioPhoton, line 8: Is it worth to pull in a dependency for typing the `package.json`? I don't really use this lib so I just ask...

packages/utils/src/lib/collect/index.ts
```diff
@@ -28,8 +28,10 @@ export type CollectOptions = GlobalCliArgs & CoreConfig;
  * Run audits, collect plugin output and aggregate it into a JSON object
  * @param options
  */
-export async function collect(options: CollectOptions): Promise<Report> {
-  const { version, name } = await readPackageJson();
+export async function collect(
```
BioPhoton, line 31: The package version is always fixed. We maybe always take the version from utils package and don't expose it to options?

I'm afraid if we expose it it get misused...

BioPhoton (changes requested):
Nice improvements to the tests! 🧩  
Left some questions regarding package versions.

packages/utils/src/lib/collect/index.ts
matejchalk: Well, we want to use package metadata for the `cli` package, so if `utils` was in charge of importing it, then `utils` would depend on `cli`, which is a big no-no for the dependency graph :grimacing: 

The only reasonable alternative I can think of is that we remove the `package.json` logic from `utils` completely. This `collect` function would the report without the package metadata (it's optional, so our types will allow it), and the CLI command adds it there afterwards. Would do you think? :thinking:

packages/utils/package.json
matejchalk: Probably not, since we only need `{ name?: string, version?: string }`. I'll remove it.

packages/utils/src/lib/collect/index.ts
BioPhoton: > and the CLI command adds it there afterwards
This sounds like a good option.

vmasek (dismissed):

BioPhoton (approved):

vmasek (approved):

PR #74, opened 2023-09-27
test: unit test improvements

### LINKED ISSUE #72: Reduce unit test flakiness - part 2
- [x] remove NPM workspaces to make tests indepedent of build
- [x] use `await expect(() => ...).rejects.toThrowError()` instead of error spies with `.catch` (improves test failure messages)
- [x] use `node -e "require('fs').writeFileSync(..., ...)"` instead of `bash -c echo ... > ...` (platform-indepedence)
- [x] use `memfs` for testing persist logic (see [`persist.spec.ts`](https://github.com/flowup/quality-metrics-cli/blob/main/packages/utils/src/lib/collect/implementation/persist.spec.ts)) with different formats (instead of real file-system) - see [`vitest-memfs-poc` branch](https://github.com/flowup/quality-metrics-cli/compare/main...vitest-memfs-poc)
- [ ] mock `process.exit` in Yargs tests

### Refactor: Platform-independant plugin runner

Image

### Refactor: Testing errors

Image

matejchalk:
Closes #72 

- no NPM workspaces
  - E2E tests import from `dist` for now, should be solved properly by #33  
- using `.rejects.toThrowError()` instead of `.catch`
- using `node` (prerequisite) instead of `bash` (relies on UNIX) to create test files
- using `memfs` instead of actual file system for some tests 

Unfortunately, I wasn't able to come up with a way to mock `process.exit` in `cli` tests (`vi.spyOn(process, 'exit')` and other variants had no effect for some reason :confused:).

### DIFF
--- a/examples/cli-e2e/mocks/config.mock.js
+++ b/examples/cli-e2e/mocks/config.mock.js
@@ -1,5 +1,6 @@
-import eslintPlugin from '@quality-metrics/eslint-plugin';
-import lighthousePlugin from '@quality-metrics/lighthouse-plugin';
+// TODO: import plugins using NPM package names using local registry: https://github.com/flowup/quality-metrics-cli/issues/33
+import eslintPlugin from '../../../dist/packages/plugin-eslint';
+import lighthousePlugin from '../../../dist/packages/plugin-lighthouse';
 
 export default {
   persist: { outputPath: 'cli-config-out.json' },
--- a/examples/cli-e2e/mocks/config.mock.js
+++ b/examples/cli-e2e/mocks/config.mock.js
@@ -3,7 +3,7 @@ import eslintPlugin from '../../../dist/packages/plugin-eslint';
 import lighthousePlugin from '../../../dist/packages/plugin-lighthouse';
 
 export default {
-  persist: { outputPath: 'cli-config-out.json' },
+  persist: { outputPath: 'tmp/cli-config-out.json' },
   categories: [],
   plugins: [
     eslintPlugin({ config: '.eslintrc.json' }),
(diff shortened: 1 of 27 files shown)


ChristopherPHolder (changes requested):
This is great, really really love it and now i only got one failing unit test. (not suggesting we deal with that in this MR). 

There is one test that probably is related to `using node (prerequisite) instead of bash (relies on UNIX) to create test files` but also seems to not be required at all. 

`utils/src/lib/collect/implementation/execute-process.spec.ts`

```ts
  if (os.platform() !== 'win32') {
    it('should work with shell command `ls`', async () => {
      const cfg = mockProcessConfig({ command: `ls`, args: ['-a'] });
      const { observer } = cfg;
      const processResult = await executeProcess(cfg);
      expect(observer?.next).toHaveBeenCalledTimes(1);
      expect(observer?.complete).toHaveBeenCalledTimes(1);
      expect(processResult.stdout).toContain('..');
    });
  }
```

Should this test be removed? Or do you see a value to having it there?

matejchalk:
> There is one test that probably is related to `using node (prerequisite) instead of bash (relies on UNIX) to create test files` but also seems to not be required at all.
> 
> `utils/src/lib/collect/implementation/execute-process.spec.ts`
> 
> ```ts
>   if (os.platform() !== 'win32') {
>     it('should work with shell command `ls`', async () => {
>       const cfg = mockProcessConfig({ command: `ls`, args: ['-a'] });
>       const { observer } = cfg;
>       const processResult = await executeProcess(cfg);
>       expect(observer?.next).toHaveBeenCalledTimes(1);
>       expect(observer?.complete).toHaveBeenCalledTimes(1);
>       expect(processResult.stdout).toContain('..');
>     });
>   }
> ```
> 
> Should this test be removed? Or do you see a value to having it there?

I don't think it's necessary really, there are already tests for other commands like `node -v` or `npx --help`, which don't rely on a UNIX environment. I've removed it now.

ChristopherPHolder (approved):

PR #81, opened 2023-10-02
feat(utils): move parts from utils into core

### LINKED ISSUE #78: refactor(utils): prepare for core refactoring
This is a suggestion on how the new files should get arranged:

**utils:**  
- md
- execute-process.ts
- report-to-md.ts
- report-to-stdout.ts
- utils.ts

**core:**  
- execute-plugin.ts
- persist.ts

BioPhoton:
closes #78


packages/cli/package.json
```diff
@@ -9,7 +9,7 @@
     "chalk": "^5.3.0",
     "yargs": "^17.7.2",
     "@quality-metrics/models": "^0.0.1",
-    "@quality-metrics/utils": "^0.0.1",
-    "zod": "^3.22.1"
+    "zod": "^3.22.1",
+    "@quality-metrics-cli/core": "*"
```
matejchalk, line 13: We've been using `^0.0.1` for our package deps, not `*`. I guess restricting it to major versions is safer, but either way we should be consistent.

packages/cli/src/lib/implementation/collect-and-persist.ts
```diff
@@ -29,7 +28,7 @@ export async function collectAndPersistReports(
       // Running checks after persisting helps while debugging as you can check the invalid output after the error
       pluginOutputSchema.parse(plugin);
     } catch (e) {
-      throw new CollectOutputError(plugin.slug, e as Error);
+      throw new Error(plugin.slug + (e as Error).message);
```
matejchalk, line 31: There's no space between slug and message, so it wouldn't be very readable. I recommend using template strings instead of concatenation, it's easier to spot such issues this way:

```suggestion
      throw new Error(`${plugin.slug} - ${(e as Error).message}`);
```

packages/core/package.json
```diff
@@ -3,5 +3,10 @@
   "version": "0.0.1",
   "dependencies": {},
   "type": "commonjs",
-  "main": "./index.cjs"
+  "main": "./index.cjs",
```
matejchalk, line 6: I guess I missed this before... Why is this a CJS package, when all our other packages are ESM? :open_mouth: 

I believe the `build` and `package.json` configuration should be exactly the same for `core` as for our other packages.

packages/core/package.json
```diff
@@ -3,5 +3,10 @@
...
+  "peerDependencies": {
+    "chalk": "5.3.0",
+    "@quality-metrics/models": "latest",
+    "@quality-metrics/utils": "latest"
```
matejchalk, line 10: Again, we need to align on how we reference versions between our packages. I'm not sure `latest` and `*` are very future proof.

.eslintrc.json
```diff
@@ -26,11 +26,6 @@
       "extends": ["plugin:@nx/typescript"],
       "rules": {}
     },
-    {
```
matejchalk, line 29: Why remove this? I don't think there's a downside to keeping the Nx preset for `.js` files (it'll only run if the glob matches anyway).

matejchalk (changes requested):

.eslintrc.json
BioPhoton: added .. guess it was due to some issues

packages/core/package.json
BioPhoton: will create a ticket for tooling

PR #83, opened 2023-10-02
fix: refactor after core package intro

BioPhoton:


packages/core/src/lib/collect-and-persist.ts
```diff
@@ -1,21 +1,17 @@
-import {
-  collect,
-  CollectOptions,
-  logPersistedResults,
-  persistReport,
-} from '@quality-metrics/core';
+import { collect, CollectOptions } from './collect';
+import { name, version } from '../../package.json';
```
matejchalk, line 2: This will mean the package name and version in our reports will be for `@code-pushup/core`, not `@code-pushup/cli`. I can't decide if that's better or worse, honestly, so happy to keep it as is, just thought I'd flag it to make sure it's intentional :man_shrugging:

packages/utils/src/index.ts
```diff
@@ -9,3 +9,4 @@ export {
 export { calcDuration, formatBytes } from './lib/collect/implementation/utils';
 export { reportToStdout } from './lib/collect/implementation/report-to-stdout';
 export { reportToMd } from './lib/collect/implementation/report-to-md';
+export { importModule } from './lib/collect/implementation/load-file';
```
matejchalk, line 12: Totally fine with this not being addressed in this PR to reduce scope, but I believe `utils` shoudn't have a `collect/implementation` folder in the end. That folder structure makes sense within `core`, but `utils` shouldn't really be aware of CLI commands like `collect`. The function themselves (`formatBytes`, `reportToMd`, `importModule`, etc.) should remain in `utils`, though, as they're pretty generic.

packages/core/src/lib/collect-and-persist.ts
BioPhoton: added docs

packages/utils/src/index.ts
BioPhoton: it was a left over

matejchalk (approved):
Very nice improvement :+1:

Issue #75, opened 2023-09-30
Should we enforce  the Dependency Inversion Principle (DIP)? 


ChristopherPHolder:
## Motivation

**This pattern promotes decoupling, making the system easier to maintain, scale, and modify. It also facilitates unit testing since you can mock the abstractions.**

1. High-level modules should not depend on low-level modules. Both should depend on abstractions.
2. Abstractions should not depend on details. Details should depend on abstractions.

|    |   |
| ---      | ---       |
| ![image](https://github.com/flowup/quality-metrics-cli/assets/40126819/af300e67-3c5e-4bb4-bebc-6acc50bf00a1) |  ![image](https://github.com/flowup/quality-metrics-cli/assets/40126819/3e167298-a212-4330-b5d9-9adccb5460fe)   |

## Implementation Overview 

We can achieve this be making sure the `Core Module`  and the `CLI` both depend on the abstractions in the `Models Module`. This means we would move the abstraction there and that module will actually dictate the behavior of the system. 

![image](https://github.com/flowup/quality-metrics-cli/assets/40126819/383bd6a3-1e4b-48c2-8d82-55dfccd5ecf6)
(In the diagram you can see the the dependency between CLI and Core is a doted line. This is because the core module should implement the abstraction in Models and therefore CLI actually depends on Models and Core could be replace with anything that satisfies the abstractions in models)

By making sure the abstraction are in the `Models Module` and that both the `Core Module`  and the `CLI` depend on these abstraction we essentially inverting the dependency such that higher-level and lower-level modules both depend on an abstraction rather than the higher-level module depending on the details of a lower-level module.

Because the `Core Module` handles implementing the abstraction from the `Models Module`, the `CLI` does not require depending on the `Utils Module`. Therefore I suggest we enforce module boundaries  on the `CLI` so that it can only depend on `Core` and `Models`. 

![image](https://github.com/flowup/quality-metrics-cli/assets/40126819/f3877101-5b2a-48a6-8b4c-e2c1b2b2d2bc)

This would also mean that the `Plugins Modules` do not require the `Core Module` , as all abstraction needed to create a plugin are located in the `Models Module` and all the function that would help the plugin authors create these plugins are located in the `Utils Module`. 

![image](https://github.com/flowup/quality-metrics-cli/assets/40126819/4bbd3544-3b8f-46bf-961e-af4b1593846f)

[...]

PR #87, opened 2023-10-04
chore: tag packages with shared/core/plugin/tooling and set dep constraints

### LINKED ISSUE #28: Nx project tags and lint rule setup
- [x] define system for tagging Nx projects
- [x] automate boundary checks using `@nx/enforce-module-boundaries` lint rule options

matejchalk:
Closes #28 

Inspired by discussion in #75 

## Nx tags and project graph (updated)

image

[Excalidraw link](https://excalidraw.com/#json=G4yPscrHmeVyCQEMCJiUW,WqpNLsWLXQLk6E_iRPKnPQ)


ChristopherPHolder:
Why are you creating a "shared" tag? 

Isn't "utils" and "models" already a common way to structure a project? 

I also don't think the CLI should be importing from "utils".  Doing this means that the cli depends directly on the implementations in utils. 

Currently it is only importing a couple of things which probably should be in core. 

For example `importModule` , will this be used by a plugin? Or is this part of the core functionality? 

Same with `logErrorBeforeThrow` and with `CollectOptions`

matejchalk:
> Why are you creating a "shared" tag?
> 
> Isn't "utils" and "models" already a common way to structure a project?
> 
> I also don't think the CLI should be importing from "utils". Doing this means that the cli depends directly on the implementations in utils.
> 
> Currently it is only importing a couple of things which probably should be in core.
> 
> For example `importModule` , will this be used by a plugin? Or is this part of the core functionality?
> 
> Same with `logErrorBeforeThrow` and with `CollectOptions`

My thinking was to separate `core` and `plugin` scopes, so everything that's available to both is `shared`.

What you're suggesting would require even more granularity in the tags. Basically, every package would need to have it's own dedicated tag, except for plugins.

In other words, instead of:

|package|tag|
|:--|:--|
|cli|core|
|core|core|
|utils|shared|
|models|shared|
|plugin-eslint|plugin|
|plugin-lighthouse|plugin|
|nx-plugin|tooling|

we would have:

|package|tag|
|:--|:--|
|cli|cli|
|core|core|
|utils|utils|
|models|models|
|plugin-eslint|plugin|
|plugin-lighthouse|plugin|
|nx-plugin|nx-plugin|

I originally wanted to avoid having many unique tags. I've previously thought of tags as a way to group projects by different dimensions (`scope` and `type`, usually), but maybe that only makes sense for applications. So if you think it's better to have single-purpose tags, then I can do it this way.

ChristopherPHolder (approved):

matejchalk:
Yeah, it's quite possible we will be wiser about the project boundaries later on, so I'm all for going with a simple version for now and extending it later once it's clearer what we need.

Adding it to the project README is a good idea :+1: I'm happy to do that in this PR, or in a follow-up if you prefer (it's about time we replaced the generated Nx README anyway).

I really the Greenforce monorepo example you shared, though :slightly_smiling_face:  I've used the same 2D tags system suggested by Nx for several application monorepos ([the portal also uses it](https://github.com/flowup/quality-metrics-portal/blob/main/.eslintrc.json#L81-L164)), and this has even more types of libraries that could be useful. Unfortunately, I don't think a lot of them apply for this repo, which is a lot smaller and simpler.

But it got me thinking if we couldn't apply some of it after all :thinking: I would consider the tags I introduced to be `scope` tags, so we maybe could introduce the `type` tag dimension as well. Specifically, I think `type:shell`, `type:feature` and `type:util` (maybe also `type:test-util` or even `type:e2e-util` in the future) could apply for us:

|project|scope|type|
|:--|:--|:--|
|cli|core|shell|
|core|core|feature|
|utils|shared|util|
|models|shared|util|
|plugin-eslint|plugin|feature|
|plugin-lighthouse|plugin|feature|
|nx-plugin|tooling|shell|

What do you think?

ChristopherPHolder:
# This is not finished, will finish and pin you 


### Applications

| Type | Contains | Allow Dependencies |
| --- | --- | --- |
| `app` |  Entry point for the executable |  `shell`  | 
| `app-e2e | Runs E2E tests on the executable | `e2e-utils` |

### Library types

| Type | Contains | Allow Dependencies |
| --- | --- | --- |
| `shell` | Entrypoint for an application or domain. Orchestration and routing. | `util` `feature` |
| `util` | Low-level utilities used by many libraries and applications (services, pure functions, contants). | `util` |
| `feature` | TODO | `domain` `util` | 


Again here i am making the distinction that the cli is an application not a library. 

By defining the `Allowed Dependencies` are specifying the scope.


The result would be this: 

| Project | Type |  
| --- | --- | 
| cli | app |
| core | |
| utils | |
| models | | 
| plugin-eslint |  |
| plugin-lighthouse |  |
| nx-plugin |  |

ChristopherPHolder (approved):

matejchalk:
@ChristopherPHolder OK, I've made some changes based on what makes sense at this time. I recommended taking a look at the updated PR description.

I introduced the 2D tagging, although I just went with types `app`, `feature`, `util` and `e2e`.

I didn't go with `domain` because I don't think that distinction is useful here as we only have one "domain". And it would constrain us from importing types from `models` into `utils`. We have quite a few of those imports (e.g. we have report formatting logic in `utils` which needs the `Report` type), and I'm not sure refactoring it would be worthwhile (not much would be left in `utils` after that).

I agree with `cli` being more of an application than a library, especially now that we want the entry point to be executable. I changed the Nx project type as well to reflect it. But since this now makes it a `type:app`, I didn't see much value in having `type:shell`, so I just went with `type:feature`, same as for plugins.

ChristopherPHolder (approved):

PR #89, opened 2023-10-05
feat(plugin-eslint): implement initializer - loads rules from config and creates audits

### LINKED ISSUE #22: Implement ESLint plugin initializer
- [x] implement plugin initializer
  - loads all ESLint rules referenced by ESLint config and transforms them to audit metadata

matejchalk:
Closes #22 

- `eslintPlugin` function implemented
  - provides plugin metadata 
  - loads ESLint rules from configuration and transforms them to audits (using [ESLint Node API](https://eslint.org/docs/latest/integrate/nodejs-api))
- lots of unit tests
  - introduced fixtures with actual eslint configurations (intend to reuse for runner, perhaps even for E2E tests)
  - snapshot testing used for main entry point in order to cover realistic data
- plugin initializer has to be async (I realized this could often be the case with plugins), solved by top-level `await` in core config
- audit metadata automatically converted from ESLint rule metadata
  - rule IDs are slugified
  - when a rule has custom options, they are hashed and appended to slug
    - this is to prevent changes to rule configuration creating a misleading audit result history
    - example:
      - if someone increases the `max` option in `max-lines`, then the number of lint errors will likely decrease without anything actually being improved in the source code, yet it would be indistuishable from real qualitative changes in the audit's progress chart (tracks changes to `value`, which for linter audits equals number of errors)
      - solved by creating different slugs, so changes to rule options will result in a different audit
  - title uses rule's description (more display friendly than rule name, e.g. `no-unused-vars` vs `Disallow unused variables`)
  - description includes info about what the ESLint rule is called, what plugin it's from and which custom options are used (Markdown-formatted)


vmasek (approved):

packages/plugin-eslint/src/lib/meta/rules.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    it('should not include rule which was turned off', async () => {
+      await expect(listRules(eslint, patterns)).resolves.not.toContainEqual(
+        expect.objectContaining({
+          ruleId: 'no-extra-semi',
+        } satisfies Partial<RuleData>),
+      );
+    });
+
+    it('should include rule added by to root config by project config', async () => {
```
Tlacenka, line 199:
```suggestion
    it('should include rule added to root config by project config', async () => {
```

packages/plugin-eslint/src/lib/eslint-plugin.spec.ts
```diff
@@ -1,7 +1,51 @@
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+import type { SpyInstance } from 'vitest';
 import { eslintPlugin } from './eslint-plugin';
 
 describe('eslintPlugin', () => {
```
Tlacenka, line 6: Non-existent `eslintrc` path behaviour could also be tested.

Tlacenka, line 6: Scenario: There is a project which has multiple root ESLint configs, e.g. separately for production and test code. The ESLint rules should be applied on the same monorepo but different files and be potentially managed by the same team. Should multiple plugins be created or will the configuration of the plugin correctly work for such configuration?

packages/plugin-eslint/src/lib/meta/rules.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+            unexpectedVar: 'Unexpected var, use let or const instead.',
+          },
+          schema: [],
+          type: 'suggestion',
+        },
+        options: [],
+      } as RuleData);
+    });
+
+    it('should include plugin rule set implicitly by extending recommended config', async () => {
```
Tlacenka, line 168: It would be valuable to add a note in regards to which config that is supposed to be for future reference. Also related to the other implicit rule.

packages/plugin-eslint/src/lib/meta/rules.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    beforeAll(() => {
+      cwdSpy.mockReturnValue(nxRootDir);
+    });
+
+    it('should list expected number of rules', async () => {
+      await expect(listRules(eslint, patterns)).resolves.toHaveLength(69);
+    });
+
+    it('should include explicitly set plugin rule with custom options', async () => {
```
Tlacenka, line 111: It could be valuable to add a check for a case where the same rule but with different custom options targetting different file pattern is present in ESLint config.

BioPhoton (approved):
nice test setup with the example!

appreciate that you placed the `slugify` function into the `utils` package! ❤️

packages/plugin-eslint/src/lib/meta/rules.spec.ts
matejchalk: I'll revisit this later. I tried, but it's pretty complex and I don't want to prolong this PR too much. There will be more ESLint-related PRs.

packages/plugin-eslint/src/lib/meta/rules.spec.ts
matejchalk: Added some comments.

packages/plugin-eslint/src/lib/eslint-plugin.spec.ts
matejchalk: Test case added.

matejchalk: Yes, that can be achieved with ESLint's glob-based configs. It could be valuable to create some helper function that creates such a config for an Nx monorepo, but that's out of scope of this PR.

Tlacenka (approved):

Issue #100, opened 2023-10-11
E2E test for ESLint plugin


matejchalk:
- [x] create Nx project in `examples` with code from [Todos app](https://github.com/flowup/todos-app)
- [x] example project will include a valid `code-pushup.config.js` with the ESLint plugin (later Lighthouse as well)
- [x] main E2E test will execute `collect` on example project and check outputs (via snapshots)

PR #101, opened 2023-10-11
feat(plugin-eslint): implement runner

### LINKED ISSUE #27: Implement ESLint runner
Implement runner

- [ ] runs ESLint using provided config
- [ ] transforms output to conform to plugin runner output models
- [ ] (nice to have) if affected files provided as environment variable (by core ), restrict ESLint to those files only

matejchalk:
Closes #27 

### Notes
- implemented and tested runner logic
- runs ESLint via Node API and converts results to audit outputs
- change audit outputs in `models` to include `slug`, but not other metadata (already provided in initializer, so unnecessary burden on plugin author)
  - added logic to `core` which looks up metadata by audit slug to create reports
- since ESLint results only include failed rules, all audits slugs have to be passed to runner so that it can report passed audits
  - might also be necessary for other "linter" plugins (Stylelint, TypeScript), in which case the core could provide a built-in solution for it :thinking: 
- additional entry point added for executable `bin.js` file
image

### Next steps
- #100


packages/plugin-eslint/src/lib/runner/index.ts
```diff
@@ -0,0 +1,58 @@
+import type { Audit, AuditOutput, RunnerConfig } from '@code-pushup/models';
+import { toArray } from '@code-pushup/utils';
+import { mkdir, writeFile } from 'fs/promises';
+import { dirname } from 'path';
+import { lint } from './lint';
+import { lintResultsToAudits } from './transform';
+
+export const RUNNER_OUTPUT_PATH =
+  'node_modules/.code-pushup/eslint/runner-output.json';
```
BioPhoton, line 9: Why do we use this folder in `node_modules`

BioPhoton (approved):
Looking forward to get this merged!

packages/plugin-eslint/src/lib/runner/index.ts
matejchalk: The idea is to use some hidden directory, so we don't pollute the Git repo. Since basically all our users will already have a `node_modules` folder in `.gitignore`, it's a convenient place to put these files. I got the idea from other packages which use it:

image

An alternative approach that's also very common is to create a dedicated folder in the root (`.code-pushup` in our case) and automatically add it to `.gitignore` in some init script (e.g. `.angular`, `.firebase`). I think that's also a viable alternative, although it adds an extra setup step.

PR #102, opened 2023-10-12
refactor(models): name file related path more specific

### LINKED ISSUE #96: Rename path related properties in core config
rename `coreconfig.persist.outputPath` to `coreconfig.persist.outputDir`
rename `coreconfig.plugins[0].runner.outputPath` to `coreconfig.plugins[0].runner.outputFile`

BioPhoton:
closes #96


packages/utils/src/lib/mock/schema-helper.mock.ts
```diff
@@ -47,14 +47,14 @@ export function mockPluginConfig(opt?: {
...
+        `require('fs').writeFileSync('${outputDir}', '${JSON.stringify(
           audits.map(
             ({ slug }) =>
               mockAuditReport({ auditSlug: slug }) satisfies AuditReport,
           ),
         )}')`,
       ],
-      outputPath: outputPath,
+      outputFile: outputDir,
     },
```
matejchalk, line 58: Changing this variable to `outputDir` is misleading, should either be `outputFile` or stay as `outputPath`.

matejchalk (changes requested):
One related naming issue that I just thought of - we have the `--configPath` CLI argument. I suppose we should change that too. Either to `--configFile` or even just `--config` would be fine. (Can be a follow-up PR if you prefer.)

packages/utils/src/lib/mock/schema-helper.mock.ts
BioPhoton: nice catch. renamed to outputFile

packages/utils/src/lib/mock/schema-helper.mock.ts
```diff
@@ -78,11 +78,11 @@ export function mockAuditConfig(opt?: {
 }
 
 export function mockPersistConfig(opt?: Partial<PersistConfig>): PersistConfig {
-  let { outputPath, format } = opt || {};
-  outputPath = outputPath || __outputPath__;
+  let { outputDir, format } = opt || {};
+  outputDir = outputDir || __outputFile__;
```
matejchalk, line 82: This looks like a mistake. Should be probably be `'tmp'`.

BioPhoton: done

matejchalk: Is it? I still see the same diff :confused: 

image

BioPhoton: now

packages/utils/src/lib/mock/schema-helper.mock.ts
```diff
@@ -13,7 +13,8 @@ const __pluginSlug__ = 'mock-plugin-slug';
 const __auditSlug__ = 'mock-audit-slug';
 const __groupSlug__ = 'mock-group-slug';
 const __categorySlug__ = 'mock-category-slug';
-const __outputPath__ = 'tmp/out-execute-plugin.json';
+const __outputDir__ = 'tmp';
+const __outputFile__ = 'out-execute-plugin.json';
```
matejchalk, line 17: All files inside `tmp`, please :pray: 

```suggestion
const __outputFile__ = `${__outputDir__}/out-execute-plugin.json`;
```

matejchalk (approved):

PR #108, opened 2023-10-12
refactor(cli): add autorun command

BioPhoton:


BioPhoton:
i cant see any formatting issues :(

e2e/cli-e2e/tests/__snapshots__/help.spec.ts.snap
```diff
@@ -4,7 +4,10 @@ exports[`CLI help > should print help 1`] = `
 "[1mCode PushUp CLI[22m
 
 Commands:
-  code-pushup          Run Plugins and collect results                 [default]
+  code-pushup          Autorun executes the collect and upload command after ano
+                       ther                                            [default]
+  code-pushup autorun  Autorun executes the collect and upload command after ano
+                       ther
```
matejchalk, line 10: The line break here is unfortunate :grimacing:

I guess we need to keep the descriptions short. How about this?

```suggestion
  code-pushup autorun  Shortcut for running collect followed by upload
```

packages/cli/src/lib/autorun/command-object.ts
```diff
@@ -0,0 +1,19 @@
...
+import { CommandBase } from '../implementation/model';
+
+export function yargsAutorunCommandObject() {
+  return {
+    command: 'autorun',
+    describe: 'Autorun executes the collect and upload command after another',
+    handler: async <T>(args: ArgumentsCamelCase<T> & CommandBase) => {
+      const options = {
+        ...args,
+        upload: uploadConfigSchema.parse(args.upload),
```
matejchalk, line 13: I don't understand why the `upload` part of the config is parsed within the handler, unlike everything else. Doesn't the middleware take care of this? Isn't this parsing it again?

packages/cli/src/lib/implementation/config-middleware.ts
```diff
@@ -1,13 +1,16 @@
...
-export async function configMiddleware<T extends ArgsCliObj>(processArgs: T) {
+export async function configMiddleware<T extends TerminalArgsObj>(
+  processArgs: T,
+) {
   const args = processArgs as T;
   const { config, ...cliOptions }: GlobalOptions =
     globalOptionsSchema.parse(args);
   const importedRc = await readCodePushupConfig(config);
   const cliConfigArgs = readCoreConfigFromCliArgs(processArgs);
+  console.log('importedRc: ', importedRc);
```
matejchalk, line 13: Leftover from debugging, I assume? We can log it of course, but only when `--verbose` is set, which is out of scope for this PR, I suppose.

```suggestion
```

packages/cli/src/lib/upload/command-object.ts
```diff
@@ -1,10 +1,18 @@
...
 
 export function yargsUploadCommandObject() {
   return {
     command: 'upload',
     describe: 'Upload report results to the portal',
-    handler: upload as unknown as CommandModule['handler'],
+    handler: async <T>(args: ArgumentsCamelCase<T> & CommandBase) => {
+      const uploadOptions = {
+        ...args,
+        upload: uploadConfigSchema.parse(args.upload),
```
matejchalk, line 13: Same question as for `autorun`.

matejchalk (changes requested):

packages/cli/src/lib/autorun/command-object.ts
BioPhoton: We have `upload` as optional in `CoreConfig` and the schema

packages/cli/src/lib/upload/command-object.ts
BioPhoton: see upload command

e2e/cli-e2e/tests/__snapshots__/help.spec.ts.snap
matejchalk: The description hasn't been changed, so marking as unresolved.

My code suggestion is just for demonstration in this case, obviously this has to be changed at source and then updated in the snapshots (with `-u`), not by just accepting the suggestion in GitHub. Would it help if I was clearer when it's a suggestion that can be accepted "as is", or when it's just a diff for illustration (has to changed in multiple places)?

packages/cli/src/lib/upload/command-object.ts
matejchalk: I see. That makes sense for the `upload` command, but it's not at all clear from the code.

How about just throwing an Error like this? I guess that would be more readable than whatever `parse` throws.

```ts
if (!args.upload) {
  throw new Error('Upload configuration not set');
}
// ... use args
```

packages/cli/src/lib/autorun/command-object.ts
matejchalk: In the case of `autorun` (unlike `upload` command), I would say if the upload config is missing, then we just don't run the `upload` command, only `collect` (we can print a warning to the console instead). I believe this is more inline with the Lighthouse CI `autorun`.

e2e/cli-e2e/tests/__snapshots__/help.spec.ts.snap
BioPhoton: missed it!

BioPhoton: done

packages/cli/src/lib/upload/command-object.ts
BioPhoton: I can throw now and in the future we can add that validation to the command options

packages/cli/src/lib/autorun/command-object.ts
BioPhoton: I like the addition to `autorun`. This might help to not loose data.

matejchalk: But then we can't point portal-less users to `code-pushup` or `code-pushup autorun`, because it'll fail. And I'm not sure `autorun` should be the default command in that case, because it won't work out of the box.

Wouldn't it be better to print something like this to the end of stdout?
"Your report is ready, but hasn't been uploaded to the portal. If you add upload credentials to the config file, you can then run `code-pushup upload`."

matejchalk: Might've misunderstood if you were agreeing with me or not, now that I see the code changes :sweat_smile:

matejchalk (dismissed):

packages/cli/src/lib/autorun/command-object.ts
BioPhoton: I was arranging.. I console.warned. 
But like to think more about "work out of the box".

> `autorun` collected that report, but hasn't been uploaded to the portal because of missing credentials. Add upload credentials to the config file, and run `code-pushup upload` to upload the collected report.

PR #112, opened 2023-10-15
refactor: add config command

BioPhoton:


packages/cli/src/lib/config/command-object.ts
```diff
@@ -0,0 +1,12 @@
+import { CommandModule } from 'yargs';
+
+export function yargsConfigCommandObject() {
+  return {
+    command: 'config',
+    describe: 'Print config',
+    handler: args => {
+      console.log('Config: ');
+      console.log(args);
```
matejchalk, line 9: I think it would be useful if this command only emitted valid JSON to the stdout, making it much easier to parse with other tools.

Towards this end, I suggest the following changes:
- drop the `Config: ` line (messes up the JSON formatting and doesn't really need to be reiterated, the user knows what command they ran)
- `console.log`ing an object won't expand nested properties, will include include color codes, etc. - to ensure it's always valid JSON, serialize it with `JSON.stringify`

image


```suggestion
      console.log(JSON.stringify(args, null, 2));
```

packages/cli/src/lib/config/command-object.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  it('should override config with CLI arguments', async () => {
+    const args = [
+      ...baseArgs,
+      ...objectToCliArgs({
+        format: 'md',
+      }),
+    ];
+    const parsedArgv = await cli(args).parseAsync();
+    expect(parsedArgv.persist.outputDir).toBe('tmp');
+    expect(parsedArgv.persist?.format).toEqual(['md']);
```
matejchalk, line 39: This is only testing the arguments, but what's logged to the console isn't tested at all, even though that's the end result we care about.

matejchalk (changes requested):

packages/cli/src/lib/config/command-object.spec.ts
BioPhoton: added snapshot to stdout

BioPhoton: mocked console, added snapshot. Only problem I have ATM is getting rid of the path in the snapshot.

BioPhoton: this is not working:

```typescript
 beforeEach(async () => {
    logs = [];
    mockConsole(msg => {
      const cleanMsg = msg.replace(
        /(\\"config\\\\": \\\\".*?\\\\")/m,
        `\\"config\\": \\"XXX\\"`,
      );
      logs.push(cleanMsg);
    });
  });
```

e2e/cli-e2e/tests/config.spec.ts
```diff
@@ -14,18 +16,61 @@ const configFile = (ext: Extension) =>
...
+        apiKey: 'e2e-api-key',
+        server: 'https://e2e.com/api',
+      },
+      persist: { outputDir: `tmp/${ext}` },
+      plugins: expect.any(Array),
+      categories: expect.any(Array),
+    });
+  });
+
+  it('should load .%s config file and merge cli arguments', async () => {
```
matejchalk, line 52: The `%s` only gets replaced when used with `.each`.

```suggestion
  it('should load .ts config file and merge cli arguments', async () => {
```

e2e/cli-e2e/tests/__snapshots__/collect.spec.ts.snap
```diff
@@ -1,5 +1,1007 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`CLI collect > should create report.md 1`] = `
```
matejchalk, line 3: There's a new snapshot, but the `collect.spec.ts` test file wasn't changed, so I supposed it's just a leftover?

You can easily remove the obsolete snapshot with `npx nx e2e cli-e2e -u`.

matejchalk (approved):

PR #131, opened 2023-10-18
fix(testing): fix tests on windows

Tlacenka:
In this PR, I identified and fixed various issue while running tests in the Windows environment:
- shell needs to be specified to run programs in command shell on Windows, as described [here](https://stackoverflow.com/questions/60386867/node-spawn-child-process-not-working-in-windows)
- path notation on Windows is different
- JSON formatting is different for UNIX and Windows
- writing into files replaced by `echo` command which works seamlessly on Windows
- glob patterns require extra quotation marks for argument expansion on Linux

I also extracted the runner config for writing report into a file to a separate function (there was a lot of duplicate code to keep consistent).


matejchalk (commented):
Really great to have this fixed finally, so that the codebase is now usable on Windows :sweat_smile: 

As discussed, I created a follow-up for automatically checking it: #141

e2e/cli-e2e/tests/collect.spec.ts
```diff
@@ -55,12 +58,7 @@ describe('CLI collect', () => {
   it('should print report summary to stdout', async () => {
     const { code, stdout, stderr } = await executeProcess({
       command: 'npx',
-      args: [
-        '../../dist/packages/cli',
-        'collect',
-        '--verbose',
-        '--persist.format=stdout',
-      ],
+      args: [cliPath, 'collect', '--verbose', '--persist.format=stdout'],
```
BioPhoton, line 61: You could use
https://github.com/flowup/quality-metrics-cli/blob/090e34d64c6ccb8ced01a615ea071c2fda83ea97/packages/utils/src/lib/execute-process.ts#L189

to generate the args in a more maintainable way

packages/cli/test/config.mock.cjs
```diff
@@ -17,23 +20,16 @@ module.exports = {
...
-            [
-              {
-                title: 'dummy-title',
-                slug: 'command-object-audit-slug',
-                value: 0,
-                score: 0,
-              },
-            ],
-          )}')`,
+      runner: createFileWriteRunnerConfig(
```
BioPhoton, line 23: we have mock functions in models for that situation living in the `test/fixture` folder

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
```diff
@@ -1,5 +1,7 @@
 import { defaultConfig } from 'lighthouse';
+import { join } from 'path';
 import { AuditOutputs, PluginConfig } from '@code-pushup/models';
+import { createFileWriteRunnerConfig } from '@code-pushup/utils';
```
BioPhoton, line 4: Lets move this into `models/test/fixtures/runner.config.mock.ts` please

tsconfig.base.json
```diff
@@ -17,6 +17,7 @@
     "skipDefaultLibCheck": true,
     "baseUrl": ".",
     "resolveJsonModule": true,
+    "allowSyntheticDefaultImports": true,
```
BioPhoton, line 20: For what code do we need this?

BioPhoton (changes requested):
Nice improvement of testing! Left some comments.

packages/utils/src/lib/file-write-runner-config.ts
```diff
@@ -0,0 +1,18 @@
...
+  outputFilePath: string,
+): RunnerConfig {
+  const auditOutput =
+    platform() === 'win32'
+      ? JSON.stringify(output)
+      : "'" + JSON.stringify(output) + "'";
+  return {
+    command: 'echo',
+    args: [auditOutput, '>', toUnixPath(outputFilePath)],
+    outputFile: toUnixPath(outputFilePath),
```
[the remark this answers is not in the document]
vmasek: maybe [path.join](https://nodejs.org/api/path.html#pathjoinpaths) or [path.normalize](https://nodejs.org/api/path.html#pathnormalizepath) could be used instead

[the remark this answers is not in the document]
matejchalk: I think just `outputFile: outputFilePath` should be fine, the file path should be platform-dependant.

The `toUnixPath` function is needed by the ESLint plugin, because it emits source file issues and needs to convert them absolute platform-dependant paths to relative Git paths. I'm not aware of any other use case.

tsconfig.base.json
Tlacenka: In order to mock platform for unit tests, I needed to create a spy for the `os.platform()` which required import directly from `os`. Unfortunately, this caused a default export TS error and I had to override it with this option.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
Tlacenka: Unfortunately, this function cannot live in the `models` package as it would cause a circular dependency. Until a separate testing utils package is created, Matthew advised me to add it to `utils` and account for one separate usage by duplicating the function logic.

packages/cli/test/config.mock.cjs
Tlacenka: Again, this is a function that is reusable for multiple tests (currently in cli and lighthouse plugin) and otherwise would need to be duplicated. In one of my new issues, I plan to address this by extracting reusable functions to a separate package instead of duplicating logic in multiple packages.

packages/utils/src/lib/file-write-runner-config.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+        : "'" + JSON.stringify(audits) + "'";
+
+    const filePath = join(process.cwd(), 'tmp/report.json');
+    const unixFilePath = toUnixPath(filePath);
+
+    expect(createFileWriteRunnerConfig(audits, filePath)).toStrictEqual({
+      command: 'echo',
+      args: [osAuditOutput, '>', unixFilePath],
+      outputFile: unixFilePath,
+    });
```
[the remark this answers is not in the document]
Tlacenka: If I did that, I would then be testing the unix path function which already has separate tests. Removed.

packages/utils/src/lib/file-write-runner-config.ts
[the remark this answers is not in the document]
Tlacenka: If you say so. I wanted to make sure that the resulting path would be safe but if that is the responsibility of the caller, sure.

e2e/cli-e2e/tests/collect.spec.ts
Tlacenka: This test was written by @matejchalk, I merely extracted one of the arguments to a variable. What do you think about this suggestion?

From my point of view, having arguments as an explicit array increases the test readability which is what you want in test code.

matejchalk: I also don't see the point of `objectToCliArgs`, to be honest. This CLI notation is what the user sees, which makes it more fitting for an E2E test, the wrapper function just obfuscates it.

Tlacenka:
There were too many conflicts so I created a new branch and PR #165

PR #134, opened 2023-10-20
refactor: use new fixtures in all tests

### LINKED ISSUE #88: Improve testing setup
The testing setup is suboptimal because of:
- VFS is boilerplatey to set up
- hard files don't get created or cleaned up in a standardised way

BioPhoton:
closes #88


packages/core/src/lib/collect-and-persist.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    cleanFolderPutGitKeep();
+  });
+
+  test('should work', async () => {
+    await collectAndPersistReports({
+      verbose: false,
+      ...minimalConfig(outputDir),
+    });
+    const result = JSON.parse(readFileSync(reportPath()).toString()) as Report;
+    expect(result.plugins[0]?.audits[0]?.slug).toBe('audit-1');
```
matejchalk, line 41: There's no `.toMatchSnapshot()`, so the `.snap` file can be deleted. If you run with `-u`, it will delete obsolete snapshots.

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -86,18 +93,28 @@ export async function executePlugin(
...
+
+    if (description) {
+      pluginReport.description = description;
+    }
+    if (docsUrl) {
+      pluginReport.docsUrl = docsUrl;
+    }
+    if (groups) {
+      pluginReport.groups = groups;
+    }
```
matejchalk, line 115: I'd prefer if we did this in a less imperative way.

```ts
return {
  ...description && { description },
  // ...
}
```

Is it important to omit the falsy properties entirely, though?

packages/core/src/lib/implementation/persist.spec.ts
```diff
@@ -34,70 +34,86 @@ const readReport = (format: 'json' | 'md') => {
...
+const setupConsole = async () => {
+  logs = [];
+  mockConsole(msg => logs.push(msg));
+};
+const teardownConsole = async () => {
+  logs = [];
+  unmockConsole();
+};
+
+// @TODO refactor away from snapshots in favour of disc space and readability
```
matejchalk, line 62: I think it's valuable to have the snapshots, but I would only have one per each format. That meangs that `should persist all formats` and `should persist some formats` wouldn't have `.toMatchSnapshot` as well, because the content is already tested elsewhere. That would also help with the file not being too big.

e2e/cli-e2e/mocks/fs.mock.ts
```diff
@@ -0,0 +1,29 @@
...
+) {
+  rmSync(dirName, { recursive: true, force: true });
+  mkdirSync(dirName);
+  writeFileSync(join(dirName, '.gitkeep'), '');
+  if (content) {
+    for (const fileName in content) {
+      writeFileSync(join(dirName, fileName), content[fileName]);
+    }
+  }
+}
```
matejchalk, line 29: I don't really see the point of these functions. The `tmp` directory is already being automatically cleaned between tests in [`global-setup.ts`](https://github.com/flowup/quality-metrics-cli/blob/main/global-setup.ts). If you want a `.gitkeep` file, you can add it there. I think that's a much better solution that duplicating it everywhere.

- Vitest docs
  - [globalSetup](https://vitest.dev/config/#globalsetup) - that's what each project should be using now
  - [setupFiles](https://vitest.dev/config/#setupfiles) - we can use this as well (advantage is you can add `beforeEach` blocks, etc.)

packages/models/src/lib/plugin-config.spec.ts
```diff
@@ -1,114 +1,84 @@
...
-  });
+    groups[0].refs[0].slug = invalidAuditRef;
+    pluginConfig.groups = groups;
 
-  it('should throw if slugs of audits are duplicated', () => {
-    const out = mockAuditOutputs({ auditSlug: ['a', 'a'] });
-    expect(() => auditOutputsSchema.parse(out)).toThrow(
-      'In plugin audits the slugs are not unique',
+    expect(() => pluginConfigSchema.parse(pluginConfig)).toThrow(
+      `slug has to follow the patter`,
```
matejchalk, line 81:
```suggestion
      `slug has to follow the pattern`,
```

packages/models/test/fixtures/categories.mock.ts
```diff
@@ -67,5 +67,5 @@ export function categoryConfig(
 }
 
 export function categoryConfigs(): CategoryConfig[] {
-  return Object.values(CATEGORIES_MAP);
+  return Object.values(JSON.parse(JSON.stringify(CATEGORIES_MAP)));
```
matejchalk, line 70: Why? :raised_eyebrow:

packages/models/test/fixtures/eslint-audits.mock.ts
```diff
@@ -838,4 +838,4 @@ export const ESLINT_AUDITS_MAP = {
     docsUrl:
       'https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/require-render-return.md',
   },
-} satisfies Record<string, AuditReport>;
+};
```
matejchalk, line 841: Please revert this :pray: It removes the type-safety of the audit slugs, I'm using `keyof typeof ESLINT_AUDITS_MAP` (I want that to be a union, not a `string`) to make the audit references more manageable :sweat_smile:

packages/models/test/fixtures/eslint-plugin.mock.ts
```diff
@@ -1,14 +1,44 @@
...
+import { runnerConfig } from './runner.mock';
+
+const eslintMeta: Omit<PluginConfig, 'audits' | 'runner'> = {
+  slug: 'eslint',
+  title: 'ESLint',
+  icon: 'eslint',
+  description: 'Official Code PushUp ESLint plugin',
+  packageName: '@code-pushup/eslint-plugin',
+  version: '0.1.0',
+};
```
matejchalk, line 17: This should be easier to manage if the `PluginConfig` changes:

```suggestion
const eslintMeta = {
  slug: 'eslint',
  title: 'ESLint',
  icon: 'eslint',
  description: 'Official Code PushUp ESLint plugin',
  packageName: '@code-pushup/eslint-plugin',
  version: '0.1.0',
} satisfies Partial<PluginConfig>;
```

packages/models/test/fixtures/persist-config.mock.ts
```diff
@@ -0,0 +1,8 @@
+import { PersistConfig } from '../../src';
+
+export function persistConfig(opt?: Partial<PersistConfig>): PersistConfig {
+  return {
+    outputDir: 'tmp',
+    ...(opt || {}),
```
matejchalk, line 6: Object spread automatically omits non-objects.

```suggestion
    ...opt,
```

packages/core/src/lib/implementation/persist.spec.ts
matejchalk: Actually, I see the actual formatting is already tested in `utils`. In that case, maybe it's enough to have it there? We should have the most detailed test "at source" (less setup needed, easier to test many different scenarios), wrapper functions don't have to test in so much detail.

matejchalk (changes requested):

packages/cli/src/lib/implementation/config-middleware.spec.ts
```diff
@@ -1,33 +1,34 @@
...
 import { configMiddleware } from './config-middleware';
-import { getDirname } from './helper.mock';
-
-const __dirname = getDirname(import.meta.url);
 
+const __dirname = dirname(fileURLToPath(import.meta.url));
 const withDirName = (path: string) => join(__dirname, path);
 const config = (ext: string) =>
-  `${withDirName('../../../test/config.mock.')}${ext}`;
+  `${withDirName(
```
Tlacenka, line 9: This path notation will break on Windows. Depending which PR gets merged first, I can fix it in mine when I rebase.

packages/models/test/fixtures/categories.mock.ts
BioPhoton: I had mutation bugs

e2e/cli-e2e/tests/__snapshots__/collect.spec.ts.snap
```diff
@@ -1,5 +1,1007 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`CLI collect > should create report.md 1`] = `
```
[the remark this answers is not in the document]
BioPhoton: Done

packages/core/src/lib/collect-and-persist.spec.ts
BioPhoton: Done

packages/core/src/lib/implementation/persist.spec.ts
BioPhoton: Done. Good point!

e2e/cli-e2e/mocks/fs.mock.ts
BioPhoton: The globalSetup is executed only once. I have overlapping files in between tests. 

I opened up a new issue to use vitest setupFile as I'm not familiar ATM.
https://github.com/flowup/quality-metrics-portal/issues/119

packages/core/src/lib/implementation/execute-plugin.ts
BioPhoton: Done, now its more impereative.

> Is it important to omit the falsy properties entirely, though?

Less data saved/sent maybe. But I have no strong feelings.

packages/models/test/fixtures/eslint-audits.mock.ts
BioPhoton: reverted

packages/cli/src/lib/implementation/config-middleware.spec.ts
```diff
@@ -1,33 +1,34 @@
...
+    '../../../../models/test/fixtures/code-pushup.config.mock.',
```
BioPhoton, line 10:
```suggestion
    join('..', '..', '..', '..', 'models', 'test', 'fixtures', 'code-pushup.config.mock.',
```

packages/cli/src/lib/implementation/config-middleware.spec.ts
BioPhoton: quick fixed... please consider it in your PR

packages/models/test/fixtures/categories.mock.ts
matejchalk: Well, that's why it's better to never mutate :sweat_smile:

For production code these additional JSON parse/serialize operations would concern me, but since this is only for test, it not a problem, just quite confusing :man_shrugging:

packages/models/test/fixtures/eslint-plugin.mock.ts
```diff
@@ -1,14 +1,44 @@
...
+        description,
+        title,
+        docsUrl,
+      } satisfies Audit),
+  );
+  return {
+    ...eslintMeta,
+    runner: runnerConfig(
+      Object.values(ESLINT_AUDITS_MAP),
+      `${outputDir}/eslint-out.json`,
```
matejchalk, line 33: For Windows compatibility:

```suggestion
      join(outputDir, 'eslint-out.json'),
```

(+ `import { join } from 'path'`)

packages/utils/src/index.ts
```diff
@@ -9,20 +9,21 @@ export {
...
   readJsonFile,
   readTextFile,
-  slugify,
   toArray,
 } from './lib/utils';
+export { formatBytes } from './lib/report';
+export { slugify } from './lib/report';
+export { calcDuration } from './lib/report';
+export { formatCount } from './lib/report';
+export { compareIssueSeverity } from './lib/report';
```
matejchalk, line 29: All the exports from `./lib/report` can be merged.

(VSCode optimizes imports/exports for me automatically on save, BTW. Maybe WebStorm also has some equivalent?)

image

packages/utils/src/lib/utils.spec.ts
```diff
@@ -136,69 +47,3 @@ describe('distinct', () => {
...
-
-describe('toArray', () => {
-  it('should transform non-array value into array with single value', () => {
-    expect(toArray('src/**/*.ts')).toEqual(['src/**/*.ts']);
-  });
-
-  it('should leave array value unchanged', () => {
-    expect(toArray(['*.ts', '*.js'])).toEqual(['*.ts', '*.js']);
-  });
-});
```
matejchalk, line 148: The `toArray` function is still in `utils.ts`. Please double-check you aren't deleting any tests :pray: 

(Overall I like the splitting of into multiple files, though :+1: )

BioPhoton: It is still here. I sorted the tests in the same order as in the file. Will mention it next time i th ePR comment.

matejchalk: Oh, I see now you just changed the order. Sorry, the diff confused me, false alarm :man_facepalming:

packages/utils/src/index.ts
BioPhoton: done!

matejchalk (approved):

PR #150, opened 2023-10-25
chore(cli): custom targets to simplify manual testing

matejchalk:
I was missing a clear way to "try out" the CLI with something that's expected to work. Replaced `cli:exec` with `cli:run-help`, `cli:run-collect` and `cli:run-print-config`. The new targets use the example app instead of the mocked config.

### DIFF
--- a/packages/cli/project.json
+++ b/packages/cli/project.json
@@ -19,10 +19,6 @@
       "command": "node tools/scripts/publish.mjs cli {args.ver} {args.tag}",
       "dependsOn": ["build"]
     },
-    "exec": {
-      "command": "npx dist/packages/cli --config=./packages/cli/code-pushup.config.ts",
-      "dependsOn": ["build"]
-    },
     "lint": {
       "executor": "@nx/linter:eslint",
       "outputs": ["{options.outputFile}"],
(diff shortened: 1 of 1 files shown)


packages/cli/project.json
```diff
@@ -40,6 +36,24 @@
         "passWithNoTests": true,
         "reportsDirectory": "../../coverage/packages/cli"
       }
+    },
+    "run-help": {
+      "command": "npx dist/packages/cli --help",
```
BioPhoton, line 41: Should be the command not the argument
```suggestion
      "command": "npx dist/packages/cli help",
```

BioPhoton (commented):
Nice small DX improvement. 

Question: It there a `code-pushup.config.json` located in root or is the target failing without an additional `--config=./cfg.ts`?

matejchalk:
> Question: It there a `code-pushup.config.json` located in root or is the target failing without an additional `--config=./cfg.ts`?

All the targets I added succeed without any additional arguments. We don't have any config in our root, but the `run-collect` and `run-print-config` targets are being run in the `examples/react-todos-app` directory (exactly the same as what's being run in E2E tests).

BioPhoton (approved):
approved

PR #162, opened 2023-10-26
feat(cli): disabled version option

### LINKED ISSUE #124: remove version option
> --version              Show version number                       [boolean]

We will remove this command for now and in the future we will add info command
https://yargs.js.org/docs/#api-reference-version

MishaSeredenkoPushBased:
Disabled --version option

Closes #124


matejchalk (approved):
Nothing major, but the convention is to write commit messages in "imperative mood", i.e. `feat(cli): disable version option` instead of `feat(cli): disabled version option` (`disable` instead of `disabled` - think of it like you're giving instructions).

Tlacenka (approved):
CLI no longer supports the version option.

PR #165, opened 2023-10-27
fix(testing): fix tests for Windows

Tlacenka:
In this PR, I identified and fixed various issue while running tests in the Windows environment:

- shell needs to be specified to run programs in command shell on Windows, as described [here](https://stackoverflow.com/questions/60386867/node-spawn-child-process-not-working-in-windows)
- path notation on Windows is different
- JSON formatting is different for UNIX and Windows
- writing into files replaced by echo command which works seamlessly on Windows
- glob patterns require extra quotation marks for argument expansion on Linux
- recreating tmp folder caused issues for me so I partially fixed it via adding the recursive option

Note: This is the same scope as #131. Unfortunately, merging changes from the test refactor did not go well so I started off a clean branch and re-applied the changes.


packages/cli/code-pushup.config.mock.ts
```diff
@@ -103,3 +96,6 @@ export default {
     },
   ],
 };
+function join(outputDir: string, arg1: string): string {
+  throw new Error('Function not implemented.');
+}
```
matejchalk, line 101: Should be removed and replaced with `import { join } from 'path'`. I suspect you clicked on the wrong VS Code suggestion :grin:

packages/models/test/fixtures/file-write-runner-config.ts
```diff
@@ -0,0 +1,17 @@
+import { platform } from 'os';
+import { AuditOutput, RunnerConfig } from '../../src';
+
+export function createFileWriteRunnerConfig(
```
matejchalk, line 4: Maybe it would be better to rename this to `runnerConfig` (i.e. keep the name of the replaced function)? It's shorter and more in line with other mock naming conventions.

packages/plugin-eslint/src/lib/runner/lint.ts
```diff
@@ -16,7 +15,7 @@ export async function lint(
   const results = lintResults.map(
     (result): LintResult => ({
       ...result,
-      relativeFilePath: result.filePath.replace(process.cwd() + sep, ''),
+      relativeFilePath: toUnixPath(result.filePath, true),
```
matejchalk, line 18: I think the usage would be more readable with the named options pattern:

```suggestion
      relativeFilePath: toUnixPath(result.filePath, { toRelative: true }),
```

matejchalk (commented):
Well done and thanks for sticking with it :sweat_smile: Great that the ESLint plugin will now work on Windows, too :+1:

packages/plugin-eslint/src/lib/runner/lint.ts
Tlacenka: Done

matejchalk (approved):

packages/models/test/fixtures/file-write-runner-config.ts
matejchalk: OK, `echoRunnerConfig` is also a better name :slightly_smiling_face:

BioPhoton:
I thought so that the last big test refactoring will break any other one related to testing.

Happy you made it!

PR #166, opened 2023-10-27
ci(testing): run tests on all major platforms

Tlacenka:
In this PR, I add matrix for running both unit and E2E tests on all three major platforms (Linux, Windows, MacOS).
- All jobs should run in parallel, so having all three platforms covered should not have any drawbacks.
- Even if one job fails, the other tests are executed.
- Timeout for E2E is increased due to slower execution in pipelines (mainly Windows).
- Unfortunately, the tests are currently flaky. I added one retry until this issue is resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest]
     name: Unit tests
     runs-on: ${{ matrix.os }}
     steps:
@@ -91,7 +91,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest]
     name: E2E tests
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


e2e/cli-e2e/tests/collect.spec.ts
```diff
@@ -9,6 +9,7 @@ describe('CLI collect', () => {
   const exampleAuditTitle =
     'Require `const` declarations for variables that are never reassigned after declared';
 
+  console.log('hello');
```
MishaSeredenkoPushBased, line 12: Looks like you've forgot to remove the log

packages/utils/src/lib/execute-process.ts
```diff
@@ -136,6 +136,7 @@ export function executeProcess(cfg: ProcessConfig): Promise<ProcessResult> {
   const { next, error, complete } = observer || {};
   const date = new Date().toISOString();
   const start = performance.now();
+  console.log('hello');
```
MishaSeredenkoPushBased, line 139: Same log here :)

e2e/cli-e2e/tests/collect.spec.ts
Tlacenka: This was added on purpose in order for nx to resolve affected projects and run some tests. By default, no projects were matched as the .github folder does not belong in any package.

matejchalk (approved):

PR #175, opened 2023-10-28
test: improve test output

### LINKED ISSUE #172: Hide console.log's in tests wherever possible
closed by #175

BioPhoton:
This PR improves the testing output in the CI and console.
console output is still tested.

closes #172


packages/cli/src/lib/autorun/command-object.spec.ts
```diff
@@ -51,13 +55,20 @@ const cli = (args: string[]) =>
...
     cleanFolderPutGitKeep();
+    mockConsole((...args: unknown[]) => {
+      logs.push(...args);
+    });
   });
   afterEach(() => {
     cleanFolderPutGitKeep();
+    logs = [];
+    unmockConsole();
   });
```
matejchalk, line 72: Seeing how the console mocking is being used, it feels unnecessarily boilerplatey. E.g. the caller still has to be manage the `logs` array themselves, despite the logic always being the same.

If we were to just go with Vitest's built-in spies, it would actually be simpler:

```suggestion
describe('autorun-command-object', () => {
  let logSpy: SpyInstance;

  beforeEach(async () => {
    vi.clearAllMocks();
    cleanFolderPutGitKeep();
    logSpy = vi.spyOn(console, 'log');
  });

  afterEach(() => {
    cleanFolderPutGitKeep();
    logSpy.mockRestore();
  });
```

The only change to the assertions is:
- `expect(logs).toContain('...')` becomes `expect(console.log).toHaveBeenCalledWith('...')`,
- `expect(logs.length).toBe(2)` becomes `expect(console.log).toHaveBeenCalledTimes(0)`.

BioPhoton: > If we were to just go with Vitest's built-in spies, it would actually be simpler:

I was not able to find that in the docs. Thanks!

BioPhoton: Actually i cant reproduce this... still see the logs in the console

matejchalk: I can't help you with that, unfortunately. Tests keep failing on my machine since the big testing refactor in #134 :disappointed: 

image

BioPhoton: will address the flakyness in the next PR

matejchalk (approved):

PR #183, opened 2023-10-29
refactor(models): refine types for better DX and exploration

### LINKED ISSUE #147: improve models types
In the code we often have to access types over the index. `AuditGroup['refs'][0]`

If we would introduce a better organisation in models we could benefit.

Possible new types:
- `type RefBase = {slug: string}`
- `type ScoreableReference = RefBase & { score: number}`
- `type WeightableReference = RefBase & { weight: number}`
...

BioPhoton:
closes #147


packages/models/src/lib/category-config.ts
```diff
@@ -13,22 +13,26 @@ type _RefsList = {
...
+      description:
+        'Discriminant for reference kind, affects where `slug` is looked up',
+    }),
+    plugin: slugSchema(
+      'Plugin slug (plugin should contain referenced audit or group)',
+    ),
+  }),
+);
+
+export type CategoryRefSchema = z.infer<typeof categoryRefSchema>;
```
matejchalk, line 31: The type name shouldn't have the `Schema`-suffix. The naming convention is `const fooSchema = ...` and `type Foo = z.infer<typeof fooSchema>`.

```suggestion
export type CategoryRef = z.infer<typeof categoryRefSchema>;
```

packages/core/src/lib/implementation/json-to-gql.ts
```diff
@@ -71,13 +71,15 @@ export function jsonToGql(report: Report) {
...
+function transformSeverity(severity: CliIssueSeverity): IssueSeverity {
   switch (severity) {
     case 'info':
       return IssueSeverity.Info;
     case 'error':
       return IssueSeverity.Error;
     case 'warning':
       return IssueSeverity.Warning;
+    default:
+      throw new Error(`Severity ${severity} unknown`);
```
matejchalk, line 83: I would omit the default case. The reason is that when we only have known cases and we extend the possible values, then TypeScript will tell us we're missing a case (thanks to `noImplicitReturns` flag) and we'll be forced to fix it at build time (instead of shipping a runtime error).

image

matejchalk (commented):
Overall a very nice improvement :slightly_smiling_face:

packages/models/src/lib/plugin-config.ts
```diff
@@ -183,6 +185,7 @@ const sourceFileLocationSchema = z.object(
   { description: 'Source file location' },
 );
 
+export type IssueSeverity = 'info' | 'warning' | 'error';
```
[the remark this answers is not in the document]
BioPhoton: done

packages/models/src/lib/category-config.ts
BioPhoton: done

matejchalk (approved):

matejchalk (approved):

PR #187, opened 2023-10-30
feat(cli): add `--persist.filename` cli option

BioPhoton:


e2e/cli-e2e/tests/collect.spec.ts
```diff
@@ -23,58 +28,70 @@ describe('CLI collect', () => {
...
   });
 
   it('should run ESLint plugin and create report.json', async () => {
+    const reportFileName = filename();
     const { code, stderr } = await executeProcess({
       command: 'npx',
-      args: [cliPath, 'collect', '--no-progress'],
+      args: [
+        ...baseArgs,
+        '--persist.format=json',
```
matejchalk, line 50: I would omit the explicit format, because I think it's valuable to test that the JSON file gets created regardless. E2E tests should focus on covering critical paths, i.e. what's most commonly used (users usually won't specify JSON format explicitly).

```suggestion
```

packages/core/src/lib/implementation/persist.spec.ts
```diff
@@ -41,15 +41,7 @@ let logs: string[] = [];
...
-  vol.fromJSON(
-    {
-      [reportPath('json')]: '',
-      [reportPath('md')]: '',
-    },
-    MEMFS_VOLUME,
-  );
-  unlinkSync(reportPath('json'));
-  unlinkSync(reportPath('md'));
+  vol.fromJSON({}, MEMFS_VOLUME);
```
matejchalk, line 44: I don't understand why you use the timestamps here. The tests are using `memfs` and the virtual files are being reset in `beforeEach`. Is this change needed here? If not, this is change is adding unnecessary complexity in this case.

packages/utils/src/lib/utils.ts
```diff
@@ -51,7 +54,11 @@ export function distinct<T extends string | number | boolean>(array: T[]): T[] {
   return Array.from(new Set(array));
 }
 
-// === File
+// === Filesystem
+
+export function ensureDirectoryExists(baseDir: string) {
+  if (!existsSync(baseDir)) mkdirSync(baseDir, { recursive: true });
+}
```
matejchalk, line 61: Should use async APIs, so that it's non-blocking.

packages/utils/src/lib/report.ts
```diff
@@ -56,3 +63,26 @@ export function compareIssueSeverity(
...
+        filename
+          ? file.includes(filename)
+          : new RegExp(REPORT_NAME_PATTERN).test(file),
+      )
+      .map(file => {
+        const filePath = join(outputDir, file);
+        return [file, readFileSync(filePath, 'utf8')];
+      })
+  );
+}
```
matejchalk, line 88: Again, async `fs` APIs. This will read the files one-by-one, can easily run in parallel with `Promise.all`.

packages/models/src/lib/implementation/utils.ts
```diff
@@ -6,6 +6,16 @@
...
+/**
+ * Regular expression for the default report filename.
+ */
+export const REPORT_NAME_PATTERN =
+  /^(report\.)(\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}\.\d{3}Z)/;
+export function reportNameFromReport<T extends { date: string }>(
+  report: T,
+): string {
+  return `report.${report.date.replace(/:/g, '-')}`;
+}
```
matejchalk, line 18: The function name is pretty awkward, don't have to mention "report" twice. How about `reportFileName`?

packages/core/src/lib/upload.ts
```diff
@@ -14,25 +18,29 @@ export type UploadOptions = Pick<CoreConfig, 'upload' | 'persist'>;
...
+        () => [filename] satisfies FileResult,
+      );
+    }),
+  );
 
-  return uploadFn({ apiKey, server, data }).catch(e => {
-    throw new Error('upload failed. ' + e.message);
-  });
+  logMultipleFileResults(uploadResults, 'Uploaded reports');
+  return uploadResults;
```
matejchalk, line 45: OK, I have problems with this:

1. This is out of scope for this PR. What has it got to do with adding `--persist.filename`?
2. Do we even want this? I don't remember ever discussing it. I don't get why `upload` should be searching for report files instead of uploading the single report file specified in the config. To me that's really unexpected behaviour, and you could easily end up re-uploading reports if the output directory isn't cleaned up.

packages/models/src/lib/persist-config.ts
```diff
@@ -6,10 +6,10 @@ export type Format = z.infer<typeof formatSchema>;
 
 export const persistConfigSchema = z.object({
   outputDir: filePathSchema('Artifacts folder'),
-  filename: fileNameSchema('Artifacts file name (without extension)').default(
-    'report',
-  ),
-  format: z.array(formatSchema).default(['stdout']).optional(), // @TODO remove default or optional value and otherwise it will not set defaults
+  filename: fileNameSchema(
+    'Artifacts file name (without extension)',
+  ).optional(),
```
matejchalk, line 11: Really not sure this is a good change.

e2e/cli-e2e/mocks/fs.mock.ts
```diff
@@ -27,3 +28,15 @@ export function cleanFolderPutGitKeep<T extends object>(
     }
   }
 }
+
+export function setupFolder<T extends object>(
+  dirName = 'tmp',
+  content?: { [key in keyof T]: string },
+) {
```
matejchalk, line 35: There's no advantage to using generics here. It has no effect on any other parameter or on what is returned.

```suggestion
export function setupFolder(
  dirName = 'tmp',
  content?: Record<string, string>,
) {
```

matejchalk (changes requested):
Sorry, but this PR is doing way too much. Please reduce the scope. I don't want to merge all these changes.

- Sure, add `--persist.filename` as a CLI args also.
- Yes, use the timestamp to avoid multiple tests interacting with the same file.
- Everything else should be refined first.

packages/core/src/lib/upload.ts
BioPhoton: I set the default report file name back to 'report.json' and move a time stamp file into another issue.

packages/models/src/lib/implementation/utils.ts
BioPhoton: done

packages/utils/src/lib/utils.ts
BioPhoton: done

packages/core/src/lib/implementation/persist.ts
```diff
@@ -22,15 +21,13 @@ export class PersistError extends Error {
   }
 }
 
-export type PersistResult = PromiseSettledResult<readonly [string, number]>[];
-
 export async function persistReport(
   report: Report,
   config: CoreConfig,
-): Promise<PersistResult> {
+): Promise<MultipleFileResults> {
```
matejchalk, line 27: This is still handling multiple files. Please revert.

packages/core/src/lib/upload.spec.ts
```diff
@@ -34,18 +36,33 @@ vi.mock('fs/promises', async () => {
 });
 
 const outputDir = MEMFS_VOLUME;
-const reportPath = (path = outputDir, format: 'json' | 'md' = 'json') =>
-  join(path, 'report.' + format);
+const filename = reportFileName({ date: new Date().toISOString() });
```
matejchalk, line 39: Again, the timestamps in the filename are unnecessary here, these test have full control over the state of "file system" because they use `memfs`.

packages/models/src/lib/implementation/utils.spec.ts
```diff
@@ -63,6 +65,14 @@ describe('filenameRegex', () => {
   });
 });
 
+describe('reportNameFromReport', () => {
```
matejchalk, line 68:
```suggestion
describe('reportFileName', () => {
```

packages/models/src/lib/persist-config.ts
```diff
@@ -7,9 +7,9 @@ export type Format = z.infer<typeof formatSchema>;
 export const persistConfigSchema = z.object({
   outputDir: filePathSchema('Artifacts folder'),
   filename: fileNameSchema('Artifacts file name (without extension)').default(
-    'report',
+    'code-pushup.report',
```
matejchalk, line 10: Outside scope of this PR.

packages/utils/src/lib/report.ts
```diff
@@ -56,3 +64,47 @@ export function compareIssueSeverity(
...
+
+export async function loadReport(
+  options: Required<Pick<PersistConfig, 'outputDir' | 'filename'>> & {
+    format: Format;
+  },
+) {
+  const { outputDir, filename, format } = options;
+  await ensureDirectoryExists(outputDir);
+  const filePath = join(outputDir, `${filename}.${format}`);
+  const content = await readFile(filePath, 'utf8');
```
matejchalk, line 99: We already have a `readJsonFile` function which would make perfect sense here.

packages/core/src/lib/upload.ts
```diff
@@ -20,19 +18,19 @@ export async function upload(
...
     project,
     commit: await latestHash(),
-    ...jsonToGql(report),
+    ...jsonToGql(reportJson),
   };
 
-  return uploadFn({ apiKey, server, data }).catch(e => {
-    throw new Error('upload failed. ' + e.message);
-  });
+  return uploadFn({ apiKey, server, data });
```
matejchalk, line 35: If understand this change correctly, then you removed our validation of the report file (`JSON.parse` and then no `reportSchema.parse` => no safety) :warning: 

If you want to use a `loadReport` function, then I would expect that function returns an already validated `Report` instead of a plain `string`.

matejchalk (changes requested):

BioPhoton:
@matejchalk I suggest we switch in the future to timestamp or hash based filenames. If you agree I would keep the logic for filename generation in tests and the logic.

WDYT?

matejchalk:
> I suggest we switch in the future to timestamp or hash based filenames. If you agree I would keep the logic for filename generation in tests and the logic.
> 
> WDYT?

I think that `report.json` is a good and simple default for our end users. Being able to change the filename is something that's useful mainly for our tests - but only those that interact with the real file-system, tests using `memfs` can stick to the default, I believe.

Timestamp would be more useful than hash, I think, but I'm not sure it's a good idea at this point. And the idea with allowing placeholder syntax in the `persist.filename` option would then make sense, but there are unintended consequences I would prefer to avoid. E.g. the `upload` command would then have to resolve the filename dynamically and can have more than one match. But uploading multiple reports makes no sense at present and introduces unexpected behavior:
- the `upload` command takes the current commit as the ID it send to the portal, so multiple reports would get matched to same commit ID and thus the "multi-upload" will result in only one report (the last one if uploaded in sequence, who knows which one if uploaded in parallel :grimacing:)
- we don't prompt the user to clean their `persist.outputDir`, which could have intended consequences of re-uploaded the same report when running `collect` and `upload` multiple times

Also, I think anything other than changing adding `--persist.filename` and using it to fix test flakiness should be out of scope of this PR. Changing the default and changing the single upload to a multi-upload ... it changes way too much and I'm not sure it's for the better.

BioPhoton:
So lets move the comments into a new issue to discuss and I remove the testing logic.

packages/models/src/lib/persist-config.ts
BioPhoton: it defaults now

e2e/cli-e2e/tests/print-config.spec.ts
```diff
@@ -27,7 +28,11 @@ const execCli = (argObj: Partial<CliArgsObject>) =>
 
 describe('print-config', () => {
   it.each(extensions)('should load .%s config file', async ext => {
-    const { code, stderr, stdout } = await execCli({ config: configFile(ext) });
+    const reportFileName = filename();
```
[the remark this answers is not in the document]
BioPhoton: Good point. It would not test the default value. Was a left over from the unique filename logic. 

Fixed.

packages/core/src/lib/implementation/persist.ts
BioPhoton: It can save `json`, `md`, etc... And the new type is just more generic. So i did not change the logic.

Please help me to understand what exactly i should do.

matejchalk: Oh, I think I got confused :man_facepalming: Thanks for explaining, multiple formats of the same report is fine, of course :+1:

You can ignore my comment here, I mixed this up with `loadReports` which is scanning the output directory for multiple different reports, that's what should be reverted. But I already mentioned that here: https://github.com/flowup/quality-metrics-cli/pull/187#discussion_r1383153436

packages/utils/src/lib/utils.ts
```diff
@@ -51,7 +54,15 @@ export function distinct<T extends string | number | boolean>(array: T[]): T[] {
   return Array.from(new Set(array));
 }
 
-// === File
+// === Filesystem
+
+export async function ensureDirectoryExists(baseDir: string) {
+  try {
+    await stat(baseDir);
```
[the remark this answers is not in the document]
BioPhoton: tired is in a different way

packages/utils/src/lib/report.ts
BioPhoton: used it + the readTextFile

e2e/cli-e2e/tests/collect.spec.ts
```diff
@@ -23,58 +23,65 @@ describe('CLI collect', () => {
     });
 
   const cliPath = join('..', '..', 'dist', 'packages', 'cli');
+  const reportPath = join('tmp', 'react-todos-app');
```
BioPhoton, line 26:
```suggestion
  // @TODO use filename over outputDir
  const reportPath = join('tmp', 'react-todos-app');
```

BioPhoton:
Did my best to reduce scope from the refactoring. Next PR is on testing lib so this will get better.

matejchalk (commented):
> Did my best to reduce scope from the refactoring. Next PR is on testing lib so this will get better.

Thanks, it's a lot better now. Added some small code style comments (we have lint rules for the unnecessary `async`s and missing `await`s, BTW, just wasn't able to turn it on yet). But much happier with the scope now :slightly_smiling_face:

packages/utils/src/lib/utils.spec.ts
```diff
@@ -75,3 +98,84 @@ describe('distinct', () => {
...
+  beforeEach(() => {
+    resetFiles();
+  });
+
+  it('should create folder', async () => {
+    resetFiles();
+    await expect(
+      ensureDirectoryExists(join(outputDir, 'sub', 'dir')),
+    ).resolves.toEqual(void 0);
+  });
```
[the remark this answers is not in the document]
BioPhoton: Thanks!

packages/utils/src/lib/utils.spec.ts
```diff
@@ -75,3 +98,84 @@ describe('distinct', () => {
     ]);
   });
 });
+
+describe('ensureDirectoryExists', () => {
+  beforeEach(() => {
+    resetFiles();
+  });
```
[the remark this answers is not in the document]
BioPhoton: addressed

packages/core/src/lib/implementation/persist.ts
```diff
@@ -86,34 +84,6 @@ export async function persistReport(
...
-    console.log(`Generated reports failed: `);
-    failedPersistedResults.forEach(result => {
-      console.log(`- ${chalk.bold(result.reason)}`);
-    });
-  }
-}
-
-function reportName(report: Report): string {
-  return `report.${report.date}`;
+export async function logPersistedResults(persistResults: MultipleFileResults) {
```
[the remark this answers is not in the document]
BioPhoton: Good catch!

matejchalk (approved):

PR #194, opened 2023-11-01
ci: adds `maxWarnings=0` as default for the lint target

### LINKED ISSUE #138: Align linting in CI and on client
In the CI there is an additional option `--max-warnings=0`. 
If we set it as default we catch those issues early.

Add target default to all lint targets. 

Also the spec files are respected on the server but not client

https://github.com/flowup/quality-metrics-cli/actions/runs/6679072872/job/18150781955?pr=139

IKatsuba:
Closes #138

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Install dependencies
         run: npm ci
       - name: Lint affected projects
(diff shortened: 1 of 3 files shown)


matejchalk (approved):
Works like a charm, thank you :slightly_smiling_face: 

Small nitpick: Please use imperative mood for commit messages ("add" instead of "adds") :pray:

Issue #197, opened 2023-11-02
Upload CLI's report to portal in CI


matejchalk:
## Motivation

- The best way to test our product is to use it ourselves!
- A lot of lint errors were introduced by stricter configuration in #198.
- We should use Code PushUp to track our progress in improving the codebase to see how useful it is.

## Scope

- [ ] create a `code-pushup.config.ts` for the CLI codebase
- [ ] configure ESLint plugin to execute linter for all Nx projects
- [ ] temporarily disabled lint rules will be included (we want to collect those errors)
- [ ] in CI workflow, run `nx build cli` and then execute as `npx dist/packages/cli`
- [ ] collect report and store as artifact in GitHub Actions
- [ ] upload report to portal's staging API (use GitHub secret for API key)

PR #198, opened 2023-11-02
build: add strict lint rules, temporarily disable failing

### LINKED ISSUE #29: Setup lint rune for uncaught Promise's
In the CLI lib we lost time when debugging because of that. A linting rule would avoid that.

matejchalk:
Closes #29 

- Researched ESLint plugins available in the ecosystem ([awesome-eslint](https://github.com/dustinspecker/awesome-eslint) is a great resource) and integrated plugins which I think will be useful for this codebase.
- Added many [ESLint plugins and rules used in portal codebase](https://github.com/flowup/quality-metrics-portal/blob/main/.eslintrc.json) (and other FlowUp projects), which I thought were applicable for the CLI (e.g. no Angular or RxJS plugins, obviously).
- Currently, with the new stricter rules we now have a large amount of lint errors and warnings in this codebase. 
  - I explicitly disabled each of those rules for now.
  - We should fix them, but we will do that incrementally.
  - **This is a great opportunity to try out our own product**. I will follow-up with #197, so that we can use the Code PushUp portal to track our own progress.
- Scope of this PR is to agree on the proposed rules as an end goal.

Seriously, **a lot of rules** were added, so I won't describe them all here (we can always discuss specifics). A good way to test the new configuration is to comment out the rules I had to disable in [`.eslintrc.json`](https://github.com/flowup/quality-metrics-cli/blob/stricter-linting/.eslintrc.json#L267-L330) and run the `nx lint` target to see what gets flagged up.

- ESLint's [core rules](https://eslint.org/docs/latest/rules/) (recommended config + some extras)
- [`@typescript-eslint` plugin](https://typescript-eslint.io/) (recommended and strict configs with type checking + some extras)
- [`import` plugin](https://github.com/import-js/eslint-plugin-import#readme) (recommended config + some extras)
- [`sonarjs` plugin](https://github.com/SonarSource/eslint-plugin-sonarjs#readme) (recommended config)
- [`unicorn` plugin](https://github.com/sindresorhus/eslint-plugin-unicorn#readme) (recommended config, but I decided to soften many of those rules from errors to warnings)
- [`vitest` plugin](https://github.com/veritem/eslint-plugin-vitest#readme) (recommended config + some extras - only for test files, obviously)
- [`functional` plugin](https://github.com/eslint-functional/eslint-plugin-functional#readme) (only hand-picked rules, it's very strict)
- [`deprecation` plugin](https://github.com/gund/eslint-plugin-deprecation#readme) (just the 1 rule)
- [`promise` plugin](https://github.com/eslint-community/eslint-plugin-promise#readme) (recommended config, no tweaking was needed here)
- [`no-secrets` plugin](https://github.com/nickdeis/eslint-plugin-no-secrets#readme) (just the 1 rule with some config to prevent false positives)


Tlacenka (approved):
LF to the `vitest` good practices suggestions!

PR #199, opened 2023-11-02
CLI e2e tests

### LINKED ISSUE #135: Add e2e tests for all commands
Some package CLI and CLI e2e are missing e2e tests. 

- document testing scope of e2e project 
- add missing tests if needed

IKatsuba:
Closes #135 

Added tests for the `autorun` and `upload` commands. There are no server tests. I created the `createMockServer` function to be sure what report was sent.

I added a simple snapshot test for the help command.

Another command has good coverage.


matejchalk:
It is unfortunate, but I agree that this PR shouldn't be merged. To be clear, it's not @IKatsuba's fault, the issue itself was misguided and it's a shame this wasn't caught earlier.

E2E tests should be a well-defined suite of tests, focused only on critical paths. There should be few of them and they should be based on user flows. As I understand this issue (the description doesn't explain much), the goal was to "add E2E tests for CLI commands that aren't covered". It's a mistake to think of E2E tests that way, I believe. 100% coverage isn't the goal, E2E testing everything adds a big maintenance burden.

Another conceptual problem is that upload tests aren't (and in fact cannot be) end-to-end, because the core functionality relies on a portal package which the CLI doesn't control. Mocking the server is correct, but that means it's not an end-to-end test. We already have extremely similar tests in [packages/core/src/lib/upload.spec.ts](https://github.com/flowup/quality-metrics-cli/blob/main/packages/core/src/lib/upload.spec.ts) and [packages/cli/src/lib/upload/command-object.spec.ts](https://github.com/flowup/quality-metrics-cli/blob/main/packages/cli/src/lib/upload/command-object.spec.ts), I don't think adding a 3rd variant (with much more complex mocking) is beneficial. 

I do like that some tests have been added for error cases, but I think E2E tests are the wrong place to put those. E2E tests should be focused on happy paths, integration tests or unit tests are a better fit for testing edge cases (e.g. `upload/command-object.spec.ts` would be a good place for these).

PR #201, opened 2023-11-03
feat(plugin-eslint): create groups from rules' metadata

### LINKED ISSUE #200: Create groups in ESLint plugin
As user, I would like to have an easier way to connect ESLint audits to categories than one-by-one.

There are couple of ways we can automatically group lint rules based on some their metadata:
- ESLint rules may have a `meta.type` of type `'problem' | 'suggestion' | 'layout'` (not every ESLint plugin uses it, but many do)
  - we can automatically create `problems`, `suggestions` and `layout` groups from these
- some ESLint plugins take advantage of `meta.docs.category`, although this is deprecated and can be any string, so it'll likely only apply to rules within a specific plugin
  - we can automatically create groups like `functional-no-mutations` and `functional-no-other-paradigms` (example for `eslint-plugin-functional`), which could also be useful in some cases

matejchalk:
Closes #200

### DIFF
--- a/e2e/cli-e2e/tests/__snapshots__/collect.spec.ts.snap
+++ b/e2e/cli-e2e/tests/__snapshots__/collect.spec.ts.snap
@@ -993,6 +993,228 @@ exports[`CLI collect > should run ESLint plugin and create report.json 1`] = `
         },
       ],
       "description": "Official Code PushUp ESLint plugin",
+      "groups": [
+        {
+          "description": "Code that either will cause an error or may cause confusing behavior. Developers should consider this a high priority to resolve.",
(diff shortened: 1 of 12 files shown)


BioPhoton (approved):
Valuable change. We should add this to the eslint plugin docs.

matejchalk:
> Valuable change. We should add this to the eslint plugin docs.

Good call, just added it :grin:

MishaSeredenkoPushBased (approved):

Tlacenka (approved):
Great DX improvement!

Props for updated tests, E2E and documentation.

PR #202, opened 2023-11-06
chore: apply different commitlint plugin for imperative tense

Tlacenka:
I researched different plugins for commitlint as the current one produced false positives.
The `commitlint-plugin-tense` has the `firstOnly` option and a better recognition of nouns and adjectives that sound like verbs.
Docs [here](https://github.com/actuallydamo/commitlint-plugin-tense).

## Past tense examples
image

## Present continuous examples
image


matejchalk (approved):
I love it! :heart_eyes: 

Looks much more robust than the plugin I added in #195 :sweat_smile:

PR #206, opened 2023-11-06
feat(utils): implement a new design for stdout

### LINKED ISSUE #190: Implement stdout report formatting
Follow-up to:
- #151 

PoC:
- #189
- implementation in [tools/pocs/stdout-report-mock.mjs](https://github.com/flowup/quality-metrics-cli/blob/stdout-poc/tools/pocs/stdout-report-mock.mjs)

### Structure

For visual details (alignment, spacing, colors, etc.), refer to PoC implementation linked above.

- [ ] for each plugin:
  - [ ] plugin title as heading (with _audits_ suffix)
  - [ ] list plugin's audits:
    - [ ] `●` colored red/yellow/green based on audit's score (thresholds are 50 and 90, same as for #148)
    - [ ] audit title
    - [ ] audit display value (or raw value if missing) - highlighted (bright cyan)
- [ ] _Categories_ heading
- [ ] table with each category:
  - [ ] category title
  - [ ] formatted score (0-100) in bold with red/yellow/green coloring
  - [ ] total number of audits (from category refs with "expanded" groups - same as #148)

### Implementation notes

- [chalk](https://www.npmjs.com/package/chalk) for colors and bold formatting
- [cliui](https://www.npmjs.com/package/cliui) for columns (audits)
- [cli-table3](https://www.npmjs.com/package/cli-table3) for table (categories)

### Design

Image

IKatsuba:
Closes #190

### DIFF
--- a/packages/core/src/lib/implementation/persist.spec.ts
+++ b/packages/core/src/lib/implementation/persist.spec.ts
@@ -76,7 +76,7 @@ describe('persistReport', () => {
 
   it('should stdout as format by default`', async () => {
     await persistReport(dummyReport, dummyConfig);
-    expect(logs).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
+    expect(logs.join('\n')).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
 
     expect(() => readReport('json')).not.toThrow();
     expect(() => readReport('md')).toThrow('no such file or directory');
@@ -89,7 +89,7 @@ describe('persistReport', () => {
       ...dummyConfig,
       persist,
     });
-    expect(logs).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
+    expect(logs.join('\n')).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
 
     expect(() => readReport('json')).not.toThrow('no such file or directory');
     expect(() => readReport('md')).toThrow('no such file or directory');
@@ -143,7 +143,7 @@ describe('persistReport', () => {
       `${FOOTER_PREFIX} [Code PushUp](${README_LINK})`,
     );
 
-    expect(logs).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
+    expect(logs.join('\n')).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
   });
 
   it('should persist some formats`', async () => {
@@ -162,7 +162,7 @@ describe('persistReport', () => {
       `${FOOTER_PREFIX} [Code PushUp](${README_LINK})`,
     );
 
-    expect(logs).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
+    expect(logs.join('\n')).toMatch(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
   });
 
   // @TODO: should throw PersistDirError
--- a/packages/core/src/lib/implementation/persist.spec.ts
+++ b/packages/core/src/lib/implementation/persist.spec.ts
@@ -12,6 +12,7 @@ import {
 import {
   CODE_PUSHUP_DOMAIN,
   FOOTER_PREFIX,
+  NEW_LINE,
   README_LINK,
 } from '@code-pushup/utils';
 import { mockConsole, unmockConsole } from '../../../test';
@@ -76,7 +77,9 @@ describe('persistReport', () => {
 
   it('should stdout as format by default`', async () => {
     await persistReport(dummyReport, dummyConfig);
-    expect(logs.join('\n')).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
+    expect(logs.join(NEW_LINE)).toContain(
+      `${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`,
+    );
 
     expect(() => readReport('json')).not.toThrow();
     expect(() => readReport('md')).toThrow('no such file or directory');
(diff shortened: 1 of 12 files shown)


matejchalk (changes requested):
I tested it out for our example app (`npx nx run cli:run-collect`), added a few visual notes. I'm not sure why the bold formatting isn't working properly in the audits section, since the code is the same as my PoC :confused: 

I got the feeling `cliui` was quite buggy in this, so maybe we could also consider some alternative package (maybe [table-layout](https://www.npmjs.com/package/table-layout)?) or even write it ourselves (all we need is the column alignment and word wrapping).

image

IKatsuba:
> I tested it out for our example app (`npx nx run cli:run-collect`) and added a few visual notes. I'm not sure why the bold formatting isn't working correctly in the audits section since the code is the same as my PoC 😕
> 
> I got the feeling `cloud` was quite buggy in this, so maybe we could also consider some alternative package (maybe [table-layout](https://www.npmjs.com/package/table-layout)?) or even write it ourselves (all we need is the column alignment and word wrapping).

All CLI packages do strange things. I left the current packages but rendered the full report one time.

Do not use `nx run-collect` if you want to test it. It removes new line symbols with no reason. Just run 
```
npx ../../dist/packages/cli collect --persist.format=stdout --persist.format=stdout
```
in `examples/react-todos-app.`

Tlacenka (commented):
I ran it on Windows in VSCode with light theme. Looking good! 🌈 

## In-scope comment
- While a preceding empty line is nice for the section, it should visually belong to the text below so I wouldn't put a new line right after it. In the original design, the preceding line is bigger than the gap after the title as well.

## OOS
- Is there any way of wrapping words to a new line so that they aren't cut off like that?
   - To be continued in #209 
- The footer line is missing a whitespace (has been ever since I joined) and I would really like to see that fixed. → Won't fix atm.
- This is a suggestion for both MD and stdout - can we sort the output e.g. based on number of errros, then warnings? The user now has to skim through it seemingly with no logical ordering.
   - To be continued in #210 

image

matejchalk (approved):
Thanks for the fixes, the output looks really good now :+1: 

I have suggestions regarding code style. But the functionality itself is now working as I would expect and the sanitized snapshot is now much more easy to understand as well.

matejchalk:
@Tlacenka While I agree with your comments, I think changes to the original design should handled as another issue, so we don't bloat this PR. These changes are already a big improvement over the formatting we have in `main`, in my opinion.

> Is there any way of wrapping words to a new line so that they aren't cut off like that?

Yes, that bugs me as well :grimacing: 

Unfortunately, it looks like `cliui` doesn't handle it properly because of some complicated ESM vs CJS stuff, I came across [this issue](https://github.com/yargs/cliui/issues/89). So we may have to switch libraries or implement the inteligent word wrapping ourselves.

But I do agree it would improve the output a lot, many of the audit titles won't fit on one line, so it's not really an edge case.

> The footer line is missing a whitespace (has been ever since I joined) and I would really like to see that fixed.

This is a weird one. The space is actually there, it's just that some terminals don't render it after the heart. E.g. my VSCode integrated terminal renders it, but my regular Linux terminal doesn't (same as your Windows terminal) :confused: 

> This is a suggestion for both MD and stdout - can we sort the output e.g. based on number of errros, then warnings? The user now has to skim through it seemingly with no logical ordering.

Yes, I can see that would be useful. Could you create a separate issue for that? I'm not sure what the best ordering would be (could be weight, score, alphabetical :man_shrugging:), that's something we can discuss during refinement.

> While a preceding empty line is nice for the section, it should visually belong to the text below so I wouldn't put a new line right after it. In the original design, the preceding line is bigger than the gap after the title as well.

Good catch :+1:  In my PoC, I put an extra blank line before the title (so it's 2 blank lines before and 1 after). That would be an easy improvement, I believe.

packages/utils/src/lib/report-to-stdout.ts
```diff
@@ -1,108 +1,105 @@
...
+  reportRawOverviewTableHeaders,
+} from './utils';
 
-const ui = cliui({ width: 60 }); // @TODO check display width
+let output = '';
+
+const addLine = (text = '') => (output += text + '\n');
+const print = (text: string) => console.log(text);
 
 export function reportToStdout(report: ScoredReport): void {
```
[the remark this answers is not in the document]
IKatsuba: I love this idea! Thank you!

IKatsuba:
I added the extra line before the titles. I agree; this is more readable.

Did small refactoring. `reportToStdout` now returns `string`.

About whitespace after ❤️. I tried to change to another symbol. 
image

@Tlacenka, could you check it out?

matejchalk:
> About whitespace after ❤️. I tried to change to another symbol.

For me that did the trick :+1: 

image

Before it rendered like this:

image

Tlacenka (dismissed):
All defined acceptance criteria is met based on the original design now. 💯 

The preceding gap for a section title is now bigger. ✔️ 

Unfortunately, the heart emoji still does not render a whitespace after it. It is now a different emoji than before (black and white) but the space is still missing. 😢 

image

matejchalk (dismissed):

packages/core/src/lib/implementation/persist.spec.ts
```diff
@@ -76,7 +76,7 @@ describe('persistReport', () => {
 
   it('should stdout as format by default`', async () => {
     await persistReport(dummyReport, dummyConfig);
-    expect(logs).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
+    expect(logs.join('\n')).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
```
BioPhoton, line 79: Can we use the const `NEW_LINE` here please

packages/core/src/lib/implementation/persist.spec.ts
```diff
@@ -162,7 +162,7 @@ describe('persistReport', () => {
       `${FOOTER_PREFIX} [Code PushUp](${README_LINK})`,
     );
 
-    expect(logs).toContain(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
+    expect(logs.join('\n')).toMatch(`${FOOTER_PREFIX} ${CODE_PUSHUP_DOMAIN}`);
```
BioPhoton, line 165: same as above, use the `NEW_LINE` constant please

packages/utils/src/lib/report-to-stdout.spec.ts
```diff
@@ -1,23 +1,12 @@
...
-    unmockConsole();
-  });
-
   it('should contain all sections when using the fixture report', () => {
-    reportToStdout(scoreReport(minimalReport('tmp')));
-    const logOutput = logs.join('\n');
-    expect(logOutput).toMatchSnapshot();
+    const logOutput = reportToStdout(scoreReport(report()));
+    // eslint-disable-next-line no-control-regex
+    expect(logOutput.replace(/\u001B\[\d+m/g, '')).toMatchSnapshot();
```
BioPhoton, line 10: please add a comment what ´/\u001B\[\d+m/g´ is doing 🙏

BioPhoton (approved):
<3

PR #207, opened 2023-11-07
test: fix and apply vitest ESLint rules

### LINKED ISSUE #204: Apply all added ESLint rules to vitest
Currently, some rules are temporarily turned off. The aim of this issue is to fix all related ESLint issues, so that the rules can permanently stay on:
- [x] `vitest/prefer-to-be`
- [x] `vitest/consistent-test-it`
- [x] `vitest/prefer-to-have-length`
- [x] `vitest/no-alias-methods`
- [x] `vitest/no-conditional-tests`
- [x] `vitest/require-hook`
- [x] `vitest/require-to-throw-message`

Tlacenka:
Closes #204 

Updated Code PushUp scores
image

### DIFF
--- a/packages/cli/src/lib/collect/command-object.spec.ts
+++ b/packages/cli/src/lib/collect/command-object.spec.ts
@@ -30,9 +30,10 @@ const cli = (args: string[]) =>
     ...DEFAULT_CLI_CONFIGURATION,
     commands: [yargsCollectCommandObject()],
   });
-let logs;
 
 describe('collect-command-object', () => {
+  let logs: unknown[];
+
   beforeEach(() => {
     logs = [];
     cleanFolderPutGitKeep();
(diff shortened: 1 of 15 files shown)


packages/cli/src/lib/collect/command-object.spec.ts
```diff
@@ -30,18 +30,18 @@ const cli = (args: string[]) =>
     ...DEFAULT_CLI_CONFIGURATION,
     commands: [yargsCollectCommandObject()],
   });
-let logs = [];
+let logs;
 
 describe('collect-command-object', () => {
   beforeEach(() => {
+    logs = [];
```
matejchalk, line 37: I would also scope this variable to the `describe` block and add some type annotation (otherwise it's implicitly `any`).

```suggestion
describe('collect-command-object', () => {
  let logs: unknown[];

  beforeEach(() => {
    logs = [];
```

matejchalk (dismissed):
Nice improvement, looking forward to seeing the impact in the reports chart :slightly_smiling_face:

packages/cli/src/lib/collect/command-object.spec.ts
Tlacenka: Problem is, in some tests this is done in some `setup` and `teardown` function which uses this variable.... but in those cases where it isn't, I can do that yes.

matejchalk (approved):

PR #208, opened 2023-11-07
ci: run code-pushup collect and upload in CI workflow

### LINKED ISSUE #197: Upload CLI's report to portal in CI

matejchalk:
Closes #197 

The configuration for the ESLint plugin turned out to be very complex when targetting the whole monorepo, which is why I had to do some complex scripting in `code-pushup.config.ts` to get it working :grimacing: I'll consider some ways this can be simplified for similar setups (e.g. provide some Nx helpers).

But anyway, it works! :tada: After this is merged, all our CI runs will upload new report to the portal in staging :raised_hands: 

The `report.json` and `report.md` files are also uploaded as an artifact to GitHub Actions (e.g. [here](https://github.com/flowup/quality-metrics-cli/actions/runs/6786622704#artifacts)).

image

image

image


Tlacenka (changes requested):
Great work with a lot of potential! Checking ourselves with our own product 😁 

Unfortunately, when I checked the artifacts of the PR pipeline, I was unable to find any ESLint errors regarding `vitest` ESLint plugins. From what I understood, you removed the temporarily removed rules in order to show the current state but somehow the test files were not scanned.

matejchalk:
> Unfortunately, when I checked the artifacts of the PR pipeline, I was unable to find any ESLint errors regarding `vitest` ESLint plugins. From what I understood, you removed the temporarily removed rules in order to show the current state but somehow the test files were not scanned.

Yes, great catch, that was not intentional :sweat_smile: Not easy to fix either, but I found a way. Now they're in :tada: 

image

Tlacenka (approved):
In the latest artifact, test errors correspond to what I have been fixing. Good job!

PR #212, opened 2023-11-09
perf(utils): improve the performance of scoring and reporting

### LINKED ISSUE #132: Make scoring and reporting logic more performant
The current scoring and reporting logic is not written in a performance way.
- The logic for groups and plugins is different from groups and audits. If we unify the data structures it would simplify the code
  - We can score groups at the beginning
  - We can add the plugin to `slug` to audits and groups to avoid a nested loop
- For scoring the categories could introduce a dict with keys like `${plugin}${type}${slug}` to reduce the iterations
- In `categoryRefToScore` and `groupRefToScore` we use `find` to get the `score` or a specific `audit` or `group`. This is done for each ref in a category. If we intro a dictionaries for plugins we can avoid it.
To improve the performance I suggest to do the following things:
- We should introduce a function that scores a report one time so ti can be reused for the different report formats

**Update:**  

After the recent changes we decided to remove the plugin in audits in `ScoredReport` and go with a dic of audits under plugins where key a combination of type, plugin,  audit/group. 

related to #137 and #133

MishaSeredenkoPushBased:
Improved the performance of scoring and reporting.
Removed nested loops where possible, added dictionaries to go constant time O(1) instead of linear O(n).
Made minor refactoring, added some checks.
@BioPhoton As task owner, let me know if I understood the task correctly and implemented everything properly. Also, maybe you could propose additional changes for further performance improvements.

Closes #132


BioPhoton:
Could you add the new logic the the benchmark tests please.

packages/core/src/lib/implementation/persist.ts
```diff
@@ -32,11 +32,13 @@ export async function persistReport(
...
   const outputDir = persist.outputDir;
   const filename = persist?.filename || reportName(report);
-  let { format } = persist;
-  format = format && format.length !== 0 ? format : ['stdout'];
-  let scoredReport;
+  const format =
+    persist.format && persist.format.length !== 0 ? persist.format : ['stdout'];
+
+  // Score the report once at the start
+  const scoredReport = scoreReport(report);
```
BioPhoton, line 39: If the formats are only json we do not need to score it. This is the reason I had is in each format case. This is more efficient and does not cause too much noise in the code i think.

BioPhoton (commented):
Looks good! 
One thing would be good to add is the new benchmark like here: https://github.com/flowup/quality-metrics-cli/pull/137

packages/utils/src/lib/report.ts
```diff
@@ -57,16 +57,26 @@ export function countCategoryAudits(
...
+    if (plugin.groups) {
+      lookup[plugin.slug] = plugin.groups.reduce<
+        Record<string, ScoredReport['plugins'][0]['groups'][0]>
+      >((groupLookup, group) => {
+        groupLookup[group.slug] = group;
+        return groupLookup;
+      }, {});
+    }
+    return lookup;
+  }, {});
```
matejchalk, line 73: This code is mixing two paradigms, it would make sense to use one or the other, otherwise they negate the benefits of each other:

## 1 - Functional (declarative, immutable) paradigm

In my opinion, this should be favoured by default because the code is easier to understand and maintain (high-level, no side-effects, etc).

Using `reduce` is very common for functional code because it's a declarative way of aggregating an array into a single value. Although JavaScript cannot enforce it, mutating inside of `reduce` goes against the intended pattern. Specifically for accumulator, the contract is that you return a new value, instead of mutating it directly. The most common way to immutably create new values of objects (and arrays) is to use the spread operator:

```suggestion
  const groupLookup = plugins.reduce<
    Record<string, Record<string, ScoredReport['plugins'][0]['groups'][0]>>
  >((lookup, plugin) => {
    if (!plugin.groups.length) {
      return lookup;
    }
    return {
       ...lookup,
       [plugin.slug]: plugin.groups.reduce<Record<string, AuditGroup>>(
          (groupLookup, group) => ({ ...groupLookup, [group.slug]: group }),
          {},
       )
    };
  }, {});
```

## 2 - Imperative (low-level, mutable) paradigm

This is generally more verbose, harder to debug (intermediate states), etc. That makes often a worse choice for readability and maintenance, but it tends to have a slight performance advantage. So I would say the rule of thumb to only use when the code is known to be a performance bottleneck (e.g. runs extremely often, or handles a lot of data).

If this code would benefit from performance optimization (this should be based on some measurement, ideally), then the imperative approach could make sense. But in that case, there are more efficient ways to iterative over arrays than `.reduce`:

```suggestion
  const groupLookup: Record<string, Record<string, AuditGroup>> = {};
  for (const plugin of plugins) {
    if (!plugin.groups.length) {
      continue;
    }
    groupsLookup[plugin.slug] ??= {};
    for (const group of plugin.groups) {
       groupsLookup[plugin.slug][group.slug] = group;       
    }
  }
```

packages/utils/src/lib/scoring.ts
```diff
@@ -42,31 +42,45 @@ function categoryRefToScore(
...
+    {},
+  );
+
+  const groupLookup = groups.reduce<Record<string, EnrichedScoredAuditGroup>>(
+    (lookup, group) => {
+      lookup[`${group.plugin}-${group.slug}`] = group;
+      return lookup;
+    },
+    {},
+  );
```
matejchalk, line 60: Same as before:
- should probably be `reduce` and no mutation, or
- if it has a performance impact, keep the mutation, but use `for` or `.forEach()`

packages/utils/src/lib/scoring.ts
```diff
@@ -77,27 +91,39 @@ export function calculateScore<T extends { weight: number }>(
...
-    (sum, ref) => sum + scoreFn(ref) * ref.weight,
-    0,
+  const { numerator, denominator } = refs.reduce(
+    (acc, ref) => {
+      acc.numerator += scoreFn(ref) * ref.weight;
+      acc.denominator += ref.weight;
+      return acc;
+    },
+    { numerator: 0, denominator: 0 },
   );
```
matejchalk, line 101: Same here, please don't mutate the accumulator in `reduce` :pray: 

```suggestion
  const { numerator, denominator } = refs.reduce(
    (acc, ref) => ({
      ...acc,
      numerator: acc.numerator + scoreFn(ref) * ref.weight,
      denominator: acc.denominator + ref.weight,
    }),
    { numerator: 0, denominator: 0 },
  );
```

packages/utils/src/lib/scoring.ts
```diff
@@ -77,27 +91,39 @@ export function calculateScore<T extends { weight: number }>(
...
-      })) || [];
+      groups?.map(group => {
+        const scoredGroup = {
+          ...group,
+          score: calculateScore(group.refs, groupRefToScore(preparedAudits)),
+          plugin: plugin.slug,
+        };
+        allScoredGroups.push(scoredGroup);
+        return scoredGroup;
+      }) || [];
```
matejchalk, line 126: Similar issue as before, but slightly different scenarios. Similar to `reduce`, `map` is intended to be used in an FP-style. Specifically, it should only return new array items, side-effects in `map` are unexpected (in this case, mutating arrays from outer scope). 

(Immutable/declarative methods are `map`, `filter`, `reduce`, `slice`, etc. While imperative/mutable methods are e.g. `forEach`, `sort`, `reverse`, `splice`.)

packages/utils/src/lib/scoring.ts
BioPhoton: I bed mutating in for each is faster. lets see the benchmarks

packages/utils/src/lib/report.ts
BioPhoton: Very nice and detailed feedback!! 

Thank you a LOT for your time! <3

packages/core/src/lib/implementation/persist.ts
MishaSeredenkoPushBased: Agree, reverted that thing

packages/core/src/lib/implementation/persist.ts
```diff
@@ -47,15 +49,18 @@ export async function persistReport(
...
+      if (!commitData) {
+        console.warn('no commit data available');
+      }
+      results.push({
+        format: 'md',
+        content: reportToMd(scoredReport, commitData),
+      });
+    } catch (e) {
+      console.warn(e);
     }
```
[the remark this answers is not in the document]
MishaSeredenkoPushBased: Sure, that's my fault, added for a check. Removed

packages/utils/src/lib/report.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { describe, expect } from 'vitest';
-import { CategoryRef } from '@code-pushup/models';
```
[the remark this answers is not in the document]
MishaSeredenkoPushBased: Checked for the unused imports, looks fine :)

packages/utils/src/lib/report.ts
```diff
@@ -57,16 +57,26 @@ export function countCategoryAudits(
   refs: CategoryRef[],
   plugins: ScoredReport['plugins'],
 ): number {
+  // Create lookup object for groups within each plugin
+  const groupLookup = plugins.reduce<
+    Record<string, Record<string, ScoredReport['plugins'][0]['groups'][0]>>
```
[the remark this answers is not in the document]
MishaSeredenkoPushBased: Thx, used your proposition

packages/utils/src/lib/report.ts
MishaSeredenkoPushBased: @matejchalk wow, that is very wide coverage of the issue, and one of the best comments I've ever seen! 😮 
Thank you for the description of both paradigms, you're right that they are mixed in my case. My fix will include declarative paradigm only, here and in another places (for this PR) ❤️

packages/utils/src/lib/scoring.ts
MishaSeredenkoPushBased: I'll keep the declarative one, but will check both in benchmark

packages/utils/src/lib/scoring.ts
MishaSeredenkoPushBased: image
Gone fully declarative ❤️

packages/utils/src/lib/scoring.ts
MishaSeredenkoPushBased: I totally agree that using the imperative approach inside the declarative is not good. Rewrote the function, it became a bit longer, but is now fully declarative, with no mutation.

MishaSeredenkoPushBased:
@BioPhoton @matejchalk  well, it looks like the changes don't affect the performance in a positive way - I added the updated functions to the benchmark, file `utils/perf/optimized3.mjs`. The performance went 3 times slower than before, and even slower than base:
image
We definitely need to try the imperative + dictionaries. But it seems like the dictionaries filling also costs some performance.
Any other suggestions? 🤔

BioPhoton:
could we get the traces for that?

MishaSeredenkoPushBased:
I checked my changes by benchmarking - it appeared that it got 535.000 ops/s. Really bad I would say, even worse than base. Then, by reverting my changes one function by one, I figured out that the initial performance for the scoring was 750.000 - looks better, but not so cool. There are different other optimized solutions in the benchmark, which gain up to 1.600.000 (on my machine).
It looks like the performance is affected by the declarative approach, multiple spread operators, .map, .reduce and so on. I tried to go ultimately imperative and see what the difference could be.
I took the `optimized2` solution from benchmark, fixed it (it didn't work properly), and optimized it more. I put it into `optimized3` check, and it gained almost 1.800.000 ops/s!
The only thing is that the code it highly imperative - a lot of mutations, nested functions and so on. Looks definitely not perfect, I expect that someone could leave the comments where it could be improved. But overall, I consider that this huge 3.5 times performance improvement is more important than prettiness or higher readability.
@BioPhoton @matejchalk FYI

matejchalk (changes requested):
A have a question about the benchmarks. What numbers of audits and groups are you using?

The fact that imperative code is faster the declarative is not surprising (that's always the case in benchmarks which use extreme values), but the trick is figuring out if it's relevant to us or not.

If the performance optimizations are not noticeable for realistic amounts of data, then optimizing for code maintenance is still the way to go. E.g. if it doesn't make a difference until we reach thousands of groups, then it's not relevant to us, because that's not a realistic scenario and we would do better to prioritize our ability to understand and maintain the code and prevent bugs (side-effects like mutations are an anthesis to this). There's a famous quote that "premature optimization is the root of all evil", after all.

I would estimate that 100s and in extreme cases 1000s of audits are realistic (e.g. for the CLI, our very strict ESLint gives us 347 audits). So if the performance optimization has an impact at this level, then it's worth it, but if we don't see the performance benefits until we reach a million audits for example, then it isn't.

For groups, the number is much lower. I can't imagine we'd have over a 100 even.

packages/utils/src/lib/scoring.ts
```diff
@@ -24,103 +24,88 @@ export type ScoredReport = Omit<Report, 'plugins' | 'categories'> & {
...
+    const item = allScoredAuditsAndGroups.get(key);
+    if (!item) {
+      throw new Error(
+        `Category has invalid ref - ${ref.type} with slug ${key} not found in ${ref.plugin} plugin`,
+      );
+    }
+    return item.score;
+  }
+
+  const scoredCategoriesMap = report.categories.reduce(
```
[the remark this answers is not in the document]
vmasek: I agree, even though `reduce` is nicer, more readable, no-sideffects, ... 😄 

We can just stick to `for` (not even `forEach`) plain `for`, iterate fast and optimized inside and have a nice functional API.

why not even `forEach`? because it creates copy of each of the iterated items for passing as parameter (same as `map`, `filter`, `reduce`), and that is unnecessary overhead.

e2e/cli-e2e/tests/__snapshots__/collect.spec.ts.snap
```diff
@@ -200,6 +202,7 @@ exports[`CLI collect > should run ESLint plugin and create report.json 1`] = `
             "issues": [],
           },
           "docsUrl": "https://eslint.org/docs/latest/rules/no-cond-assign",
+          "plugin": "eslint",
```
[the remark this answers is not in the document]
MishaSeredenkoPushBased: Fixed, the snapshot is reverted.

packages/utils/src/lib/transformation.spec.ts
```diff
@@ -96,3 +97,29 @@ describe('distinct', () => {
...
+    expect(cloned).not.toBe(obj);
+    expect(cloned.c).toEqual(obj.c);
+    expect(cloned.c).not.toBe(obj.c);
+    expect(cloned.c[0]).toEqual(obj.c[0]);
+    expect(cloned.c[0]).not.toBe(obj.c[0]);
+    expect(cloned.c[1]).toEqual(obj.c[1]);
+    expect(cloned.c[1]).not.toBe(obj.c[1]);
+    expect(cloned.d).toBe(obj.d);
+    expect(cloned.e).toBe(obj.e);
+  });
```
matejchalk, line 124: Nice test :+1:

matejchalk (approved):

packages/utils/src/lib/scoring.ts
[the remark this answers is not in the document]
MishaSeredenkoPushBased: Changed to for...of, performance is improved

MishaSeredenkoPushBased:
Overall, as a comment on the changes - the final performance improvement is 2.3 times in comparison to base solution, with 1000 audits and 100 groups.
image
The more audits and groups are present, the bigger is the difference. Here are crazy 10.000 of audits and 100 groups:
image
Almost 3 times difference - looks like it's a really performant solution.

Issue #92, opened 2023-10-10
Enhance DX of plugin authors


BioPhoton:
To improve the DX of plugin author's we should provide a couple of changes to the current data structure.

## Adoption of `RunnerConfig`

- [x] ~#227 => https://github.com/flowup/quality-metrics-cli/issues/227#issuecomment-1806756227~

This will make it easier to work with different cwd and config location settings.  
`npx ../../dist/packages/cli collect --config=../../packages/cli/test/plugin-progress-bar.config.mock.ts --outputDir=../../tmp` 

## Adoption of `RunnerConfig`

- [x] #228. 
   clodes by #229

If a transform function is provided,
the output of a plugin is piped through the provided transform function after we read the `plugin-output.josn`. 

```json
{
  plugins: [
    {
      outputPath: "out.json",
      transform: (data: Record<string, any>): PluginOutput => {
          return {
            //...
          }
      }
    }
  ]
}
```

## Adoption of `runner` type

- [x] #230

Would get executed inside the cli and its return value is directly used int the report.

```json
{
  plugins: [
    {
      runner: async (): void => {
         return // audit outputs
      }
    }
  ]
}
```

Implementation should be in different PRs.

matejchalk:
@BioPhoton Can this be closed? All the referenced issues are done.

PR #229, opened 2023-11-10
feat: add transform to persist config

### LINKED ISSUE #228: Adoption of `RunnerConfig` with a optional `transform` function.
If a transform function is provided,
the output of a plugin is piped through the provided transform function after we read the `plugin-output.josn`. 

```json
{
  plugins: [
    {
      outputPath: "out.json",
      transform: (data: Record<string, any>): PluginOutput => {
          return {
            //...
          }
      }
    }
  ]
}
```

related to https://github.com/flowup/quality-metrics-cli/issues/92

BioPhoton:
This PR is part of #92 and includes:
- add `transform` to `RunnerConfig`
- use `transform` in `executePlugin`
- use  `transform` in tests to reduce noise
- considers a todo : `@TODO consider just resting/spreading the values` to reduce code

closes #228


packages/models/src/lib/plugin-config-runner.ts
```diff
@@ -1,13 +1,21 @@
 import { z } from 'zod';
 import { filePathSchema } from './implementation/schemas';
 
+export const outputTransformSchema = z
+  .function()
+  .args(z.array(z.record(z.string(), z.unknown())));
+// @TODO add returns definition as AuditOutputs - this fails ATM when compiling
```
BioPhoton, line 7: Here I could need some help

packages/models/src/lib/plugin-config-runner.ts
```diff
@@ -1,5 +1,13 @@
 import { z } from 'zod';
 import { filePathSchema } from './implementation/schemas';
+import { auditOutputsSchema } from './plugin-process-output';
+
+export const outputTransformSchema = z
+  .function()
+  .args(z.array(z.record(z.string(), z.unknown())))
```
matejchalk, line 7: This is too restrictive, I believe. It could be any JSON, not necessarily an array of objects. E.g. many tools don't output the array at top-level, but it's nested within an object. Or it could even be a plain string that the transform function is gonna parse. 

```suggestion
  .args(z.unknown())
```

packages/models/src/lib/plugin-config-runner.ts
```diff
@@ -1,5 +1,13 @@
...
+  .returns(auditOutputsSchema);
```
matejchalk, line 8: Maybe we can support async transformations? :thinking: 

```suggestion
  .returns(z.union([auditOutputsSchema, z.promise(auditOutputsSchema)]));
```

packages/models/src/lib/plugin-config-runner.ts
matejchalk: Done, pushed a fix. The problem was a circular dependency caused by `pluginOutputSchema`, but that was unused so I just removed it.

matejchalk (changes requested):

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -54,70 +52,59 @@ export async function executePlugin(
...
+
+  // validate auditOutputs
+  auditOutputsCorrelateWithPluginOutput(auditOutputs, pluginConfigAudits);
+
+  // enrich `AuditOutputs` to `AuditReport`
+  const audits: AuditReport[] = auditOutputs.map((auditOutput: AuditOutput) =>
+    auditReportSchema.parse({
+      ...auditOutput,
+      ...pluginConfigAudits.find(audit => audit.slug === auditOutput.slug),
+    }),
```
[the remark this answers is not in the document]
BioPhoton: In the future we will receive AuditOutputs form the inline plugin so I would keep that in.

If you like I can then remove the first validation for auditOutput. WDYS?

packages/models/src/lib/plugin-config-runner.ts
BioPhoton: ❤️

packages/models/src/lib/plugin-config-runner.ts
BioPhoton: how did u figure out the dept?

packages/core/src/lib/implementation/execute-plugin.ts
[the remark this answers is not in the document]
BioPhoton: will implement it in the future

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -54,70 +51,60 @@ export async function executePlugin(
...
-      ...(description && { description }),
-      ...(docsUrl && { docsUrl }),
-      ...(groups && { groups }),
-    } satisfies PluginReport;
-  } catch (error) {
-    const e = error as Error;
-    throw new PluginOutputError(slug, e);
-  }
+        ...pluginConfigAudits.find(audit => audit.slug === auditOutput.slug),
+      } as unknown as AuditReport),
```
matejchalk, line 97: The `as unknown as` isn't very safe to use. In this case there's a more precise way by just casting the result of `find` (since you know you checked it already with `auditOutputsCorrelateWithPluginOutput`). That way it will still report other causes of errors.

```suggestion
        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
        ...pluginConfigAudits.find(audit => audit.slug === auditOutput.slug)!,
      }),
```

matejchalk (approved):

packages/models/src/lib/plugin-config-runner.ts
matejchalk: I got there through the `TypeError: Cannot read properties of undefined (reading 'merge')` error. This often means a circular dependency error, it seems (Zod couldn't create the object schema).

packages/core/src/lib/implementation/execute-plugin.ts
BioPhoton: fixed it differently

matejchalk (approved):

PR #238, opened 2023-11-11
refactor(core): improve report loading

### LINKED ISSUE #222: Consider handling undefined exports and test it
💡 🧩 utils: TODO consider handling undefined exports
https://github.com/flowup/quality-metrics-cli/blob/317167e955a56bfe8579192d6463bfeab585037e/packages/utils/src/lib/load-file.ts#L13

BioPhoton:
closes #222

### DIFF
--- a/packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
+++ b/packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
@@ -30,7 +30,7 @@ describe('readCodePushupConfig', () => {
     );
   });
 
-  it('should throw if file does not exisit', async () => {
+  it('should throw if file does not exist', async () => {
     const filepath = join('invalid-path', 'valid-export.mjs');
     await expect(readCodePushupConfig(filepath)).rejects.toThrow(
       `invalid-path`,
--- a/packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
+++ b/packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
@@ -32,7 +32,9 @@ describe('readCodePushupConfig', () => {
 
   it('should throw if file does not exisit', async () => {
     const filepath = join('invalid-path', 'valid-export.mjs');
-    await expect(readCodePushupConfig(filepath)).rejects.toThrow(`invalid-path${sep}valid-export.mjs`);
+    await expect(readCodePushupConfig(filepath)).rejects.toThrow(
+      `invalid-path${sep}valid-export.mjs`,
+    );
   });
 
   it('should throw if config is invalid', async () => {
--- a/packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
+++ b/packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
@@ -1,4 +1,4 @@
-import { join, sep } from 'path';
+import { join } from 'path';
 import * as process from 'process';
 import { describe, expect } from 'vitest';
 import { persistConfig, uploadConfig } from '@code-pushup/models/testing';
@@ -33,7 +33,7 @@ describe('readCodePushupConfig', () => {
   it('should throw if file does not exisit', async () => {
     const filepath = join('invalid-path', 'valid-export.mjs');
     await expect(readCodePushupConfig(filepath)).rejects.toThrow(
-      `invalid-path${sep}valid-export.mjs`,
+      `invalid-path`,
     );
   });
 
--- a/packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
+++ b/packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
@@ -1,0 +1,52 @@
+import { join } from 'path';
+import * as process from 'process';
+import { describe, expect } from 'vitest';
+import { persistConfig, uploadConfig } from '@code-pushup/models/testing';
+import { readCodePushupConfig } from './read-code-pushup-config';
+
+describe('readCodePushupConfig', () => {
+  it('should load file', async () => {
+    const module = await readCodePushupConfig(
+      join(process.cwd(), 'packages', 'core', 'test', 'minimal.config.mjs'),
+    );
+    expect(module).toEqual(
+      expect.objectContaining({
+        persist: expect.objectContaining(persistConfig()),
+        upload: expect.objectContaining({
+          ...uploadConfig(),
+          organization: 'code-pushup',
+        }),
+        categories: expect.arrayContaining([
+          expect.objectContaining({
+            slug: 'category-slug-1',
+          }),
+        ]),
+        plugins: expect.arrayContaining([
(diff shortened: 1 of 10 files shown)


BioPhoton:
@Tlacenka could you have a look why this fails on windows pls

BioPhoton:
@Tlacenka could you have a look please :)

packages/utils/test/fixtures/valid-export.mjs
Tlacenka, line 1: I suggest using a mocked file system with different types of valid and invalid files. While the bundleRequire logic needs to be tested, I think that passing it a path to a mocked file does not prevent us from testing the parsing of the file content. What do you think?

BioPhoton: BundleRequire can't be mocked unfo.

We should document that in the file imho.

packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+        plugins: expect.arrayContaining([
+          expect.objectContaining({
+            slug: 'mock-plugin-slug',
+          }),
+        ]),
+      }),
+    );
+  });
+
+  it('should throw if file does not exisit', async () => {
```
matejchalk, line 33:
```suggestion
  it('should throw if file does not exist', async () => {
```

packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+            slug: 'category-slug-1',
+          }),
+        ]),
+        plugins: expect.arrayContaining([
+          expect.objectContaining({
+            slug: 'mock-plugin-slug',
+          }),
+        ]),
+      }),
+    );
```
matejchalk, line 30: The `expect.objectContaining` is useful when you only want check some object properties, but not all. Other than `categories` and `plugins`, it appears to me like you're matching everything anyway. Doesn't it also pass without the `expect.objectContaining`? 

```suggestion
    expect(module).toEqual({
      persist: persistConfig(),
      upload: {
        ...uploadConfig(),
        organization: 'code-pushup',
      },
      categories: expect.arrayContaining([
        expect.objectContaining({
          slug: 'category-slug-1',
        }),
      ]),
      plugins: expect.arrayContaining([
        expect.objectContaining({
          slug: 'mock-plugin-slug',
        }),
      ]),
    });
```

packages/models/test/fixtures/categories.mock.ts
```diff
@@ -60,12 +60,21 @@ const CATEGORIES_MAP = {
   },
 } satisfies Record<string, CategoryConfig>;
 
-type CategorySlug = keyof typeof CATEGORIES_MAP;
-
-export function categoryConfig(
-  slug: CategorySlug = 'performance',
-): CategoryConfig {
-  return CATEGORIES_MAP[slug];
+export function categoryConfig(slug = 'category-slug-1'): CategoryConfig {
```
matejchalk, line 63: The old version had a realistic category prepared. Making it into a dummy config isn't a good change, I think.

packages/utils/src/lib/load-file.ts
```diff
@@ -1,5 +1,17 @@
...
+  constructor(filepath: string) {
+    super(`No export found in ${filepath}`);
+  }
+}
+
+export class NoFileError extends Error {
+  constructor(filepath: string) {
+    super(`The entry point "${filepath}" cannot be marked as external`);
+  }
+}
```
matejchalk, line 13: Is this used anywhere?

packages/utils/src/lib/load-file.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  join(process.cwd(), 'packages', 'utils', 'test', 'fixtures', fileName);
+describe('importEsmModule', () => {
+  it('should load file', async () => {
+    const module = await importEsmModule<{ name: string }>({
+      filepath: getFilepath('valid-export.mjs'),
+    });
+    expect(module).toBe('valid-export');
+  });
+
+  it('should throw if file does not existing', async () => {
```
matejchalk, line 15:
```suggestion
  it('should throw if file does not exist', async () => {
```

packages/utils/src/lib/load-file.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  it('should throw if export is not defined', async () => {
+    const filepath = getFilepath('no-export.mjs');
+    await expect(
+      importEsmModule<{ name: string }>({
+        filepath,
+      }),
+    ).rejects.toThrow(new NoExportError(filepath));
+  });
+
+  it('should throw if export is undefined defined', async () => {
```
matejchalk, line 32:
```suggestion
  it('should throw if export is undefined', async () => {
```

matejchalk (changes requested):

packages/utils/src/lib/load-file.ts
BioPhoton: It was used in a test before @Tlacenka pointed out the test will pass without the error 😅

packages/models/test/fixtures/categories.mock.ts
BioPhoton: I would move this data into lighthouse package and provide a more realistic mock?

packages/core/src/lib/implementation/read-code-pushup-config.spec.ts
BioPhoton: no

packages/utils/src/lib/load-file.ts
BioPhoton: removed

packages/utils/src/lib/load-file.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { join } from 'path';
+import * as process from 'process';
+import { NoExportError, importEsmModule } from '../lib/load-file';
+
+const getFilepath = (fileName: string) =>
+  join(process.cwd(), 'packages', 'utils', 'test', 'fixtures', fileName);
+describe('importEsmModule', () => {
```
matejchalk, line 7:
```suggestion
const getFilepath = (fileName: string) =>
  join(process.cwd(), 'packages', 'utils', 'test', 'fixtures', fileName);

describe('importEsmModule', () => {
```

packages/models/test/fixtures/categories.mock.ts
matejchalk: I would keep it here, otherwise we're mixing `scope:plugin` with `scope:core`. And the categories are cross-plugin anyway. 

Why not just keep this function as it was?

BioPhoton: done

matejchalk (approved):

PR #246, opened 2023-11-13
feat(cli): introduce the `onlyPlugins` option

### LINKED ISSUE #119: add argument `onlyPlugins` for commands `autorun`, `print-config` and `collect`
add argument `onlyPlugins` for commands `autorun`, `print-config` and `collect`:  
- `--onlyPlugin=eslint`
- `--onlyPlugin=eslint,lighthouse`

IKatsuba:
Close #119

I modified `configMiddleware` and filtered all plugins. This allowed us not to modify the commands.

### DIFF
--- a/e2e/cli-e2e/tests/print-config.spec.ts
+++ b/e2e/cli-e2e/tests/print-config.spec.ts
@@ -37,6 +37,7 @@ describe('print-config', () => {
       ]),
       // @TODO add test data to config file
       categories: expect.any(Array),
+      onlyPlugins: [],
     });
   });
 
@@ -61,6 +62,7 @@ describe('print-config', () => {
       }),
       plugins: expect.any(Array),
       categories: expect.any(Array),
+      onlyPlugins: [],
     });
   });
 
--- a/packages/cli/src/lib/implementation/config-middleware.spec.ts
+++ b/packages/cli/src/lib/implementation/config-middleware.spec.ts
@@ -1,7 +1,13 @@
 import { dirname, join } from 'path';
 import { fileURLToPath } from 'url';
-import { expect } from 'vitest';
-import { configMiddleware } from './config-middleware';
+import { SpyInstance, afterEach, beforeEach, describe, expect } from 'vitest';
+import { CoreConfig } from '@code-pushup/models';
+import {
+  configMiddleware,
+  filterCategoryByOnlyPluginsOption,
+  filterPluginsByOnlyPluginsOption,
+  validateOnlyPluginsOption,
+} from './config-middleware';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const withDirName = (path: string) => join(__dirname, path);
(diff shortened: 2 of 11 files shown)


packages/models/src/lib/global-options.ts
```diff
@@ -15,6 +15,11 @@ export const globalOptionsSchema = z.object({
   config: filePathSchema(
     "Path to config file in format `ts` or `mjs`. defaults to 'code-pushup.config.js'",
   ).default('code-pushup.config.js'),
+  onlyPlugins: z
```
BioPhoton, line 18: can we add the property to the CollectOptions and AutorunOptions type instead of global options please

BioPhoton (changes requested):
left a small comment. As the options does not apply for all commands e.g. `upload` we should put the type and options in the specific command.

packages/models/src/lib/only-plugins-options.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from 'zod';
+
+export const onlyPluginsOptionsSchema = z.object({
```
BioPhoton, line 3: I wonder if this should live in `core`... 🤔 or in `cli` it is definitely more a cli than a models thing.
I don't really know where to put it. Spent already some time moving the CLI command options around... Maybe @matejchalk has good input?

BioPhoton (commented):
I have one question on where to put the CLI option schema.. but no good answer. :D
Looks good otherwise!! 
🎉🎉🎉

packages/cli/src/lib/implementation/config-middleware.ts
```diff
@@ -1,29 +1,36 @@
...
+      },
+      persist: {
+        ...importedRc.persist,
+        ...cliOptions?.persist,
+      },
+      plugins: cliOptions.onlyPlugins?.length
+        ? importedRc.plugins.filter(plugin =>
+            cliOptions.onlyPlugins.includes(plugin.slug),
+          )
+        : importedRc.plugins,
```
matejchalk, line 30: How should we handle when a user passes an invalid slug to `--onlyPlugin`? This code will just ignore it, but maybe better to at least print a warning for the user, so they realize that they maybe just made a typo :thinking: 

Something like:
```
The --onlyPlugin argument references a plugin with "ligthouse" slug, but no such plugin is present in the configuration. Expected one of the following plugin slugs: "eslint", "lighthouse".
```

packages/models/src/lib/only-plugins-options.ts
matejchalk: I don't believe this schema is being used anywhere, since it's not part of the core config. In which case, we only need the type definition, since only Yargs validates it. The type definition should belong in `core`, because that's where it's used. (If we wanted to allow `onlyPlugins` in the config file, then that would make it part of `models`.)

matejchalk (changes requested):

packages/cli/src/lib/implementation/config-middleware.ts
```diff
@@ -1,29 +1,36 @@
...
+      categories: importedRc.categories,
```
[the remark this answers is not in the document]
BioPhoton: I think the option should work without any further code changes. So removing the refs in categories will make sense.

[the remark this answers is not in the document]
matejchalk: > I think the option should work without any further code changes.

Agreed.

> So removing the refs in categories will make sense.

Removing only the refs? I would remove the whole category. Otherwise the category score meaning changes. And categories without any refs aren't valid.

e2e/cli-e2e/tests/print-config.spec.ts
```diff
@@ -37,6 +37,8 @@ describe('print-config', () => {
       ]),
       // @TODO add test data to config file
       categories: expect.any(Array),
+      onlyPlugins: [],
+      'only-plugins': [],
```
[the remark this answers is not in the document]
BioPhoton: IMHO this is a result of yargs and slipped in..

BioPhoton (changes requested):
Ideally this package should only impact the `cli` package and is not reflected in `core` or `models`.

I opened #258 to make the scope here smaller. Added you as reviewer.

BioPhoton:
@IKatsuba 

PR closed please merge

packages/cli/src/lib/implementation/config-middleware.ts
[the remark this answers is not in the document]
IKatsuba: @matejchalk, are you suggesting removing the whole category if it has at least one removed ref?

IKatsuba:
> Ideally this package should only impact the `cli` package and is not reflected in `core` or `models`.

@BioPhoton I did it!

BioPhoton (approved):
Sweet!

packages/cli/src/lib/implementation/only-config-option.ts
```diff
@@ -0,0 +1,8 @@
+import { Options } from 'yargs';
+
+export const onlyPluginsOption: Options = {
```
BioPhoton, line 3: i guess this file needs to get removed?

BioPhoton (changes requested):

packages/cli/src/lib/implementation/config-middleware.ts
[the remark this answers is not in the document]
matejchalk: I would say yes, because skipping category entirely is the simplest option. We could log a warning, e.g. `Category "Performance" is ignored because it references audits from skipped plugin "Lighthouse"`.

If we wanted to be more clever about it, then we could tolerate if all the skipped plugin references have weight 0, because then it doesn't impact the score.

The alternative is to only filter out the references, but that can impact the category score drastically and is probably misleading. And there is a high likelihood that a category only references a single plugin, in which case no meaningful score is possible when that plugin is skipped.

Since I understand the motivation behind `--onlyPlugin` is for debugging a plugin and not really generating a complete report, then I think keeping it a simple and skipping categories which we're not sure what to do with is reasonable.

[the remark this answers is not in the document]
IKatsuba: @BioPhoton @matejchalk, let's draw the finish line — we skip the whole category if it has at least one skipped plugin ref. And it will work like this until another need arises. Am I right?

packages/cli/src/lib/implementation/only-config-option.ts
IKatsuba: No. It is an option config for commands

packages/cli/src/lib/implementation/config-middleware.ts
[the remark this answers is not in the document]
matejchalk: Agreed :+1:

packages/cli/src/lib/implementation/filter-kebab-case-keys.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+      camelCase: 'value',
+      snake_case: 'value',
+    };
+    const filtered = filterKebabCaseKeys(obj);
+    expect(filtered).toEqual({
+      camelCase: 'value',
+      snake_case: 'value',
+    });
+  });
+});
```
matejchalk, line 17: Nice :+1:

packages/cli/src/lib/print-config/command-object.ts
```diff
@@ -1,13 +1,21 @@
...
+    builder: {
+      onlyPlugins: onlyPluginsOption,
+    },
+    handler: yargsArgs => {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      const { _, $0, ...cleanArgs } = args;
+      const { _, $0, ...args } = yargsArgs;
+      // it is important to filter out kebab case keys
+      // because yargs duplicates options in camel case and kebab case
+      const cleanArgs = filterKebabCaseKeys(args);
```
matejchalk, line 18: Perfect :ok_hand:

BioPhoton (dismissed):
Thx for the changes

packages/cli/src/lib/implementation/config-middleware.spec.ts
```diff
@@ -52,3 +58,110 @@ describe('applyConfigMiddleware', () => {
...
+// without the `no-secrets` rule, this would be flagged as a security issue
+// eslint-disable-next-line no-secrets/no-secrets
+describe('filterCategoryByOnlyPluginsOption', () => {
+  let logSpy: SpyInstance;
+  beforeEach(() => {
+    logSpy = vi.spyOn(console, 'log');
+  });
+
+  afterEach(() => {
+    logSpy.mockClear();
```
matejchalk, line 101:
```suggestion
    logSpy.mockRestore();
```

Otherwise it won't revert to the original implementation.

packages/cli/src/lib/implementation/config-middleware.spec.ts
```diff
@@ -52,3 +58,110 @@ describe('applyConfigMiddleware', () => {
...
+});
+
+describe('validateOnlyPluginsOption', () => {
+  let logSpy: SpyInstance;
+  beforeEach(() => {
+    logSpy = vi.spyOn(console, 'log');
+  });
+
+  afterEach(() => {
+    logSpy.mockClear();
```
matejchalk, line 149:
```suggestion
    logSpy.mockRestore();
```

BioPhoton (dismissed):

matejchalk (approved):

Issue #215, opened 2023-11-10
core package test refactor


Tlacenka:
As a contributor to Code PushUp, I would like for the tests to be consistently structured and follow good practices.
There are several areas in which I identified room for improvement. In this issue, the following targeted changes should be made.

## Coverage
- [x] Add a happy path test to `read-code-pushup-config` and invalid config error scenario.
- [x] Test both the regular approach and ESM plugin approach once #248 is merged.

In both `collect` and `collect-and-persist`, the main cases should be tested. This means:
- [x] Add a test case for no plugins passed to `collect`.
- [x] Add a test case for explicitly adding multiple output options for `collect-and-persist` and explicitly mark each use case.

## Files
These tests are mostly in scope of **integration tests** (except for `json-to-gql`). Hence ,they should be categorised as integration tests and structured accordingly.
**Note**: `nx` configuration of integration vs unit tests is OOS. This is only about the location and naming of the files.
- [x] Each test file name reflects that it is an integration test (to be finalised).
- [x] Move all configurations and mocks for FS and other under `testing-utils`.
- [x] Remove the unused `types.ts` file.
- [x] It should be decided whether `read-code-pushup-config` and `json-to-gql` and the progress configs belong to this package - in my opinion, the first two functions belong in `utils` and the configurations should be moved to `testing-utils`. However, perhaps there is a reason to keep them in `core`?
- [x] Keep only one `README` file. Currently, there are two - one in the root folder and one under `docs`.
- [x] Merge `DEFAULT_TESTING_CLI_OPTIONS` from `constant.ts`, also used in other packages, under `testing-utils` or use hardcoded values for better readability where applicable.

## Good practices
In order to create and maintain tests easily, their readability should be prioritised. This means minimising boiler plate code, keeping the test flow linear and naming things based on their purpose.
- [x] Instead of mocking console, use `spyOn` for `console.log` in combination with `shouldHaveBeenCalledWith()`.
- [x] Move `portal-client`, `fs` and `fs/promises` mock setup to `testing-utils` and share setup for all FS vs upload flows there.
- [x] Remove boiler plate regarding `pluginConfig` in `execute-plugin`. Having one valid configuration and then overriding its options without the need for helper functions would increase the readability of the tests.
- [x] Remove branching logic in `collect-and-persist` and explicitly state which formats are tested in which file.
- [x] Improve test descriptions and assertions purpose in `persist` - it is not clear when JSON vs `stdout` vs `MD` is created.
[...]

PR #248, opened 2023-11-14
feat(core): add esm plugin logic

### LINKED ISSUE #177: Cleanup runnerconfig
Clean up runner config mocks. There are 2 files doing the same. 
https://github.com/flowup/quality-metrics-cli/pull/139#discussion_r1375396226

Follow up of #139 

closed by https://github.com/code-pushup/cli/pull/248

### LINKED ISSUE #230: Introduce new type for `runner`
Would get executed inside the cli and it's return value is directly used int the report.

```json
{
  plugins: [
    {
      runner: async (): void => {
         return // audit outputs
      }
    }
  ]
}
```

related to #92

BioPhoton:
This PR is related to #92 

> ⚠️ NOTICE:  
> This PR needs to get considered in #215 as it will simplify the way we test plugins ATM
> cc @Tlacenka 

It includes:
- separate logic for default runner (`RunnerConfig` type)
  - add more tests to cover the adopted logic
- add logic and tests for  (`RunnerFunction` type)
- Refactor `Observer` to `OnProgress`
- refactor `executePlugin` logic
  - avoid duplicated parsing
  - handle runner types
  - improve types
- resolves #177

closes #230 #177


packages/core/src/lib/implementation/execute-plugin.spec.ts
```diff
@@ -42,16 +43,47 @@ describe('executePlugin', () => {
...
+    const pluginCfg: PluginConfig = {
+      ...validPluginCfg,
+      runner: esmRunnerConfig,
+    };
+    const pluginResult = await executePlugin(pluginCfg);
+    expect(pluginResult.audits[0]?.slug).toBe('mock-audit-slug');
+    expect(() => auditOutputsSchema.parse(pluginResult.audits)).not.toThrow();
+  });
+
+  it('should throw if invalid runnerOutput is produced with transform', async () => {
```
matejchalk, line 78: Misleading name, the test doesn't use `outputTransform`.

packages/models/src/lib/global-options.ts
```diff
@@ -12,6 +12,7 @@ export const globalOptionsSchema = z.object({
       description: 'Outputs additional information for a run',
     })
     .default(false),
+  // @TODO move to cli package as it is only used there
```
matejchalk, line 15: Good point, probably doesn't need to be a schema then.

packages/models/src/lib/plugin-config-runner.ts
```diff
@@ -24,3 +24,15 @@ export const runnerConfigSchema = z.object(
...
+
+export const esmObserver = z.object({
+  next: z.function().args(z.unknown()).returns(z.void()).optional(),
+});
+export type EsmObserver = z.infer<typeof esmObserver>;
+
+export const esmRunnerConfigSchema = z
+  .function()
+  .args(esmObserver.optional())
+  .returns(z.promise(auditOutputsSchema));
```
matejchalk, line 36: I would suggest sync implementation should also be possible.

```suggestion
  .returns(z.union([auditOutputsSchema, z.promise(auditOutputsSchema)]));
```

packages/models/src/lib/plugin-config-runner.ts
```diff
@@ -24,3 +24,15 @@ export const runnerConfigSchema = z.object(
...
+
+export type EsmRunnerConfig = z.infer<typeof esmRunnerConfigSchema>;
```
matejchalk, line 38: Why not just call it `RunnerFunction`? The "Config"-suffix seems strange when it's not an object, and I'm not sure what to make of the "Esm"-prefix.

packages/models/src/lib/plugin-config-runner.ts
```diff
@@ -24,3 +24,15 @@ export const runnerConfigSchema = z.object(
...
 export type RunnerConfig = z.infer<typeof runnerConfigSchema>;
+
+export const esmObserver = z.object({
+  next: z.function().args(z.unknown()).returns(z.void()).optional(),
+});
+export type EsmObserver = z.infer<typeof esmObserver>;
+
+export const esmRunnerConfigSchema = z
+  .function()
+  .args(esmObserver.optional())
```
matejchalk, line 35: What is the purpose of the Observer argument? As a Code PushUp maintainer, I don't understand what I should use it for. Do I have to use it? If so, why is it optional? I would guess to a user it would make even less sense.

packages/models/test/fixtures/runner-config.mock.ts
```diff
@@ -37,3 +41,15 @@ export function echoRunnerConfig(
     outputFile,
   };
 }
+
+/**
+ * Use this helper as a general purpose mock with working defaults
+ * @param options
+ */
+export function esmRunnerConfig(
+  options?: Partial<EsmRunnerConfig>,
```
matejchalk, line 50: The options aren't used and I don't understand the intention (`Partial` of a function type?). 

The implementation creates invalid data. Do we want that for our tests?

I would suggest that by default the runner function returns an empty array, but that I can pass in my own function as an argument.

packages/core/src/lib/implementation/runner.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export async function executeEsmRunner(
+  runner: EsmRunnerConfig,
+  observer?: Observer,
+): Promise<RunnerResult> {
+  const date = new Date().toISOString();
+  const start = performance.now();
+
+  // execute plugin runner
+  const audits = await runner(observer);
```
matejchalk, line 60: I would parse the output with Zod. If the user uses a TypeScript config file, then there is some type-safety (although no guarantee), but for JavaScript configs there is none.

matejchalk: Although maybe Zod automatically handles it? Not sure, I would like to see a test case for that ideally.

packages/core/src/lib/implementation/runner.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  it('should throw if plugin throws', async () => {
+    const nextSpy = vi.fn();
+    await expect(
+      executeEsmRunner(
+        () => Promise.reject(new Error('plugin exex mock error')),
+        { next: nextSpy },
+      ),
+    ).rejects.toThrow('plugin exex mock error');
+  });
```
matejchalk, line 78:
```suggestion
  it('should throw if plugin throws', async () => {
    const nextSpy = vi.fn();
    await expect(
      executeEsmRunner(
        () => Promise.reject(new Error('plugin exec mock error')),
        { next: nextSpy },
      ),
    ).rejects.toThrow('plugin exec mock error');
  });
```

matejchalk (changes requested):

packages/models/src/lib/global-options.ts
BioPhoton: In another PR

packages/models/src/lib/plugin-config-runner.ts
BioPhoton: done

packages/models/src/lib/plugin-config-runner.ts
BioPhoton: refactored the observer part

packages/core/src/lib/implementation/runner.ts
BioPhoton: done

BioPhoton: test added

BioPhoton:
Like it!! This is a nice feature and a really good cleanup

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -30,7 +26,7 @@ export class PluginOutputMissingAuditError extends Error {
  *
  * @public
  * @param pluginConfig - {@link ProcessConfig} object with runner and meta
- * @param observer - process {@link ProcessObserver}
+ * @param observer - process {@link Observer}
```
matejchalk, line 29:
```suggestion
 * @param onProgress - progress handler {@link OnProgress}
```

packages/models/src/lib/plugin-config-runner.ts
```diff
@@ -24,3 +24,16 @@ export const runnerConfigSchema = z.object(
 );
 
 export type RunnerConfig = z.infer<typeof runnerConfigSchema>;
+
+export const onProgressSchema = z
+  .function()
+  .args(z.unknown())
```
matejchalk, line 30: Should the argument really be of type `unknown`? As user, this means I have no clue what I should pass in. I notice one test uses `'update'`, so is it a `string`? And why would I want to pass anything in? Genuine questions, there's no description.

packages/utils/src/lib/execute-process.ts
```diff
@@ -85,20 +85,20 @@ export type ProcessConfig = {
...
- *  next: (stdout) => console.log(stdout)
+ *  onStdout: (stdout) => console.log(stdout)
  *  }
  */
 export type ProcessObserver = {
-  next?: (stdout: string) => void;
+  onStdout?: (stdout: string) => void;
   error?: (error: ProcessError) => void;
   complete?: () => void;
 };
```
matejchalk, line 104: I think all these event handlers should be prefixed by `on`, or none of them should.

packages/models/src/lib/plugin-config-runner.ts
BioPhoton: For `RunnerConfig` it could be `string` `StringBuffer` and it is the output of the onStdin event in most of the cases. 
If `RunnerFuntion` is used it can be anything basically as it is up to the plugin author

BioPhoton (commented):
I implemented all changes

matejchalk (approved):

PR #253, opened 2023-11-14
chore(utils): move benchmarks to typescript, adjust types

### LINKED ISSUE #154: Turn the benchmarks in `perf` folders to TS and use `ts-node` script to execute them

MishaSeredenkoPushBased:
All benchmarks are now .ts, added types where necessary. 
Fixed some places in the benchmarked functions, where they returned not proper values - it was the reason of the high scores of optimized1 and optimized2 benchmarks 🙂 
Used tsx for runtime, modified perf command, changed tsconfig.lib at utils.
Closes #154


packages/utils/tsconfig.lib.json
```diff
@@ -5,13 +5,12 @@
     "declaration": true,
     "types": ["node"]
   },
-  "include": ["src/**/*.ts"],
+  "include": ["src/**/*.ts", "perf/**/*.ts"],
```
BioPhoton, line 8: shouldn't this be the other way round? Not included in the lib bundle and lib included in perf bundle?

BioPhoton (commented):
Nice that we caught the broken implementations with TS :D

packages/utils/tsconfig.lib.json
MishaSeredenkoPushBased: Added `tsconfig.perf.json`

matejchalk (approved):

PR #258, opened 2023-11-15
refactor: global options location

BioPhoton:
This PR could go before #246 as there are a couple of questions it would resolve.

**It includes:**
- relocating of global options
- separation responsibilities between `cli` and `core`
- type improvements for commands
- deleting empty file 👻
- small type tweaks 

**Out of scope 😇:**   
- 🧹 utils - merge file utils into 1 file


.gitignore
```diff
@@ -1,7 +1,7 @@
 # See http://help.github.com/ignore-files/ for more about ignoring files.
 
 .env
-
+.nx
```
matejchalk, line 4: What's in `.nx`? Ideally we should have a comment explaing why it's ignored.

packages/cli/src/lib/autorun/command-object.ts
```diff
@@ -1,10 +1,14 @@
...
+  CollectOptions,
+  UploadOptions,
+  collectAndPersistReports,
+  upload,
+} from '@code-pushup/core';
 import { CLI_NAME } from '../cli';
-import { GeneralCliOptions } from '../implementation/model';
 
+type AutorunOptions = CollectOptions & UploadOptions;
 export function yargsAutorunCommandObject() {
```
matejchalk, line 12:
```suggestion
type AutorunOptions = CollectOptions & UploadOptions;

export function yargsAutorunCommandObject() {
```

packages/core/package.json
```diff
@@ -5,7 +5,8 @@
     "@code-pushup/models": "*",
     "@code-pushup/utils": "*",
     "@code-pushup/portal-client": "^0.1.2",
-    "chalk": "^5.3.0"
+    "chalk": "^5.3.0",
+    "zod": "^3.22.4"
```
matejchalk, line 9: Why? Only `models` should use `zod` directly, I believe. 

You don't need `globalOptionsSchema`, just the `GlobalOptions` type. Yargs doesn't use Zod schemas.

packages/core/src/lib/implementation/read-code-pushup-config.ts
```diff
@@ -9,15 +13,16 @@ export class ConfigPathError extends Error {
 }
 
 export async function readCodePushupConfig(filepath: string) {
-  const isFile = (await stat(filepath)).isFile();
+  const validFilepath = filePathSchema('config file path').parse(filepath);
```
matejchalk, line 16: I would revert these changes. We don't need to include Zod just for this, the "`filePathSchema`" only checks string length. Just check if the file exists with `stat`, we don't need to complicate things.

matejchalk (changes requested):

.gitignore
BioPhoton: done

packages/core/src/lib/implementation/read-code-pushup-config.ts
BioPhoton: done

packages/core/package.json
BioPhoton: ATM zod is used to provide default values. Ill refactor to yargs and types only

packages/core/src/lib/implementation/read-code-pushup-config.ts
```diff
@@ -9,15 +9,18 @@ export class ConfigPathError extends Error {
...
 
+  const isFile = (await stat(filepath)).isFile();
   if (!isFile) {
     throw new ConfigPathError(filepath);
   }
 
   return importEsmModule<CoreConfig>(
     {
-      filepath,
+      filepath: filepath,
```
BioPhoton, line 23:
```suggestion
      filepath,
```

matejchalk (approved):
LGTM :slightly_smiling_face:

PR #273, opened 2023-11-15
feat(cli): add history command

### LINKED ISSUE #216: History command
> As a user,
> I would like to run a command that collets past commits automatically as then I would see history data in the portal with reports showing the trends and progress

- [ ] collect report & if configured portal also upload 
- [ ] support for selecting branch 
- [x] collect by number of commits from the latest in target branch
- [ ] collect by date in range today till provided time in past

BioPhoton:
# Add history command to the CLI.

> [!note]
> *Skipped e2e test*
>
> One remaining flaw in this PR is the e2e test of history that requires a clean status. 
> By default the command fails if you don't set `forceCleanStatus` to true.
> In both cases I is disturbing while developing on a PR. I'm very open for feedback.
> 
> Screenshot 2024-03-13 at 18 22 24
> We will skip this tests

This command helps to crawl a repositories commit history and produce a report for each run.

To run the command execute `code-pushup history`. 
This will check out the target branch, by default 'main' and crawl the history of commits. By default the last 5 commits are crawled.

See the options table for all possible arguments of the command: https://github.com/code-pushup/cli/blob/e4b3f19b4fc0dd77c929bb7b0675b37b6d46ee4a/packages/cli/README.md#history-command

**This PR includes:**
- [x] adding helpers for git #469 
  - [x] unit-tests
  - [x] integration-tests
- [x] adding the history logic to `core`
  - [x] unit-tests
  - [x] integration-tests 
- [x] adding the command logic to `cli` 
  - [x] unit-tests
  - [x] integration-tests
  - [x] integrate e2e-tests
- [x] adding docs to [cli/Readme.md](https://github.com/code-pushup/cli/blob/e4b3f19b4fc0dd77c929bb7b0675b37b6d46ee4a/packages/cli/README.md#history-command)

**Example recording of the run:**

https://github.com/code-pushup/cli/assets/10064416/7ad22c2a-6604-4dda-8630-fdf098847f0d

closes #216 

**Follow up PRs:**

- adding prompt helper
  - unit-tests
  - integration-tests


packages/core/src/lib/implementation/persist.ts
```diff
@@ -26,12 +26,12 @@ export class PersistError extends Error {
 
 export async function persistReport(
   report: Report,
-  options: Required<PersistConfig>,
+  options: Required<PersistConfig>
```
BioPhoton, line 29:
```suggestion
  options: Required<PersistConfig>,
```

.prettierignore
```diff
@@ -1,4 +1,4 @@
 # Add files here to ignore them from prettier formatting
 /dist
 /coverage
-/.nx/cache
\ No newline at end of file
+/.nx/cache
```
BioPhoton, line 4:
```suggestion
```

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -507,7 +507,7 @@ describe('toAuditOutputs', () => {
               data: 'data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAYEBQYFBAYGBQYHBwY...',
             },
           ],
-        },
+        } satisfies Details.Filmstrip,
```
BioPhoton, line 510:
```suggestion
        },
```

BioPhoton, line 510:
```suggestion
        },
```

e2e/cli-e2e/tests/history.e2e.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      ...report,
+      plugins: report.plugins.map(omitVariableData) as PluginReport[],
+    });
+  /* eslint-enable @typescript-eslint/no-unused-vars */
+
+  beforeEach(async () => {
+    await cleanTestFolder('tmp/e2e');
+  });
+
+  it('should run ESLint plugin and create report.json for the last 5 commits', async () => {
```
Tlacenka, line 26: In our E2E tests, we should not run the commands on our own repository. Since the `history` command also forces a clean state, it would delete people's progress, that is a hard no.

A way of testing a `history` command would be to create a temporary repository, say with the `todos-app` data (so 2 commits, relatively small codebase) temporarily, so that our original repo is not affected and the number of commits is minimal - so 2 (5 is pushing it and could take a long time).

One could then run the report for that temporary repository and delete it at the end of the test.

One important question though is to ask oneself if the time spent on the setup is worth the payoff.

e2e/cli-e2e/tests/history.e2e.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    const outputDirFromRoot = join('tmp', 'e2e', 'react-todos-app', 'history');
+    const reportPaths = await readdir(outputDirFromRoot);
+    const results = await Promise.all(
+      reportPaths.map(path => readJsonFile(join(outputDirFromRoot, path))),
+    );
+
+    expect(results).toHaveLength(5);
+    results.forEach(report => {
+      expect(() => reportSchema.parse(report)).not.toThrow();
+    });
```
Tlacenka, line 51: There should be no loops in tests (same for `map` above). If only 2 commits are chosen, it is clear how many reports the commands run for. Now it wasn't even clear to me until I read the docs that 5 commits are the default number - 5 seemed like a random number.

packages/cli/README.md
```diff
@@ -240,6 +240,28 @@ Run plugins, collect results and upload the report to the Code PushUp portal.
 
 Refer to the [Common Command Options](#common-command-options) for the list of available options.
 
+#### `history` command
+
+Usage:
+`code-pushup history`
+
+Description:
+Run plugins, collect results and upload the report to the Code PushUp portal for a specified amount of the repository history.
```
Tlacenka, line 249:
```suggestion
Run plugins, collect results and upload the report to the Code PushUp portal for a specified number of commits.
```

packages/cli/README.md
```diff
@@ -240,6 +240,28 @@ Run plugins, collect results and upload the report to the Code PushUp portal.
...
+
+Refer to the [Common Command Options](#common-command-options) for the list of available options.
+
+| Option                   | Type      | Default  | Description                                                      |
+| ------------------------ | --------- | -------- | ---------------------------------------------------------------- |
+| **`--targetBranch`**     | `string`  | 'main'   | Branch to crawl history.                                         |
+| **`--forceCleanStatus`** | `boolean` | `report` | If we reset the status to a clean git history forcefully or not. |
+| **`--maxCount`**         | `number`  | 5        | Number of steps in history.                                      |
```
Tlacenka, line 257:
```suggestion
| **`--maxCount`**         | `number`  | 5        | Number of commits.                                      |
```

packages/cli/src/lib/history/history.options.ts
```diff
@@ -0,0 +1,42 @@
...
+    from: {
+      // https://git-scm.com/docs/git-log#_description
+      describe: 'hash to start in history',
+      type: 'string',
+    },
+    to: {
+      // https://git-scm.com/docs/git-log#_description
+      describe: 'hash to start in history',
+      type: 'string',
+    },
```
Tlacenka, line 40: Just the `git log` documentation is too long. But it has links to sections/options one can use to help user understand how it is done by `git` only.

```suggestion
    from: {
      // https://git-scm.com/docs/git-log#Documentation/git-log.txt-ltrevision-rangegt
      describe: 'hash to first commit in history',
      type: 'string',
    },
    to: {
      // https://git-scm.com/docs/git-log#Documentation/git-log.txt-ltrevision-rangegt
      describe: 'hash to last commit in history',
      type: 'string',
    },
```

packages/cli/src/lib/history/history.options.ts
```diff
@@ -0,0 +1,42 @@
...
+      type: 'number',
+      // eslint-disable-next-line no-magic-numbers
+      default: 5,
+    },
+    from: {
+      // https://git-scm.com/docs/git-log#_description
+      describe: 'hash to start in history',
+      type: 'string',
+    },
+    to: {
```
Tlacenka, line 36: As a user, I would assume the default `to` would be `HEAD` yet the `history` command does not put a default value here.

packages/cli/src/lib/history/history-command.ts
```diff
@@ -0,0 +1,68 @@
...
+        ...restOptions
+      } = args as unknown as HistoryCliOptions & HistoryOptions;
+
+      // determine history to walk
+      const git = simpleGit();
+      const log = await git.log({ maxCount, from, to });
+      const commitsToAudit = log.all
+        .map(({ hash }) => hash)
+        // crawl from oldest to newest
+        .reverse();
```
Tlacenka, line 49: This is done by the `getHashes` function which is now unused except for tests. Either use the function or delete it and leave it hardcoded.

packages/cli/src/lib/history/history-command.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      // run history logic
+      const reports: unknown[] = await history(
```
Tlacenka, line 52: Why cast it as `unknown[]`? The `history` function returns `string[]` promise.
```suggestion
      const reports = await history(
```

packages/cli/src/lib/history/history-command.ts
```diff
@@ -0,0 +1,68 @@
...
+        {
+          ...restOptions,
+          targetBranch,
+          forceCleanStatus,
+        },
+        commitsToAudit,
+      );
+
+      // go back to initial branch
```
Tlacenka, line 61: Will this happen even if the `history` command throws an error or will the user end up somewhere in `git` history?

packages/core/src/index.ts
Tlacenka, line 1: I noticed that the type `HistoryOptions` makes typescript complain in `core/src/lib/history.unit.test.ts`. Please fix the typing.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -31,7 +31,7 @@ describe('git utils in a git repo', () => {
   describe('without a branch and commits', () => {
     it('getCurrentBranchOrTag should throw if no branch or tag is given', async () => {
       await expect(getCurrentBranchOrTag(emptyGit)).rejects.toThrow(
-        'Could not get current tag or branch.',
+        'No names found, cannot describe anything',
```
Tlacenka, line 34: This message is confusing, as a user I wouldn't know what is going on.

packages/utils/src/lib/git.ts
```diff
@@ -37,35 +37,68 @@ export async function toGitPath(
   return formatGitPath(path, gitRoot);
 }
 
+export class GitStatueError extends Error {
```
Tlacenka, line 40: what is a `GitStatue`?

packages/cli/src/lib/history/history-command.unit.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      ['history', '--config=/test/code-pushup.config.ts'],
+      {
+        ...DEFAULT_CLI_CONFIGURATION,
+        commands: [yargsHistoryCommandObject()],
+      },
+    ).parseAsync();
+
+    expect(result).toEqual(
+      expect.objectContaining({
+        onlyPlugins: [],
```
Tlacenka, line 67: `onlyPlugins` are not fully related to this command. Testing their value should be in the `onlyPlugins` options tests.

packages/cli/README.md
```diff
@@ -240,6 +240,28 @@ Run plugins, collect results and upload the report to the Code PushUp portal.
...
+| **`--skipUploads`**      | `boolean` | `false`  | Upload created reports                                           |
+| **`--from`**             | `string`  | n/a      | Hash to start in history                                         |
+| **`--to`**               | `string`  | n/a      | Hash to end in history                                           |
```
Tlacenka, line 260: Wouldn't it be more user friendly to have a default value `HEAD` (current commit)?

packages/cli/src/lib/history/history-command.unit.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  it('should have correct default arguments if no cli options are provided', async () => {
+    const result = await yargsCli(
+      ['history', '--config=/test/code-pushup.config.ts'],
+      {
+        ...DEFAULT_CLI_CONFIGURATION,
+        commands: [yargsHistoryCommandObject()],
+      },
+    ).parseAsync();
+
+    expect(result).toEqual(
```
Tlacenka, line 65: Please create separate yargs integration tests. Unit tests of a given command only test its internal logic, not parsing of its options. That is done in `yargs-cli.integration.test.ts`. Same for the test below.

packages/cli/src/lib/history/history.model.ts
```diff
@@ -0,0 +1,7 @@
+import { LogOptions } from 'simple-git';
+import { HistoryOnlyOptions } from '@code-pushup/core';
+
+export type HistoryCliOptions = {
+  targetBranch?: string;
+} & Pick<LogOptions, 'maxCount' | 'from' | 'to'> &
+  Required<HistoryOnlyOptions>;
```
Tlacenka, line 7: Why is this `Required`? The `forceCleanStatus` does not have a default value.

packages/utils/src/lib/execute-process.ts
```diff
@@ -81,6 +81,7 @@ export type ProcessConfig = {
   command: string;
   args?: string[];
   cwd?: string;
+  shell?: boolean;
```
Tlacenka, line 84: I don't see this used anywhere. What is the purpose of this?

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -154,13 +154,39 @@ describe('git utils in a git repo', () => {
...
+              {
+                path: 'new-file.md',
+                index: '?',
+                working_dir: '?',
+              },
+            ],
+          },
+          null,
+          2,
+        )}`,
```
Tlacenka, line 170: Please use `toMatch` or similar matchers to assert parts of this message. Not every character needs to be asserted here. It bloats the test. Same for the test below.

project.json
```diff
@@ -75,6 +75,21 @@
           "target": "build"
         }
       ]
+    },
+    "run-history": {
```
Tlacenka, line 79: Please use **all** dependencies from the other commands, including the `plugin-js-packages`.

Tlacenka (changes requested):
Has potential.
At the same time, this PR still raises a lot of question marks for me.

UX feedback: When I need to run the `history` command, it requires me to clean my repo. But what if I need to change the `code-pushup.config.ts` file before running it? I now need to  push this temporary change and manually enter `--to` option as it is no longer `HEAD`.

e2e/cli-e2e/tests/history.e2e.test.ts
BioPhoton: I'm happy to skip e2e :)

e2e/cli-e2e/tests/history.e2e.test.ts
BioPhoton: added 5 random `.at(POSITION)` to avoid the loop. same for the `map`.

packages/cli/src/lib/history/history.model.ts
BioPhoton: removed

packages/utils/src/lib/execute-process.ts
BioPhoton: nice catch THX!! Shouldnt there be a linting rule for this?

packages/cli/src/lib/history/history-command.ts
```diff
@@ -0,0 +1,68 @@
...
+        Object.keys(yargsHistoryOptionsDefinition()),
+        'History Options:',
+      );
+      return yargs;
+    },
+    handler: async args => {
+      // eslint-disable-next-line no-console
+      console.log(chalk.bold(CLI_NAME));
+      // eslint-disable-next-line no-console
+      console.log(chalk.gray(`Run ${command}`));
```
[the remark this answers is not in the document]
BioPhoton: used ui logger. I switched to `log` as `info` is not available n certain scenarios, at least on my machine

project.json
BioPhoton: done

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
```diff
@@ -9,6 +9,7 @@ Commands:
   code-pushup autorun       Shortcut for running collect followed by upload
   code-pushup collect       Run Plugins and collect results
   code-pushup upload        Upload report results to the portal
+  code-pushup history       Create history of commits
```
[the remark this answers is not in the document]
BioPhoton: The arguments belong to history only therefor they are logged only for history. That is the behavor of the tool.

packages/cli/src/lib/history/history-command.unit.test.ts
BioPhoton: No other command is tested like this, so I will stick to the current style and we can do a refactoring in another PR.

packages/cli/src/lib/history/history.options.ts
BioPhoton: done

packages/cli/src/lib/history/history-command.ts
BioPhoton: use it now

e2e/cli-e2e/tests/history.e2e.test.ts
matejchalk: If we were to go down that route, then @Tlacenka's approach is a good one - something along the lines of `const git = simpleGit('examples/react-todos-app'); git.init(); ...` and always clean up afterwards. In the [E2E tests for the `compare` command](https://github.com/code-pushup/cli/blob/main/e2e/cli-e2e/tests/compare.e2e.test.ts), I actually make changes to the example app and clean them up afterwards, that's quite similar. The only extra prerequisite is that there are no unstaged changes to the example app.

On the other hand, I wouldn't yet consider the `history` command as a critical flow. I understand it's for one-off usages by us, not something every customer will run in their CI. So I'm leaning towards E2E tests not being worth the hassle in this case.

packages/cli/src/lib/history/history-command.ts
BioPhoton: removed. THX!

e2e/cli-e2e/tests/history.e2e.test.ts
BioPhoton: ok

packages/utils/src/lib/git.integration.test.ts
BioPhoton: When I think about it... I guess I will forward the git error directly.

packages/utils/src/lib/git.ts
BioPhoton: A special document ;) 

Will fix ir!

BioPhoton: fixed

packages/cli/README.md
BioPhoton: I commented on it here:
https://github.com/code-pushup/cli/blob/4f877217698a990e25e3cc83279d9e49f0fa548c/packages/core/src/lib/history.ts#L67-L105

And here:
https://github.com/code-pushup/cli/blob/6a7c7e681a5e97210997d7530997f49e34d767e8/packages/core/src/lib/history.ts#L113-L116

Also the tests would fail because of errors.

packages/utils/src/lib/git.integration.test.ts
BioPhoton: I checked and actually did it already. So what we could do is adding a statement where in the code it happened, but then the stack track gets obsulete.

I suggest we keep the git error message.

packages/core/src/index.ts
BioPhoton: done!

packages/cli/src/lib/history/history-command.ts
BioPhoton: try catched

e2e/cli-e2e/tests/history.e2e.test.ts
BioPhoton: added a issue for the additional tests #568

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
matejchalk, line 12:
```suggestion
  code-pushup history       Collect reports for commit history
```

e2e/cli-e2e/tests/history.e2e.test.ts
matejchalk: In order for this PR to be merged, please either _remove these E2E tests_, or rewrite them as discussed _in this PR_. I don't want to merge anything which messes around with our actual Git repo, particularly if it's using a force flag to do it.

e2e/cli-e2e/tests/history.e2e.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    expect(stderr).toBe('');
+
+    const outputDirFromRoot = join('tmp', 'e2e', 'react-todos-app', 'history');
+    const reportPaths = await readdir(outputDirFromRoot);
+    const results = await Promise.all([
+      readJsonFile(join(outputDirFromRoot, reportPaths.at(0) as string)),
+      readJsonFile(join(outputDirFromRoot, reportPaths.at(1) as string)),
+      readJsonFile(join(outputDirFromRoot, reportPaths.at(2) as string)),
+      readJsonFile(join(outputDirFromRoot, reportPaths.at(3) as string)),
+      readJsonFile(join(outputDirFromRoot, reportPaths.at(4) as string)),
```
matejchalk, line 49:
```suggestion
      readJsonFile(join(outputDirFromRoot, reportPaths.at(0)!)),
      readJsonFile(join(outputDirFromRoot, reportPaths.at(1)!)),
      readJsonFile(join(outputDirFromRoot, reportPaths.at(2)!)),
      readJsonFile(join(outputDirFromRoot, reportPaths.at(3)!)),
      readJsonFile(join(outputDirFromRoot, reportPaths.at(4)!)),
```

packages/cli/README.md
```diff
@@ -240,6 +240,28 @@ Run plugins, collect results and upload the report to the Code PushUp portal.
...
+
+Description:
+Run plugins, collect results and upload the report to the Code PushUp portal for a specified number of commits.
+
+Refer to the [Common Command Options](#common-command-options) for the list of available options.
+
+| Option                   | Type      | Default  | Description                                                      |
+| ------------------------ | --------- | -------- | ---------------------------------------------------------------- |
+| **`--targetBranch`**     | `string`  | 'main'   | Branch to crawl history.                                         |
+| **`--forceCleanStatus`** | `boolean` | `report` | If we reset the status to a clean git history forcefully or not. |
```
matejchalk, line 256: I'm guessing the default of a boolean flag is `true` or `false`, not `report` :sweat_smile:

packages/cli/README.md
```diff
@@ -240,6 +240,28 @@ Run plugins, collect results and upload the report to the Code PushUp portal.
...
+| **`--maxCount`**         | `number`  | 5        | Number of commits.                                               |
+| **`--skipUploads`**      | `boolean` | `false`  | Upload created reports                                           |
+| **`--from`**             | `string`  | n/a      | Hash to start in history                                         |
+| **`--to`**               | `string`  | n/a      | Hash to end in history                                           |
+
+> [!NOTE]  
+> If no target branch is set the current branch is used as target branch
```
matejchalk, line 263: This appears to be at odds with the options table above, which has `'main'` in its default column.

packages/cli/src/lib/history/history-command.ts
```diff
@@ -0,0 +1,63 @@
...
+        maxCount,
+        from,
+        to,
+        ...restOptions
+      } = args as unknown as HistoryCliOptions & HistoryOptions;
+
+      // determine history to walk
+      const git = simpleGit();
+
+      const commits: string[] = await getHashes({ maxCount, from, to }, git);
```
matejchalk, line 44:
```suggestion
      const commits: string[] = await getHashes({ maxCount, from, to });
```

packages/cli/src/lib/history/history-command.ts
```diff
@@ -0,0 +1,63 @@
...
+        const reports = await history(
+          {
+            ...restOptions,
+            targetBranch,
+            forceCleanStatus,
+          },
+          commits,
+        );
+
+        ui().logger.log(`Reports: ', ${reports.length}`);
```
matejchalk, line 56: I guess the intention isn't to log `Reports: ', 5`, but `Reports: 5`:

```suggestion
        ui().logger.log(`Reports: ${reports.length}`);
```

packages/utils/src/lib/git.ts
```diff
@@ -37,35 +37,68 @@ export async function toGitPath(
...
+
 export async function guardAgainstLocalChanges(
   git = simpleGit(),
 ): Promise<void> {
-  const isClean = await git.status(['-s']).then(r => r.files.length === 0);
-  if (!isClean) {
-    throw new Error(
-      'Working directory needs to be clean before we you can proceed. Commit your local changes or stash them.',
-    );
+  const status = await git.status(['-s']); //.then(r => r.files.length === 0);
```
matejchalk, line 83:
```suggestion
  const status = await git.status(['-s']);
```

project.json
```diff
@@ -75,6 +75,22 @@
...
+            "plugin-eslint",
+            "plugin-coverage",
+            "plugin-js-packages",
+            "examples-plugins",
+            "react-todos-app"
+          ],
+          "target": "build"
+        }
+      ]
     }
```
matejchalk, line 94: It's no longer need to configure individual Nx target to run CLI commands. Since I merged #574, there is now a single `code-pushup` target you can add any command to - e.g. `npx nx code-pushup -- history ...`.

testing/test-utils/src/lib/utils/minimal-config.mock.ts
```diff
@@ -50,7 +50,7 @@ export const MINIMAL_CONFIG_MOCK: CoreConfig = {
   plugins: [MINIMAL_PLUGIN_CONFIG_MOCK],
 };
 
-export const MINIMAL_HISTORY_CONFIG_MOCK: CoreConfig = {
```
matejchalk, line 53: Why remove the type annotation? :raised_eyebrow:

matejchalk (changes requested):

e2e/cli-e2e/tests/history.e2e.test.ts
BioPhoton: Removed

packages/cli/README.md
BioPhoton: set to false in options as default and in the docs.

packages/cli/README.md
BioPhoton: removed hit

project.json
BioPhoton: removed

testing/test-utils/src/lib/utils/minimal-config.mock.ts
BioPhoton: added back

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
matejchalk: The suggestion is marked as resolved, but the description wasn't changed. Could you update it please? "Create history of commits" sounds a like a description of `git log`, it's not clear that `history`'s running `code-pushup` on those commits. Obviously, it needs to be changed in the source code (`packages/cli/src/lib/history/history-command.ts`) as well as in this snapshot.

image

BioPhoton: Thanks! Got confused when merging main and lost this comment. I unfortunately went with the wrong choice then. 

Updated it in the test as well as source code. 
https://github.com/code-pushup/cli/pull/273/commits/ea9083a5a78d4ebcffe1e50c078a9a4f6bbf8b64

matejchalk (approved):

packages/cli/src/lib/history/history-command.unit.test.ts
Tlacenka: 🧪 Testing consistency: All commands are tested in the way I described. So please adjust the tests to be consistent with the adopted style, this one mixes two different things.
- Autorun has unit tests only with function calls [here](https://github.com/code-pushup/cli/blob/main/packages/cli/src/lib/autorun/autorun-command.unit.test.ts). Its options are tested **separately** in yargs [integration tests](https://github.com/code-pushup/cli/blob/main/packages/cli/src/lib/yargs-cli.integration.test.ts).
- Collect same [here](https://github.com/code-pushup/cli/blob/main/packages/cli/src/lib/collect/collect-command.unit.test.ts).
etc.

packages/core/src/lib/history.unit.test.ts
```diff
@@ -24,7 +24,10 @@ vi.mock('./upload', () => ({
 
 describe('history', () => {
   it('should check out all passed commits and reset to initial branch or tag', async () => {
-    await history(MINIMAL_HISTORY_CONFIG_MOCK, ['abc', 'def']);
+    await history(MINIMAL_HISTORY_CONFIG_MOCK as HistoryOptions, [
```
Tlacenka, line 27: ❔ Why not create `MINIMAL_HISTORY_CONFIG_MOCK` directly as the  type `HistoryOptions`, so that it doesn't have to be overriden everywhere?

packages/utils/src/lib/git.ts
```diff
@@ -37,35 +37,68 @@ export async function toGitPath(
...
-      // @TODO use simple git
-      return await git
-        .raw(['describe', '--tags', '--exact-match'])
-        .then(out => out.trim());
-    }
-  } catch {
-    // Return a custom error message when something goes wrong
-    throw new Error('Could not get current tag or branch.');
+  const branch = await git.branch().then(r => r.current);
+  // eslint-disable-next-line unicorn/prefer-ternary
```
Tlacenka, line 93: ⚙️ ESLint standards: Why is this rule disabled instead of turning this statement to ternary operator?

Tlacenka (commented):
The main thing that needs to be changed before this PR is merged is making the unit and integration tests consistent for the history command.

packages/core/src/lib/history.unit.test.ts
BioPhoton: Thx! implemented

packages/utils/src/lib/git.ts
BioPhoton: removed!

packages/utils/src/lib/git.ts
```diff
@@ -37,36 +37,66 @@ export async function toGitPath(
   return formatGitPath(path, gitRoot);
 }
 
+export class GitStatusError extends Error {
```
matejchalk, line 40: The class contains a lot of logic, so I would suggest it should be covered by unit tests (as suggested by our GitHub Action's annotations :grin:).

image

BioPhoton: How beautiful this comments are! 
🔥 😍 🦚

BioPhoton: Done

matejchalk: I don't see the new tests :confused:

BioPhoton: It was gone for some reason... added the test now.

matejchalk (approved):

Tlacenka (approved):
Thank you for splitting the tests, looking nice!

PR #275, opened 2023-11-16
feat(core): change to execute all plugins before throwing on failed

### LINKED ISSUE #159: plugin execution should be a graceful exit and run all registered plugins before throwing
Use promise settled and log the promise result.

MishaSeredenkoPushBased:
Changed executePlugins() logic to run all plugins, even if one or more plugins have failed. Then console.error the failed plugin errors, and throw.
Closes #159


packages/core/src/lib/implementation/execute-plugin.spec.ts
```diff
@@ -129,6 +129,33 @@ describe('executePlugins', () => {
...
+    const pluginConfig2 = {
+      ...validPluginCfg2,
+      runner: vi.fn().mockResolvedValue([]),
+    };
+    const pluginConfig3 = {
+      ...validPluginCfg,
+      runner: vi.fn().mockRejectedValue('plugin 3 error'),
+    };
+    const plugins = [pluginConfig, pluginConfig2, pluginConfig3];
+    const errorSpy = vi.spyOn(console, 'error');
```
matejchalk, line 146: The spy should be restored, so that `console.error` then reverts to its original implementation.

You could put `errorSpy.mockRestore()` at the end of the `it`-block, but then it will only be run when all assertions pass (otherwise it wouldn't be reached).

A better way is to use hooks, which the test runner will always execute before/after each/all tests:

```ts
describe('test suite', () => {
  let errorSpy: SpyInstance;
  
  beforeEach(() => {
    errorSpy = vi.spyOn(console, 'error');
  });
  
  afterEach(() => {
    errorSpy.mockRestore();
  });
  
  it('test 1', () => ...);

  it('test 2', () => ...);

  // ...
})
```

MishaSeredenkoPushBased: Fixed, thanks

packages/core/src/lib/implementation/execute-plugin.spec.ts
```diff
@@ -105,6 +105,16 @@ describe('executePlugin', () => {
 });
 
 describe('executePlugins', () => {
+  let errorSpy: SpyInstance;
+
+  beforeEach(() => {
+    errorSpy = vi.spyOn(console, 'error');
```
BioPhoton, line 111: Nice! we should replace also other places with this mock. 
ATM we use no mocking techniques in many place.

cc @Tlacenka

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -123,15 +135,43 @@ export async function executePlugins(
...
+        error: e instanceof Error ? e.message : String(e),
+        status: 'rejected',
+      });
+    }
+  }, Promise.resolve([] as ExecutePluginResult[]));
 
   progressBar?.endProgress('Done running plugins');
 
-  return pluginsResult;
+  const errors = pluginsResult.filter(
```
BioPhoton, line 153: We have a helper method [`logMultipleResults`](https://github.com/code-pushup/cli/blob/b071ce9c444f74f81cb71cbd1c089801dbcc0531/packages/utils/src/lib/file-system.ts#L44). It is used here https://github.com/code-pushup/cli/blob/b071ce9c444f74f81cb71cbd1c089801dbcc0531/packages/core/src/lib/implementation/persist.ts#L87

matejchalk (approved):

BioPhoton (changes requested):
We could use already existing helper logic.

packages/core/src/lib/implementation/execute-plugin.spec.ts
MishaSeredenkoPushBased: It's a good idea, but I'd propose it to do outside of the scope of this task

BioPhoton: yes not here

Tlacenka: This will be resolved in #277

packages/core/src/lib/implementation/execute-plugin.ts
MishaSeredenkoPushBased: Moved and reused `logMultipleResults` logic for logging multiple results by `Promise.allSettled`. Reused both at `persist.ts` and `execute-plugin.ts`

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -123,15 +125,49 @@ export async function executePlugins(
...
+    }
+  }, Promise.resolve([] as Promise<ExecutePluginResult>[]));
 
   progressBar?.endProgress('Done running plugins');
 
-  return pluginsResult;
+  const errorsCallback = ({ reason }: PromiseRejectedResult) =>
+    console.error(reason);
+
+  return Promise.allSettled(pluginsResult).then(results => {
```
IKatsuba, line 145: Maybe change to `await` instead of `then`?

```suggestion
const results = await Promise.allSettled(pluginsResult);
```

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -123,15 +125,49 @@ export async function executePlugins(
...
+    logMultipleResults(results, 'Plugins', undefined, errorsCallback);
+
+    const failedResults = results.filter(
```
IKatsuba, line 148: We need something like groupBy here.

```
const {fulfilled, rejected} = groupBy('status', results);
```

I can be more readable and less complex.

packages/core/src/lib/implementation/execute-plugin.ts
BioPhoton: I would intro a helper function encapsulating the logging logic.

packages/core/src/lib/implementation/execute-plugin.ts
BioPhoton: We can move it into a new issue?

here a PoC:

```typescript
function groupBy<T extends Record<Key, any>>(key: Key, array: T[]): Record<string, T[]> {
  return array.reduce((accumulator, currentValue) => {
    const groupKey = currentValue[key];
    if (!accumulator[groupKey]) {
      accumulator[groupKey] = [];
    }
    accumulator[groupKey].push(currentValue);
    return accumulator;
  }, {} as Record<string, T[]>);
}
```

BioPhoton (commented):
left comments mostly nit picking

packages/core/src/lib/implementation/execute-plugin.ts
MishaSeredenkoPushBased: New ticket #287 was created, will be in that scope.

packages/core/src/lib/implementation/execute-plugin.ts
MishaSeredenkoPushBased: Moved logging logic from `LogMultipleResults()` to `logPluginExecution()`, same file.

packages/core/src/lib/implementation/execute-plugin.ts
MishaSeredenkoPushBased: Good point, fixed

packages/utils/src/lib/log-results.ts
```diff
@@ -0,0 +1,53 @@
...
+  succeededCallback?: (result: PromiseFulfilledResult<T>) => void,
+  failedCallback?: (result: PromiseRejectedResult) => void,
+) {
+  if (succeededCallback) {
+    const succeededResults = results.filter(
+      (result): result is PromiseFulfilledResult<T> =>
+        result.status === 'fulfilled',
+    );
+
+    logPluginExecution(
```
BioPhoton, line 18: What I don't understand is why is a specific function `logPluginExecution` inside a generic `logMultipleResults`?

We already have a this implemented correctly here:
- https://github.com/code-pushup/cli/blob/4c4fa0df08304dae4d3ab3edddbc38eb4fdd1848/packages/core/src/lib/implementation/persist.ts#L87
and used here:
- https://github.com/code-pushup/cli/blob/4c4fa0df08304dae4d3ab3edddbc38eb4fdd1848/packages/core/src/lib/collect-and-persist.ts#L20

In `logPluginExecution` we see that we use the generic function `logMultipleResults` and make a specific function out of it.

BioPhoton (changes requested):
Hi! I guess we have a bit of a confusion here with generic vs specific functions and how to reuse generic logic. 

Please have a look at my comments

packages/utils/src/lib/file-system.ts
```diff
@@ -44,33 +45,24 @@ export type MultipleFileResults = PromiseSettledResult<FileResult>[];
 export function logMultipleFileResults(
   persistResult: MultipleFileResults,
```
BioPhoton, line 46:
```suggestion
  fileResults: MultipleFileResults,
```

packages/utils/src/lib/log-results.ts
MishaSeredenkoPushBased: Sorry 'bout that. Will it be fine if I change `logPluginExecution` to generic also?
I added this function basically becaise of the prev comment 'to encapsulate logging logic'.

matejchalk: > What I don't understand is why is a specific function logPluginExecution inside a generic logMultipleResults?

Good point :+1: But I think that's just a naming problem. The implementation of `logPluginExecution` doesn't actually have anything to do with plugins in particular, so the name is misleading.

packages/utils/src/lib/log-results.ts
```diff
@@ -0,0 +1,53 @@
+type PromiseFulfilledResult<T> = {
+  status: 'fulfilled';
+  value: T;
+};
+
```
matejchalk, line 5: This type is already in the global namespace.

```suggestion
```

image

packages/utils/src/lib/log-results.ts
```diff
@@ -0,0 +1,53 @@
...
+): void {
+  if (results.length) {
+    console.log(logMessage);
+    results.forEach(
+      callback as (
+        result: PromiseFulfilledResult<T> | PromiseRejectedResult,
+      ) => void,
+    );
+  }
+}
```
matejchalk, line 53: You can take advantage of the fact that you've already narrowed the types of `succeededResults` and `failedResults` when you call this function. The function parameter types are a little too loose, but you can tighten them up by making the result type a generic type parameter (it'll be narrowed to just one option when you call it), so you then don't need to cast the `callback` in `forEach`.

```suggestion
export function logPluginExecution<
  T extends PromiseFulfilledResult<unknown> | PromiseRejectedResult,
>(results: T[], logMessage: string, callback: (result: T) => void): void {
  if (results.length) {
    console.log(logMessage);
    results.forEach(callback);
  }
}
```

packages/utils/src/lib/log-results.ts
```diff
@@ -0,0 +1,53 @@
...
+  results: PromiseSettledResult<T>[],
+  messagePrefix: string,
+  succeededCallback?: (result: PromiseFulfilledResult<T>) => void,
+  failedCallback?: (result: PromiseRejectedResult) => void,
+) {
+  if (succeededCallback) {
+    const succeededResults = results.filter(
+      (result): result is PromiseFulfilledResult<T> =>
+        result.status === 'fulfilled',
+    );
```
matejchalk, line 16: I think it would be convenient to create a type guard in `utils`:

```ts
export function isPromiseFulfilledResult<T>(
  result: PromiseSettledResult<T>,
): result is PromiseFulfilledResult<T> {
  return result.status === 'fulfilled';
}
```

Then we can reuse whenever we have `Array.filter` on a `PromiseSettledResult<T>[]`:

```suggestion
    const succeededResults = results.filter(isPromiseFulfilledResult);
```

packages/utils/src/lib/log-results.ts
```diff
@@ -0,0 +1,53 @@
...
+      succeededResults,
+      `${messagePrefix} successfully: `,
+      succeededCallback,
+    );
+  }
+
+  if (failedCallback) {
+    const failedResults = results.filter(
+      (result): result is PromiseRejectedResult => result.status === 'rejected',
+    );
```
matejchalk, line 28: Same as for fulfilled results, a type guard in `utils` could be useful:

```ts
export function isPromiseRejectedResult(
  result: PromiseSettledResult<unknown>,
): result is PromiseRejectedResult {
  return result.status === 'rejected';
}
```

```suggestion
    const failedResults = results.filter(isPromiseRejectedResult);
```

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -123,15 +125,50 @@ export async function executePlugins(
...
+  }
+
+  return results
+    .filter(
+      (
+        result: PromiseSettledResult<ExecutePluginResult>,
+      ): result is PromiseFulfilledResult<PluginReport> =>
+        result.status === 'fulfilled',
+    )
+    .map((result: PromiseFulfilledResult<PluginReport>) => result.value);
```
matejchalk, line 171: The type guards I suggest could be useful here as well, I see :grin: 

```suggestion
  const failedResults = results.filter(isPromiseRejectedResult);
  if (failedResults.length) {
    const errorMessages = failedResults
      .map(({ reason }) => reason)
      .join(', ');
    throw new Error(
      `Plugins failed: ${failedResults.length} errors: ${errorMessages}`,
    );
  }

  return results
    .filter(isPromiseFulfilledResult)
    .map(result => result.value);
```

packages/utils/src/lib/log-results.ts
MishaSeredenkoPushBased: Thanks @BioPhoton @matejchalk , renamed function to `logPromiseResults`, I believe it will help to avoid the confusion

packages/utils/src/lib/file-system.ts
MishaSeredenkoPushBased: Fixed

packages/utils/src/lib/log-results.ts
MishaSeredenkoPushBased: Removed

packages/utils/src/lib/log-results.ts
MishaSeredenkoPushBased: Great point, fixed

packages/utils/src/lib/log-results.ts
MishaSeredenkoPushBased: Added guard

packages/core/src/lib/implementation/execute-plugin.ts
MishaSeredenkoPushBased: Done. It really helps both for DRY and readability

packages/utils/src/lib/log-results.ts
MishaSeredenkoPushBased: Done

MishaSeredenkoPushBased:
The comments are fixed, as well as the code was improved for better readability and support.
@BioPhoton @matejchalk your comments were very useful and helpful, thanks!

packages/utils/src/lib/log-results.unit.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+      succeededCallbackMock,
+      failedCallbackMock,
+    );
+
+    expect(succeededCallbackMock).toHaveBeenCalled();
+    expect(failedCallbackMock).toHaveBeenCalled();
+  });
+});
+
+describe('logPluginExecution', () => {
```
matejchalk, line 81:
```suggestion
describe('logPromiseResults', () => {
```

matejchalk (changes requested):

packages/utils/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.1",
   "dependencies": {
     "@code-pushup/models": "*",
+    "vitest": "~0.32.0",
```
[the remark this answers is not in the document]
MishaSeredenkoPushBased: Thanks, because I had no idea why it required the package. Renamed

packages/utils/src/lib/log-results.unit.spec.ts
MishaSeredenkoPushBased: Renamed for all mentions and usages

matejchalk (approved):

BioPhoton (approved):

PR #281, opened 2023-11-16
test(cli): add verdaccio for local publishing and e2e testing

IKatsuba:


e2e/cli-e2e/mocks/fs.mock.ts
```diff
@@ -4,7 +4,7 @@ import { ensureDirectoryExists } from '@code-pushup/utils';
 
 // @TODO move into testing library
 export function cleanFolder<T extends object>(
-  dirName = 'tmp',
+  dirName = 'tmp/cli-e2e',
```
BioPhoton, line 7: Why is this here? If it avoids testing problems in the file system I suggest we go with --persist.filename instead

global-setup.ts
```diff
@@ -1,4 +1,6 @@
 import { mkdir, rm } from 'fs/promises';
+import startLocalRegistry from './tools/scripts/start-local-registry';
```
BioPhoton, line 2: This is unused here. Should this only be imported in global.e2e?

BioPhoton (commented):
Left some comments. I tried it last time and was not able to figure it out. 

Pretty cool you did! 🙌

global-setup.ts
IKatsuba: I will remove it. Thanks

e2e/cli-e2e/mocks/fs.mock.ts
IKatsuba: Verdaccio, by default, saves all data in the tmp directory. But when the test is starting, tmp is already removed. Changing the default dirName is avoiding it.

BioPhoton: > But when the test is starting, tmp is already removed. 

Could it be fixed by adopting `global-setup.ts` ? If not we should maybe document it a bit as I would not have thought about this reason.

BioPhoton (dismissed):
Perfect!!!

BioPhoton (approved):
Lgtm

PR #282, opened 2023-11-18
fix(cli): replace clui on @isaacs/cliui

IKatsuba:
`cliui` has the [issue](https://github.com/yargs/cliui/issues/89) with text wraping. `@isaacs/cliui` does not have it for now


BioPhoton (approved):
Sweet!

PR #283, opened 2023-11-19
ci: add release action

BioPhoton:


IKatsuba:
@BioPhoton, I took a look at this PR. If you need to update the version, I think, tags must be on the main branch.

IKatsuba (dismissed):

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: npm
+      - name: Install dependencies
+        run: npm ci
+      - name: Execute version target on all affected packages and release them
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: npx nx affected --base=last-release --target=version --dryRun
```
matejchalk, line 42: Why does this need a `GITHUB_TOKEN` if it's a dry run? :raised_eyebrow: What does it do?

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: npm ci
+      - name: Execute version target on all affected packages and release them
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: npx nx affected --base=last-release --target=version
+      - name: Tag last-release
+        shell: bash
+        run: |
+          git tag -f last-release
+          git push origin last-release --force
```
matejchalk, line 72: Do we need this `last-release` tag? Why not just use `--base=origin/main`?

matejchalk (changes requested):

.github/workflows/release.yml
IKatsuba: Because this code will run on the origin/main branch. If we use `--base=origin/main`, packages will not be released.

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
+name: Release Packages
+
+on:
+  pull_request:
+    branches:
+      - main
```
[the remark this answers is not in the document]
IKatsuba: It is for dry-run mode. But I will add `push` next to it.

.github/workflows/release.yml
IKatsuba: Nice, thanks. I will remove it

.github/workflows/release.yml
IKatsuba: > Could we rename the tag to `latest`, though? I think that's more common.

Yes, we could!

.github/workflows/release.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: npm
+      - name: Setup Git
```
matejchalk, line 65:
```suggestion
      - name: Set up Git
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          git config user.name "GitHub Bot"
+          git config user.email "gituser@example.com"
+      - name: Install dependencies
+        run: npm ci
+      - name: Execute version target on all affected packages and release them
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: npx nx affected --base=latest --target=version
```
matejchalk, line 74: I suggest we exclude `plugin-lighthouse` and `nx-plugin` for now, since it will still be some time before we're ready to release those.

```suggestion
        # un-exclude packages when they're ready for public release
        run: npx nx affected --base=latest --target=version --exclude=plugin-lighthouse,nx-plugin
```

matejchalk (approved):

Issue #63, opened 2023-09-22
nx plugin - `cli` executor for the code pushup CLI


BioPhoton:
### User story

As a user of Nx I want to have convenient configuration of my code-pushup related target.
A custom executor for the code-pushup CLI would reduce configuration and setup cost.

### Acceptance criteria

- [x] #737
- [x] forward `stdout` and `stderr` from CLI process in executor 
- #740
   - [x] create a `zod` schema `envSchema` to parse environment variables
- [x] ~bundle package in multiple formats - https://nx.dev/recipes/tips-n-tricks/compile-multiple-formats~
   - https://github.com/nrwl/nx/issues/15682#issuecomment-1700298264
   - (possible alternative) https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js/ 👀
- generate a `schema.json` from a `zod` schema in `schema.ts` (optional)

#### Dual Build 

Problems regarding the dual build:
- @poppinss/cliui also doesn't support CJS for some time.
   Possible workaround for @poppinss/cliui would be a dynamic import
- non-CJS-compatible runtime dependencies are lighthouse and lighthouse-logger
- big refactoring as turning all functions async that touch one of the incompatible tools

Conclusion use the CLI in a separate process instead and duplicate some of the logic.

### Implementation details

https://github.com/code-pushup/cli/tree/add-executor-to-nx-plugin 

### Related
- https://github.com/code-pushup/cli/issues/841

PR #294, opened 2023-11-20
feat(nx-plugin): add configuration generator

### LINKED ISSUE #61: nx plugin - `configuration` generator for standalone workspace layout
The nx plugin should have a generator that is responsible for adding the `code-pushup.config.json` to a project in the workspace.

Todo:
- [ ] The project should get added over a CLI argument
- [ ] The project, if not given in the CLI arguments, should show up as select options in the CLI prompts
- [ ] In a standalone workspace layout a script should get added to the root `package.json` 
- [ ] In a common workspace layout (e.g. libs/apps) an error should get thrown that mentions that the feature will be implemented in the future.

IKatsuba:
Close #61


packages/nx-plugin/src/generators/add-to-project/generator.ts
```diff
@@ -0,0 +1,47 @@
...
+    return;
+  }
+
+  generateFiles(tree, path.join(__dirname, 'files'), root, options);
+
+  updateProjectConfiguration(tree, options.project, {
+    ...projectConfiguration,
+    targets: {
+      ...projectConfiguration.targets,
+      collect: {
```
BioPhoton, line 31: Would it make sense to name the target code-pushup? As we can run different commands. 
 (`autorun` is default)

packages/nx-plugin/src/generators/add-to-project/generator.ts
```diff
@@ -0,0 +1,47 @@
...
+        executor: 'nx:run-commands',
```
BioPhoton, line 32: This will turn into a custom executor in a future PR i think?
Good, so it works ATM...

packages/nx-plugin/src/generators/add-to-project/generator.unit.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  let tree: Tree;
+  const testProjectName = 'test-app';
+
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+
+    addProjectConfiguration(tree, testProjectName, {
+      root: `apps/${testProjectName}`,
+      projectType: 'library',
+      sourceRoot: `apps/${testProjectName}/src`,
```
BioPhoton, line 21: We should check for the workspace layout. In our repo e.g. we have no apps/libs but packages/examples/e2e

packages/nx-plugin/generators.json
```diff
@@ -4,6 +4,11 @@
       "factory": "./src/generators/init/generator",
       "schema": "./src/generators/init/schema.json",
       "description": "init generator"
+    },
+    "add-to-project": {
```
BioPhoton, line 8: Maybe we can get inspired by existing generator names from Nx.

Here the vite generators: https://github.com/nrwl/nx/blob/master/packages/vite/generators.json

BioPhoton (commented):
Left small comments regarding naming

packages/nx-plugin/src/generators/add-to-project/generator.ts
IKatsuba: Yes! But let's make the executor a separate issue.

packages/nx-plugin/src/generators/add-to-project/generator.unit.test.ts
IKatsuba: This may need to be clarified. We just read the project's configuration and added the code-pushup configuration to its root. If the project root is the same as the repo root, we add the `code-pushup` config to the repo root. Any other logic here may be extra.

packages/nx-plugin/src/generators/add-to-project/generator.ts
IKatsuba: There is the issue about executor #63

matejchalk (approved):

PR #302, opened 2023-11-22
feat(utils): add audits sorting for reports

### LINKED ISSUE #210: Sorting audits in stdout and MD format
As a CLI only user, I want to make sense of my text output. Currently, I have to skim through the whole list of audits as they are not ordered in any obvious way.

This ordering could be inspired by the one in portal where we also want to ensure simple access to what user needs. We could maybe even create a separate argument which would accept an enumerated value which would be used for the ordering.

## Implementation details
- Primarily based on **weight** in descending order - users will most likely want to prioritise seeing the biggest score contributors.
- Secondarily based on **score** in the ascenading order - errors first.
- Tertiary ordering is based on **value** in descending order - usually bigger value means worse.
- Lastly the **alphabetical** order for the title.

MishaSeredenkoPushBased:
Added audits sorting for both reports, md and stdout, added sorting of audits at the categories and groups.
Closes #210


packages/utils/src/lib/__snapshots__/report-to-stdout.integration.test.ts.snap
```diff
@@ -6,71 +6,71 @@ exports[`report-to-stdout > should contain all sections when using the fixture r
...
+● Require \`let\` or \`const\` instead of \`var\`                           0
+● Require calls to \`isNaN()\` when checking for \`NaN\`                  0
+● Require or disallow \\"Yoda\\" conditions                               0
+● Require using arrow functions for callbacks                         0
 
 
 Lighthouse audits
 
 ● First Contentful Paint                                              1.2 s
 ● Largest Contentful Paint                                            1.5 s
```
Tlacenka, line 70: Why isn't Largest Contentful Paint first? Its value is the highest and I cannot see any score.

packages/utils/src/lib/__snapshots__/report-to-md.integration.test.ts.snap
```diff
@@ -16,106 +16,108 @@ exports[`report-to-md > should contain all sections when using the fixture repor
...
+- 🟩 [Disallow negating the left operand of relational operators](#disallow-negating-the-left-operand-of-relational-operators-eslint) (_eslint_) - **0**
+- 🟩 [Disallow use of optional chaining in contexts where the \`undefined\` value is not allowed](#disallow-use-of-optional-chaining-in-contexts-where-the-undefined-value-is-not-allowed-eslint) (_eslint_) - **0**
+- 🟩 [Require calls to \`isNaN()\` when checking for \`NaN\`](#require-calls-to-isnan-when-checking-for-nan-eslint) (_eslint_) - **0**
+- 🟩 [Enforce comparing \`typeof\` expressions against valid strings](#enforce-comparing-typeof-expressions-against-valid-strings-eslint) (_eslint_) - **0**
+- 🟥 [Require the use of \`===\` and \`!==\`](#require-the-use-of--and--eslint) (_eslint_) - **1 warning**
+- 🟥 [Disallow missing \`key\` props in iterators/collection literals](#disallow-missing-key-props-in-iterators-collection-literals-eslint) (_eslint_) - **1 warning**
+- 🟥 [Disallow missing props validation in a React component definition](#disallow-missing-props-validation-in-a-react-component-definition-eslint) (_eslint_) - **6 warnings**
```
Tlacenka, line 44: The audits within a category seem to be sorted as they were defined in the config.

Maybe there was a misunderstanding - while I agree that the categories themselves should have an order in which they were defined, the audits and groups within them should be sorted based on weight > score > value > name in order not to mix the positive and negative results. What do you think?

packages/utils/src/lib/report.ts
```diff
@@ -161,6 +167,103 @@ export function countCategoryAudits(
...
+  ) as WeighedAuditReport[];
+  const audits = groupAudits.sort(sortCategoryAudits);
+
+  return {
+    ...groupWithAudits,
+    audits,
+  };
+}
+
+export function sortCategoryAudits(
```
Tlacenka, line 233: It would be valuable to document the sorting logic by creating unit tests for it.

packages/utils/src/lib/__snapshots__/report-to-stdout.integration.test.ts.snap
MishaSeredenkoPushBased: As we found out, the sorting is proper here, but the score is not present in the view.

packages/utils/src/lib/report.ts
MishaSeredenkoPushBased: Added tests

packages/utils/src/lib/report-to-md.ts
```diff
@@ -87,14 +95,41 @@ function reportToCategoriesSection(report: ScoredReport): string {
...
+        if (ref.type === 'group') {
+          acc.groups = [
+            ...acc.groups,
+            getGroupWithAudits(ref.slug, ref.plugin, plugins),
+          ];
+        } else {
+          acc.audits = [...acc.audits, getAuditByRef(ref, plugins)];
+        }
+        return { ...acc };
+      },
```
matejchalk, line 115: You don't need to mutate the accumulator here, use a conditional object spread instead:

```suggestion
      ) => ({
        ...acc,
        ...ref.type === 'group' ? {
          groups: [...acc.groups, getGroupWithAudits(ref.slug, ref.plugin, plugins)],
        } : {
          audits: [...acc.audits, getAuditByRef(ref, plugins)],
        },
      }),
```

packages/utils/src/lib/report-to-md.ts
```diff
@@ -87,14 +95,41 @@ function reportToCategoriesSection(report: ScoredReport): string {
...
+      { groups: [], audits: [] },
+    );
+
+    const audits = auditsAndGroups.audits
+      .sort(sortCategoryAudits)
+      .reduce((acc, audit) => {
+        acc += auditItemToCategorySection(audit);
+        acc += NEW_LINE;
+        return acc;
+      }, '');
```
matejchalk, line 125: Doesn't need to be `reduce`, you can use `map` and `join`:

```suggestion
      .map(auditItemToCategorySection)
      .join(NEW_LINE);
```

packages/utils/src/lib/report-to-md.ts
```diff
@@ -87,14 +95,41 @@ function reportToCategoriesSection(report: ScoredReport): string {
...
+    const groups = auditsAndGroups.groups.reduce(
+      (acc: string, group: EnrichedScoredAuditGroupWithAudits) => {
+        acc += groupItemToCategorySection(group);
+        return acc;
+      },
+      '',
+    );
```
matejchalk, line 132: Same here:

```suggestion
    const groups = auditsAndGroups.groups
       .map(groupItemToCategorySection)
       .join('');
```

packages/utils/src/lib/report-to-md.ts
```diff
@@ -105,73 +140,37 @@ function reportToCategoriesSection(report: ScoredReport): string {
...
-  return (
-    li(
-      `${getSquaredScoreMarker(pluginAudit.score)} ${auditTitle} (_${
-        plugin.title
-      }_) - ${getAuditResult(pluginAudit)}`,
-    ) + NEW_LINE
+  return li(
+    `${getSquaredScoreMarker(audit.score)} ${auditTitle} (_${
+      audit.plugin
+    }_) - ${getAuditResult(audit)}`,
```
matejchalk, line 160: This is where you should use the plugin `title` instead of `slug`, I believe (see comment above).

packages/utils/src/lib/report-to-md.ts
```diff
@@ -105,73 +140,37 @@ function reportToCategoriesSection(report: ScoredReport): string {
...
-  ) as EnrichedScoredAuditGroup;
   const groupScore = Number(formatReportScore(group?.score || 0));
-
-  if (!group) {
-    throwIsNotPresentError(`Group ${refSlug}`, plugin?.slug);
-  }
-
   const groupTitle = li(
-    `${getRoundScoreMarker(groupScore)} ${group.title} (_${plugin.title}_)`,
+    `${getRoundScoreMarker(groupScore)} ${group.title} (_${group.plugin}_)`,
```
matejchalk, line 169: Same here, make sure this is the plugin's title, please :pray:

packages/utils/src/lib/report-to-md.ts
```diff
@@ -187,67 +186,68 @@ function groupRefItemToCategorySection(
...
-      acc += h3(auditTitle);
-
-      acc += NEW_LINE;
-      acc += NEW_LINE;
-
-      if (!audit.details?.issues?.length) {
-        acc += detailsTitle;
+  const auditsSection = report.plugins.reduce((acc, plugin) => {
+    const auditsData = plugin.audits
+      .sort((a, b) => sortAudits(a, b))
```
matejchalk, line 191: Can be simplified:

```suggestion
      .sort(sortAudits)
```

packages/utils/src/lib/report-to-stdout.ts
```diff
@@ -73,24 +74,28 @@ function reportToDetailSection(report: ScoredReport): string {
...
-          text: chalk.cyanBright(displayValue || `${value}`),
-          width: 10,
-          padding: [0, 0, 0, 0],
-        },
-      );
-    });
+    audits
+      .sort((a: EnrichedAuditReport, b: EnrichedAuditReport) =>
+        sortAudits(a, b),
+      )
```
matejchalk, line 80:
```suggestion
      .sort(sortAudits)
```

packages/utils/src/lib/report.ts
```diff
@@ -161,6 +167,103 @@ export function countCategoryAudits(
...
+  );
+}
+
+export function getAuditByRef(
+  { slug, weight, plugin }: CategoryRef,
+  plugins: ScoredReport['plugins'],
+): WeighedAuditReport {
+  const auditPlugin = plugins.find(
+    ({ slug }) => slug === plugin,
+  ) as PluginReport;
```
matejchalk, line 185: It would be better to handle this safely. I see you use optional chaining anyway (`auditPlugin?.`).

```suggestion
  const auditPlugin = plugins.find(
    ({ slug }) => slug === plugin,
  );
```

packages/utils/src/lib/report.ts
```diff
@@ -161,6 +167,103 @@ export function countCategoryAudits(
...
+
+export function getGroupWithAudits(
+  refSlug: string,
+  refPlugin: string,
+  plugins: ScoredReport['plugins'],
+): EnrichedScoredAuditGroupWithAudits {
+  const plugin = plugins.find(({ slug }) => slug === refPlugin) as PluginReport;
+  const groupWithAudits = plugin?.groups?.find(
+    ({ slug }) => slug === refSlug,
+  ) as EnrichedScoredAuditGroupWithAudits;
```
matejchalk, line 207: Same here. Using `as T` is just an alternative syntax for `!` in these cases, BTW. I see it defeats our linter, but it's best avoided for the exact same reasons. (I'll look around for a lint rule that covers this :thinking:)

```suggestion
  const plugin = plugins.find(({ slug }) => slug === refPlugin);
  const groupWithAudits = plugin?.groups?.find(
    ({ slug }) => slug === refSlug,
  );
```

matejchalk (changes requested):

packages/utils/src/lib/report.unit.test.ts
```diff
@@ -182,3 +185,86 @@ describe('loadReport', () => {
     ).rejects.toThrow('validation');
   });
 });
+
+describe('audits sorting', () => {
```
Tlacenka, line 189: It would be great to split testing the sort based on individual entity and priority and minimise the number of items in each case to showcase it in the simplest way.

packages/utils/src/lib/__snapshots__/report-to-md.integration.test.ts.snap
```diff
@@ -169,255 +167,257 @@ ESLint rule **curly**. [📖 Docs](https://eslint.org/docs/latest/rules/curly)
...
 
-ESLint rule **require-render-return**, from _react_ plugin. [📖 Docs](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/require-render-return.md)
+ESLint rule **prefer-arrow-callback**. [📖 Docs](https://eslint.org/docs/latest/rules/prefer-arrow-callback)
 
-### First Contentful Paint (Lighthouse)
+### First Contentful Paint (lighthouse)
 
 🟨 <b>1.2 s</b> (score: 76)
 
 [📖 Docs](https://developer.chrome.com/docs/lighthouse/performance/first-contentful-paint/)
```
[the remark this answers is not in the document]
MishaSeredenkoPushBased: Great that you've noticed - lost that during the refactoring. Done

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: Fixed

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: Done

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: Done

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: Fixed

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: Fixed

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: Done

packages/utils/src/lib/report-to-stdout.ts
MishaSeredenkoPushBased: Done

packages/utils/src/lib/report.ts
MishaSeredenkoPushBased: Done

packages/utils/src/lib/report.ts
MishaSeredenkoPushBased: You're correct, removed

packages/utils/src/lib/report.unit.test.ts
MishaSeredenkoPushBased: Added unit tests for both `sortAudits()` and `sortCategoryAudits()`, covered all conditions.

packages/utils/src/lib/report-to-md.ts
```diff
@@ -105,73 +137,46 @@ function reportToCategoriesSection(report: ScoredReport): string {
...
-  const pluginAudit = plugin?.audits.find(({ slug }) => slug === refSlug);
-
-  if (!pluginAudit) {
-    throwIsNotPresentError(`Audit ${refSlug}`, plugin?.slug);
-  }
-
   const auditTitle = link(
-    `#${slugify(pluginAudit.title)}-${slugify(plugin.title)}`,
-    pluginAudit?.title,
+    `#${slugify(audit.title)}-${slugify(audit.plugin)}`,
```
matejchalk, line 154: This should slugify the plugin title:

```suggestion
    `#${slugify(audit.title)}-${slugify(getPluginNameFromSlug(
      audit.plugin,
      plugins,
    ))}`,
```

(:top: Probably best to assign result of `getPluginNameFromSlug` to a variable, since you reuse it.)

packages/utils/src/lib/report-to-md.ts
```diff
@@ -105,73 +137,46 @@ function reportToCategoriesSection(report: ScoredReport): string {
...
-    }
-
-    return acc;
-  }, []);
-
-  const groupAudits = foundAudits.reduce((acc, audit) => {
+  const groupAudits = group.audits.reduce((acc, audit) => {
     const auditTitle = link(
-      `#${slugify(audit.title)}-${slugify(plugin.title)}`,
+      `#${slugify(audit.title)}-${slugify(group.plugin)}`,
```
matejchalk, line 179:
```suggestion
      `#${slugify(audit.title)}-${slugify(getPluginNameFromSlug(group.plugin, plugins))}`,
```

packages/utils/src/lib/report.ts
```diff
@@ -161,6 +166,105 @@ export function countCategoryAudits(
   }, 0);
 }
 
+export function getAuditsFromAllPlugins(
+  report: ScoredReport,
+): EnrichedAuditReport[] {
+  return report.plugins.reduce<EnrichedAuditReport[]>(
+    (acc, plugin) => [...acc, ...plugin.audits],
+    [],
+  );
```
matejchalk, line 175: You can use `flatMap` here:

```suggestion
  return report.plugins.flatMap(plugin => plugin.audits);
```

packages/utils/src/lib/report.unit.test.ts
```diff
@@ -182,3 +190,104 @@ describe('loadReport', () => {
...
+      {
+        slug: 'plugin-b',
+        title: 'Plugin B',
+        date: '',
+        duration: 0,
+        icon: 'todo',
+        audits: [],
+        groups: [],
+      },
+    ];
```
matejchalk, line 289: A small trick I like to do in tests (only in tests, though!) is to do a soft type-cast that lets me omit properties which are required under usual circumstances, but aren't relevant to the test. The advantage is that it removes the extra noise from the test.

```suggestion
    const plugins = [
      {
        slug: 'plugin-a',
        title: 'Plugin A',
      },
      {
        slug: 'plugin-b',
        title: 'Plugin B',
      },
    ] as ScoredReport['plugins'];
```

This is just a suggestion, though. It's still a small test, so the extra noise doesn't matter so much.

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: Reused the plugin title

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: Done

packages/utils/src/lib/report.ts
MishaSeredenkoPushBased: Removed that function - as it was no more used due to the code optimization

packages/utils/src/lib/report.unit.test.ts
MishaSeredenkoPushBased: Awesome hack! Implemented for all new tests, as they also contain mocks

packages/utils/src/lib/report.unit.test.ts
```diff
@@ -182,3 +190,88 @@ describe('loadReport', () => {
...
+describe('sortCategoryAudits', () => {
+  it('should sort audits by weight and score', () => {
+    const mockAudits = [
+      { weight: 0, score: 0.1 },
+      { weight: 5, score: 1 },
+      { weight: 0, score: 0.7 },
+      { weight: 10, score: 1 },
+    ] as WeighedAuditReport[];
+    const sortedAudits = [...mockAudits].sort(sortCategoryAudits);
+    expect(sortedAudits[0]).toEqual(mockAudits[3]);
```
Tlacenka, line 203: It would increase the readability of the tests if you asserted the whole `sortedAudits` array and included the expected values directly.

matejchalk (approved):

packages/utils/src/lib/report.unit.test.ts
MishaSeredenkoPushBased: I haven't quite understood the proposition 🙂 
Would you like me to compare the whole array?
```suggestion
    expect(sortedAudits).toEqual([
      { weight: 10, score: 1 },
      { weight: 5, score: 1 }, // ...
    ]);
```
Or probably you'd like the values to be compared?
```suggestion
    expect(sortedAudits[0].weight).toEqual(10);
    expect(sortedAudits[0].score).toEqual(1); // comparing the values directly
```

Tlacenka: The whole array, yes. Then the person reading the test can tell straight away if the sorting result is valid (sees the values below each other).

MishaSeredenkoPushBased: Done

matejchalk (approved):

Tlacenka (approved):
Well done!

PR #303, opened 2023-11-22
build: lint .spec.ts files to report filename errors

matejchalk:
So that we can catch mistakes like this:

image

Context: https://github.com/code-pushup/cli/pull/275#discussion_r1401776075


Tlacenka (approved):
Adding a probable inconsistent format widely used for test files is a good idea.

Issue #211, opened 2023-11-09
CLI package test refactor


Tlacenka:
As a contributor to Code PushUp, I would like for the tests to be consistently structured and follow good practices.
There are several areas in which I identified room for improvement. In this issue, the following targeted changes should be made.

## Coverage
From what I understood, all the `command-object` tests test the same thing - parsing arguments with respective command object. The parsing itself is then tested in `yargs-cli` and configuration parsing in `config-middleware`. Hence, there is no need for repeated tests of the same thing in multiple test files.
- [x] For each command object, only test a happy path.
- [x] Additionally, for command objects using `upload` command, make sure that the configuration and all upload related values are correctly entered.
- [x] Move the argument override test to `yargs-cli` where a few use cases should be created to illustrate different scenarios.
- [x] The main file for testing the configuration is the `config-middleware`. Add different types of configurations (with groups, categories, plugins) and check their main properties there.

## Files
These tests are **integration tests** which check the integration of our CLI with `yargs`. Hence ,they should be categorised as integration tests and structured accordingly.
**Note**: `nx` configuration of integration vs unit tests is OOS. This is only about the location and naming of the files.
- [x] Each test file name reflects that it is an integration test (to be finalised).
- [x] Rename all test files to reflect the command being tested in them. Each CLI command (`autorun`, `upload`, ...). Currently, all are called `command-object.spec.ts` which makes it hard to read without the full path.
- [x] All configurations and mocks for FS and other should be moved under `testing-utils`. The `test` folder is overloaded with unstructured files.
- [x] Each configuration should serve a definitive purpose. I am unable to distinguish between `minimal`, `all-values` and `cli-parsing` config, they seem interchangeable for the purpose of checking the arguments parsing. They should be merged into one (unless I am missing something).
- [x] Research exit process in tests  
#243

## Good practices
In order to create and maintain tests easily, their readability should be prioritised. This means minimising boiler plate code, keeping the test flow linear and naming things based on their purpose.
- [x] `baseArgs` and `objectToCliArgs` should disappear and be passed hardcoded in an array instead.
- [x] Use `toThrow(<message>)` to check that a test fails and returns expected error message (in `config-middleware`).
- [x] Remove leftover assertions which would make any value pass (`yargs-cli`, checking the `config` in the last test case).
- [x] Remove duplicate tests (global options default and global option parsed in `yargs-cli`).
- [x] Fix type error for currently compulsory `filename` within `persist` configuration property.
[...]

PR #306, opened 2023-11-22
test(cli): clean up unit and e2e for CLI commands

Tlacenka:
Related to #211 

This PR covers only part of the test refactor. Creating a shared setup, unifying mocks and cleaning up the rest of the CLI package tests is yet to come.

In this PR I did the following:
- Reduced purpose of unit tests only to the relevant CLI command.
- Reduced boiler-plate code in tests.
- Used `fs` and console mocks to test the command behaviour in unit tests.
- Reduced the number of E2E tests for `print-config` (covered already in unit tests).
- Fixed `filterKebabCaseKeys` and updated its unit tests.


packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -0,0 +1,117 @@
...
+});
+vi.mock('fs/promises', async () => {
+  const memfs: typeof import('memfs') = await vi.importActual('memfs');
+  return memfs.fs.promises;
+});
+
+vi.mock('@code-pushup/core', async () => {
+  const core = await vi.importActual('@code-pushup/core');
+  return {
+    ...(core as object),
```
matejchalk, line 41: Tiny thing, but we can easily make the type more specific.

```suggestion
  const core: typeof import('@code-pushup/core') =
    await vi.importActual('@code-pushup/core');
  return {
    ...core,
```

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -0,0 +1,117 @@
...
+});
+
+describe('autorun-command', () => {
+  beforeEach(async () => {
+    vi.clearAllMocks();
+    vol.reset();
+    vol.fromJSON(
+      {
+        'my-report.json': JSON.stringify(report()),
+        'code-pushup.config.ts': '', // only needs to exist for stat inside bundleRequire
```
matejchalk, line 64:
```suggestion
        'code-pushup.config.ts': '', // only needs to exist for stat inside readCodePushupConfig
```

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    await vi.importActual('@code-pushup/models/testing');
+  return {
+    bundleRequire: vi
+      .fn()
+      .mockResolvedValue({ mod: { default: minimalConfig() } }),
+  };
+});
+
+describe('autorun-command', () => {
+  beforeEach(async () => {
```
matejchalk, line 58: There's no `await`, so no need for `async`. (We have a lint rule for this, but errors need fixing first.)

```suggestion
  beforeEach(() => {
```

packages/cli/src/lib/yargs-cli.integration.test.ts
```diff
@@ -1,69 +1,104 @@
...
-    const args: string[] = ['--config=./config.a.ts'];
-    const parsedArgv = await yargsCli(args, {
-      options,
-    }).parseAsync();
+  it('should parse a single config argument as a string', async () => {
+    const parsedArgv = await yargsCli<GeneralCliOptions>(
+      ['--config=./config.a.ts'],
+      {
+        options,
+      },
```
matejchalk, line 31: Tiny formatting improvement, saves some real estate.

The trick is to put the 1st property on the same line as `{` before you format, then Prettier will try to put it on one line. But if the 1st property is already on another line, Prettier always formats it as multiline.

```suggestion
      { options },
```

matejchalk (dismissed):
I :heart: this refactor :tada:

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
Tlacenka: When I tried to do that, I got the `Static imports of lazy-loaded libraries are forbidden` lint error :(

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
Tlacenka: Good point, fixed!

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
Tlacenka: Nice catch!

packages/cli/src/lib/yargs-cli.integration.test.ts
Tlacenka: I didn't know about this, that's a game changer 😮

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
matejchalk: Oh, interesting :D OK, then leave it as is, it doesn't matter very much in this case.

matejchalk (approved):

PR #308, opened 2023-11-23
test(cli): clean up mocks, migrate reusable utils to testing-utils

Tlacenka:
Related to #211 

This PR is the second out of three PRs concerned with cleaning up CLI testing. In it I do the following:
- I split up `onlyPlugins` logic to a separate file
- I improved the quality of `config-middleware` and `onlyPlugins` tests
- I removed all newly unused mocks and renamed the CLI testing utility folder to `mocks`
- `config-middleware` test config files have been moved to `testing-utils`
- other CLI tests also use configuration files from `testing-utils` or a specially created `CORE_CONFIG_MOCK` and `REPORT_MOCK` to be reused

There is still one PR to go which will create shared setup for mocked parts of tests and add it to the vitest config.


packages/cli/test/minimal-plugin-process.mock.mjs
matejchalk, line 1: Not sure if this should be deleted. I think it was being used for manually testing the progress bar.

Should at least be moved to `testing-utils`, though.

testing-utils/src/lib/fixtures/configs/core-config.mock.ts
```diff
@@ -0,0 +1,73 @@
+import { CoreConfig } from '../../../../../packages/models/src';
+
+export const CORE_CONFIG_MOCK = {
```
matejchalk, line 3: I would keep the "minimal" in the name, so that we distinguish it from the more realistic test fixtures we have in `models`.

matejchalk (commented):
Overall a very good cleanup :+1:

packages/cli/test/minimal-plugin-process.mock.mjs
Tlacenka: Yeah that's the problem with some of the test mocks - I could not find any instance where this was used anymore but maybe you are right (?) not sure really.

testing-utils/src/lib/fixtures/configs/core-config.mock.ts
Tlacenka: It is not really minimal as it has plugins and audits filled in. A minimal config would only have the required values of a minimum extent, so most arrays would be empty.

matejchalk: So maybe we could name the others "example" (e.g. `EXAMPLE_REPORT`)?

packages/cli/test/minimal-plugin-process.mock.mjs
Tlacenka: I looked at the original PR and apparently, it is in [here](https://github.com/code-pushup/cli/blob/main/packages/core/test/plugin-progress-bar-plugin-process.mock.mjs)

matejchalk: I see. So maybe this is truly unused, then :man_shrugging:

testing-utils/src/lib/fixtures/configs/core-config.mock.ts
Tlacenka: For the report, I will use `MINIMAL_REPORT` because it does only have minimum values. As for the others, once I get to `models` and other packages, I intend to unify as many of the configs as possible, so we will see then.

matejchalk (approved):

PR #310, opened 2023-11-23
refactor(cli): remove 'stdout' option from report output formats

### LINKED ISSUE #247: Remove `stdout` as format argument
Extract of #191
The stdout format should be removed. We will always have some output, but some of it will be behind --verbose flag.

## Current behavior

The report summary is printed to stdout based on the `persist.format` value (set in `code-pushup.config.ts` or via `--persist.format` CLI argument).

The default `format` is `['json', 'stdout']`. It is not possible to skip `'json'` (we create it anyway regardless, it's needed for the upload). However, if a user wants to get a Markdown report as well, they can easily omit stdout without realizing it. E.g. if they add `--persist.format=md`, then it removes `stdout` because the array is replaced.

## Proposed change

We print the report summary regardless, and `stdout` will be removed from possible `format` values. If the user specifies `--persist.format=md`, then they will still see the report summary in the stdout.

The results of audits and categories are the most relevant information for the user when they run `collect`, and the report summary is minimal by design.

IKatsuba:
Close #247

### DIFF
--- a/packages/core/src/lib/implementation/persist.integration.test.ts
+++ b/packages/core/src/lib/implementation/persist.integration.test.ts
@@ -85,7 +85,7 @@ describe('persistReport', () => {
     expect(() => readReport('md')).toThrow('no such file or directory');
   });
 
-  it('should log to console when format is stdout`', async () => {
+  it('should log to console regardless of format`', async () => {
     const persist = persistConfig({ outputDir, format: [] });
 
     await persistReport(dummyReport, {
--- a/packages/core/src/lib/implementation/persist.integration.test.ts
+++ b/packages/core/src/lib/implementation/persist.integration.test.ts
@@ -86,7 +86,7 @@ describe('persistReport', () => {
   });
 
   it('should log to console when format is stdout`', async () => {
-    const persist = persistConfig({ outputDir, format: ['stdout'] });
+    const persist = persistConfig({ outputDir, format: [] });
 
     await persistReport(dummyReport, {
       ...dummyConfig,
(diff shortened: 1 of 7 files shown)


packages/core/src/lib/implementation/persist.integration.test.ts
```diff
@@ -86,7 +86,7 @@ describe('persistReport', () => {
   });
 
   it('should log to console when format is stdout`', async () => {
```
matejchalk, line 88:
```suggestion
  it('should log to console regardless of format`', async () => {
```

packages/core/src/lib/implementation/persist.ts
```diff
@@ -31,12 +31,10 @@ export async function persistReport(
   const outputDir = persist.outputDir;
   const filename = persist.filename;
   const format =
-    persist.format && persist.format.length !== 0 ? persist.format : ['stdout'];
-  let scoredReport;
-  if (format.includes('stdout')) {
-    scoredReport = scoreReport(report);
-    console.log(reportToStdout(scoredReport));
-  }
+    persist.format && persist.format.length !== 0 ? persist.format : [];
```
matejchalk, line 34: I believe this is now equivalent:

```suggestion
    persist.format ?? [];
```

matejchalk (approved):

packages/core/src/lib/implementation/persist.ts
IKatsuba: Let's try it)

matejchalk (approved):

PR #314, opened 2023-11-24
docs(cli): update CLI documentation in README

### LINKED ISSUE #241: Document CLI command in the README manually
Example is in pushbased/userflow in the top of the readme where all commands and options are listed. 

https://github.com/push-based/user-flow#cli

IKatsuba:
Close #241

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -148,12 +148,12 @@ Run commands over:
 Description:  
 Run Plugins and collect results.
 
-| Option                    | Type                 | Default | Description                                          |
-| ------------------------- | -------------------- | ------- | ---------------------------------------------------- |
-| **`--onlyPlugins`**       | `string[]`           | `[]`    | Only run the specified plugins.                      |
-| **`--persist.outputDir`** | `string`             | n/a     | Directory for the produced reports.                  |
-| **`--persist.filename`**  | `string`             | n/a     | Filename for the produced reports without extension. |
-| **`--persist.format`**    | `('json' \| 'md')[]` | `json`  | Format(s) of the report file.                        |
+| Option                    | Type                 | Default  | Description                                          |
+| ------------------------- | -------------------- | -------- | ---------------------------------------------------- |
+| **`--onlyPlugins`**       | `string[]`           | `[]`     | Only run the specified plugins.                      |
+| **`--persist.outputDir`** | `string`             | n/a      | Directory for the produced reports.                  |
+| **`--persist.filename`**  | `string`             | `report` | Filename for the produced reports without extension. |
+| **`--persist.format`**    | `('json' \| 'md')[]` | `json`   | Format(s) of the report file.                        |
 
 #### `upload` command
 
@@ -163,15 +163,15 @@ Run commands over:
 Description:
 Upload reports to the Code PushUp portal.
 
-| Option                      | Type                 | Default | Description                                          |
-| --------------------------- | -------------------- | ------- | ---------------------------------------------------- |
-| **`--persist.outputDir`**   | `string`             | n/a     | Directory for the produced reports.                  |
-| **`--persist.filename`**    | `string`             | n/a     | Filename for the produced reports without extension. |
-| **`--persist.format`**      | `('json' \| 'md')[]` | `json`  | Format(s) of the report file.                        |
-| **`--upload.organization`** | `string`             | n/a     | Organization slug from portal.                       |
-| **`--upload.project`**      | `string`             | n/a     | Project slug from portal.                            |
-| **`--upload.server`**       | `string`             | n/a     | URL to your portal server.                           |
-| **`--upload.apiKey`**       | `string`             | n/a     | API key for the portal server.                       |
+| Option                      | Type                 | Default  | Description                                          |
+| --------------------------- | -------------------- | -------- | ---------------------------------------------------- |
+| **`--persist.outputDir`**   | `string`             | n/a      | Directory for the produced reports.                  |
+| **`--persist.filename`**    | `string`             | `report` | Filename for the produced reports without extension. |
+| **`--persist.format`**      | `('json' \| 'md')[]` | `json`   | Format(s) of the report file.                        |
+| **`--upload.organization`** | `string`             | n/a      | Organization slug from portal.                       |
+| **`--upload.project`**      | `string`             | n/a      | Project slug from portal.                            |
+| **`--upload.server`**       | `string`             | n/a      | URL to your portal server.                           |
+| **`--upload.apiKey`**       | `string`             | n/a      | API key for the portal server.                       |
 
 #### `autorun` command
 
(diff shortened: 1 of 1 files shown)


packages/cli/README.md
```diff
@@ -127,3 +127,86 @@ jobs:
...
+
+# CLI
+
+## Global Options
+
+| Option           | Type      | Default                 | Description                                                            |
+| ---------------- | --------- | ----------------------- | ---------------------------------------------------------------------- |
+| **`--progress`** | `boolean` | `true`                  | Show progress bar in stdout.                                           |
+| **`--verbose`**  | `boolean` | `false`                 | When true creates more verbose output. This is helpful when debugging. |
+| **`--config`**   | `string`  | `code-pushup.config.js` | Path the the config file, e.g. code-pushup.config.js                   |
```
BioPhoton, line 139:
```suggestion
| **`--config`**   | `string`  | `code-pushup.config.js` | Path to the config file, e.g. code-pushup.config.js                   |
```

BioPhoton (changes requested):

BioPhoton:
Left comment to fix typo

packages/cli/README.md
```diff
@@ -127,3 +127,86 @@ jobs:
       - run: npm ci
       - run: npx code-pushup autorun --upload.apiKey=${{ secrets.PORTAL_API_KEY }}
 ```
+
+# CLI
```
matejchalk, line 131: There is already a heading level 1 (`#`) at the top of this document, so this should be level 2 (`##`), and the sub-sections should then be nested accordingly (`##` :arrow_right: `###`, etc.).

(We could integrate [markdownlint](https://github.com/DavidAnson/markdownlint) to these catch typograhic errors, BTW. I personally use the VSCode extension.)

Also, maybe "CLI reference" or "CLI commands and options" would help distinguish this section from what's above? :thinking:

packages/cli/README.md
```diff
@@ -127,3 +127,86 @@ jobs:
...
+`code-pushup collect [options]`
+
+Description:  
+Run Plugins and collect results.
+
+| Option                    | Type       | Default | Description                                     |
+| ------------------------- | ---------- | ------- | ----------------------------------------------- |
+| **`--onlyPlugins`**       | `string[]` | `[]`    | Only run the specified plugins.                 |
+| **`--persist.outputDir`** | `string`   | n/a     | Directory for the produced reports.             |
+| **`--persist.filename`**  | `string`   | n/a     | Filename for the produced reports.              |
```
matejchalk, line 155: Maybe good to specify "without extension"?

packages/cli/README.md
```diff
@@ -127,3 +127,86 @@ jobs:
...
+| **`--persist.format`**    | `enum`     | n/a     | Format of the report output. e.g. `md`, `json`. |
```
matejchalk, line 156:
```suggestion
| **`--persist.format`**    | `('json' | 'md')[]`     | 'json'     | Format(s) of the report file. |
```

packages/cli/README.md
matejchalk, line 155:
```suggestion
| **`--persist.filename`**  | `string`   | `'report'`    | Filename for the produced reports.              |
```

packages/cli/README.md
```diff
@@ -127,3 +127,86 @@ jobs:
...
+Run commands over:
+`code-pushup collect [options]`
+
+Description:  
+Run Plugins and collect results.
+
+| Option                    | Type       | Default | Description                                     |
+| ------------------------- | ---------- | ------- | ----------------------------------------------- |
+| **`--onlyPlugins`**       | `string[]` | `[]`    | Only run the specified plugins.                 |
+| **`--persist.outputDir`** | `string`   | n/a     | Directory for the produced reports.             |
```
matejchalk, line 154: Maybe out of scope of this PR, but why don't we specify a default location for `outputDir`? E.g. `'.code-pushup'`.

packages/cli/README.md
```diff
@@ -127,3 +127,86 @@ jobs:
...
+| **`--upload.server`**       | `string` | n/a     | URL to your portal server.                      |
+| **`--upload.apiKey`**       | `string` | n/a     | API key for the portal server.                  |
+
+### `autorun` command
+
+Run commands over:
+`code-pushup autorun [options]`
+
+Description:
+Run Plugins, collect results and upload reports to the Code PushUp portal.
```
matejchalk, line 182:
```suggestion
Run plugins, collect results and upload report to the Code PushUp portal.
```

packages/cli/README.md
```diff
@@ -127,3 +127,86 @@ jobs:
...
+
+## Commands
+
+### `collect` command
+
+Run commands over:
+`code-pushup collect [options]`
+
+Description:  
+Run Plugins and collect results.
```
matejchalk, line 149:
```suggestion
Run plugins and collect results.
```

packages/cli/README.md
```diff
@@ -127,3 +127,86 @@ jobs:
...
+
+| Option                      | Type     | Default | Description                                     |
+| --------------------------- | -------- | ------- | ----------------------------------------------- |
+| **`--persist.outputDir`**   | `string` | n/a     | Directory for the produced reports.             |
+| **`--persist.filename`**    | `string` | n/a     | Filename for the produced reports.              |
+| **`--persist.format`**      | `enum`   | n/a     | Format of the report output. e.g. `md`, `json`. |
+| **`--upload.organization`** | `string` | n/a     | Organization slug from portal.                  |
+| **`--upload.project`**      | `string` | n/a     | Project slug from portal.                       |
+| **`--upload.server`**       | `string` | n/a     | URL to your portal server.                      |
+| **`--upload.apiKey`**       | `string` | n/a     | API key for the portal server.                  |
```
matejchalk, line 174: Do we need to specify these 4 times? They're always the same, so it feels like a waste of space.

Instead, I would explain that these options are read from the config file, so the CLI arguments act as an override.

packages/cli/README.md
BioPhoton: Also thought about it!

packages/cli/README.md
IKatsuba: Agree. 

Create issue #320 about markdownlint

packages/cli/README.md
IKatsuba: @matejchalk created #321

packages/cli/README.md
```diff
@@ -127,3 +127,82 @@ jobs:
...
+| **`--persist.outputDir`**   | `string`             | n/a      | Directory for the produced reports.                  |
+| **`--persist.filename`**    | `string`             | `report` | Filename for the produced reports without extension. |
+| **`--persist.format`**      | `('json' \| 'md')[]` | `json`   | Format(s) of the report file.                        |
+| **`--upload.organization`** | `string`             | n/a      | Organization slug from portal.                       |
+| **`--upload.project`**      | `string`             | n/a      | Project slug from portal.                            |
+| **`--upload.server`**       | `string`             | n/a      | URL to your portal server.                           |
+| **`--upload.apiKey`**       | `string`             | n/a      | API key for the portal server.                       |
+
+> [!NOTE]  
+> All common options can be specified in the configuration file as well. CLI arguments take precedence over configuration file options.
```
matejchalk, line 154: Perfect :ok_hand: Also, I don't know GitHub's Markdown supports this `[!NOTE]` syntax, very nice :+1:

packages/cli/README.md
```diff
@@ -127,3 +127,82 @@ jobs:
...
+
+### Commands
+
+#### `collect` command
+
+Run commands over:
```
matejchalk, line 160: I don't think this grammar works :confused: How about just "Usage:"?

matejchalk (approved):

Tlacenka (commented):
Taking unhelpful line break and expected skill set of an end user into consideration, is the "this is useful when debugging" necessary to include?

packages/cli/README.md
```diff
@@ -127,3 +127,82 @@ jobs:
...
+| **`--progress`** | `boolean` | `true`                  | Show progress bar in stdout.                                           |
+| **`--verbose`**  | `boolean` | `false`                 | When true creates more verbose output. This is helpful when debugging. |
+| **`--config`**   | `string`  | `code-pushup.config.js` | Path to the config file, e.g. code-pushup.config.js                    |
+
+### Common Command Options
+
+| Option                      | Type                 | Default  | Description                                          |
+| --------------------------- | -------------------- | -------- | ---------------------------------------------------- |
+| **`--persist.outputDir`**   | `string`             | n/a      | Directory for the produced reports.                  |
+| **`--persist.filename`**    | `string`             | `report` | Filename for the produced reports without extension. |
```
Tlacenka, line 146: I noticed you are using reports in plural in several places. As the ClI only produces one report at a time, I think a singular "report" should be used.

packages/cli/README.md
```diff
@@ -127,3 +127,82 @@ jobs:
...
+
+### Commands
+
+#### `collect` command
+
+Run commands over:
+`code-pushup collect [options]`
+
+Description:  
+Run plugins and collect results.
```
Tlacenka, line 164: If I read the description without the previous knowledge of the project, I would not understand which command to use to produce a report - I suggest adding that this command produces a report based on the collected results as well.

packages/cli/README.md
matejchalk: There is only one report, but can be multiple report files (`report.json` and `report.md`).

BioPhoton (approved):
LGTM

PR #315, opened 2023-11-24
docs(examples-plugins): add example plugin file-size

BioPhoton:
This PR includes:

- add project `example-plugins`
- add project & lint;test targets
  - add memfs constant to `@code-pushup/testing-utils`
- add file size plugin
  - add unit tests
  - add integration tests


vmasek:
Good example 👍🏻

examples/plugins/.eslintrc.json
```diff
@@ -0,0 +1,39 @@
+{
+  "extends": ["../../.eslintrc.json"],
+  "ignorePatterns": ["!**/*"],
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+      "rules": {
+        "@nx/enforce-module-boundaries": ["off"]
```
matejchalk, line 8: Why turn this off?

examples/plugins/src/file-size.plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+type RunnerOptions = PluginOptions;
+
+export const pluginSlug = 'file-size';
+
+const fileSizeAuditSlug = 'file-size-check';
+export const auditsMap = {
+  [fileSizeAuditSlug]: {
+    slug: fileSizeAuditSlug,
+    title: 'File Size Audit',
+    description: 'A audit to check JavaScript file size in a directory.',
```
matejchalk, line 28:
```suggestion
    description: 'An audit to check JavaScript file size in a directory.',
```

examples/plugins/src/file-size.plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+  return `File ${basename(filePath)} is OK. (size: ${formatBytes(size)})`;
+}
+
+export function errorMessage(filePath: string, size: number, budget: number) {
+  const sizeDifference = formatBytes(size - budget);
+  const byteSize = formatBytes(size);
+  const byteBudget = formatBytes(budget);
+  return `File ${basename(
+    filePath,
+  )} has ${byteSize} this is ${sizeDifference} too big. (budget: ${byteBudget})`;
```
matejchalk, line 182:
```suggestion
  return `File ${basename(
    filePath,
  )} has ${byteSize}, this is ${sizeDifference} too big. (budget: ${byteBudget})`;
```

examples/plugins/src/file-size.plugin.unit.test.ts
```diff
@@ -0,0 +1,357 @@
...
+    [2, 1],
+  ])('should return error message', (size, budget) => {
+    expect(errorMessage('test.js', size, budget)).toBe(
+      `File ${file} has ${size} B this is ${1} B too big. (budget: ${budget} B)`,
+    );
+  });
+});
+
+describe('assertFileSize', () => {
+  it.each([[-1], [0], [1]])(
```
matejchalk, line 68: Based on the [`test.each` docs](https://jestjs.io/docs/api#1-testeachtablename-fn-timeout), this can be simplified:
> If you pass in a 1D array of primitives, internally it will be mapped to a table i.e. `[1, 2, 3] -> [[1], [2], [3]]`

```suggestion
  it.each([-1, 0, 1])(
```

examples/plugins/src/file-size.plugin.unit.test.ts
```diff
@@ -0,0 +1,357 @@
...
+    const str = 'Hello World'
+    const num = 42;
+    const obj = ${projectJson};
+  `;
+const readmeMd = '# Docs';
+
+const file = 'test.js';
+
+describe('infoMessage', () => {
+  it.each([['index.js'], [join('src', 'index.js')]])(
```
matejchalk, line 46:
```suggestion
  it.each(['index.js', join('src', 'index.js')])(
```

examples/plugins/src/file-size.plugin.unit.test.ts
```diff
@@ -0,0 +1,357 @@
...
+    'should return info message',
```
matejchalk, line 47: You should reference the test arguments in the title, otherwise you generate duplicate test titles:

```suggestion
    'should return info message for file %s',
```

examples/plugins/src/file-size.plugin.unit.test.ts
```diff
@@ -0,0 +1,357 @@
...
+      );
+    },
+  );
+});
+
+describe('errorMessage', () => {
+  it.each([
+    [1, 0],
+    [2, 1],
+  ])('should return error message', (size, budget) => {
```
matejchalk, line 60:
```suggestion
  ])('should return error message for size %i with budget %i', (size, budget) => {
```

matejchalk (commented):
Looks good :slightly_smiling_face: I like that all the tests are using `memfs`, should be much more stable :grin:

examples/plugins/.eslintrc.json
BioPhoton: because in the custom plugins I rely on direct imports ATM
`import { formatBytes, pluralize } from '../../../dist/packages/utils';`

When we have published the package we can change them

examples/plugins/src/file-size.plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+      severity = 'error';
+      message = errorMessage(file, size, budget);
+    }
+  }
+  // return Issue
+  return {
+    message,
+    severity,
+    source: {
+      file,
```
matejchalk, line 209: You can use the [`toUnixPath` function](https://github.com/code-pushup/cli/blob/main/packages/utils/src/lib/file-system.ts#L7-L18) from `utils`, this is what's [used by the ESLint plugin](https://github.com/code-pushup/cli/blob/main/packages/plugin-eslint/src/lib/runner/lint.ts#L19) to normalize file paths.

```suggestion
      file: toUnixPath(file, { toRelative: true }),
```

Then you wouldn't have the absolute path included:
image

BioPhoton: done

examples/plugins/src/file-size.plugin.unit.test.ts
```diff
@@ -0,0 +1,361 @@
...
+
+// Mock file system API's
+vi.mock('fs', async () => {
+  const memfs: typeof import('memfs') = await vi.importActual('memfs');
+  return memfs.fs;
+});
+vi.mock('fs/promises', async () => {
+  const memfs: typeof import('memfs') = await vi.importActual('memfs');
+  return memfs.fs.promises;
+});
```
Tlacenka, line 24: If you merge current `main` to your branch, you can set up shared mocks for the `fs` by adding path `../../testing-utils/src/lib/setup/fs.mock.ts` to `setupFiles` in `vite.config.unit.ts`.

examples/plugins/src/file-size.plugin.unit.test.ts
```diff
@@ -0,0 +1,361 @@
...
+    'should return correct score (files: %s, errors: %s, score: %s)',
+    (files, errors, score) => {
+      expect(scoreFilesizeAudit(files, errors)).toBe(score);
+    },
+  );
+});
+
+describe('fileSizePlugin', () => {
+  beforeEach(() => {
+    vol.reset();
```
Tlacenka, line 127: Similarly to the fs mocks, you can also include `reset.mocks.ts` from the `testing-utils` `setup` folder that automatically resets the volume before each test.

examples/plugins/project.json
```diff
@@ -0,0 +1,38 @@
...
+      "outputs": ["{options.outputFile}"],
+      "options": {
+        "lintFilePatterns": ["examples/plugins/**/*.ts"]
+      },
+      "dependsOn": ["^build"]
+    },
+    "unit-test": {
+      "executor": "@nx/vite:test",
+      "outputs": ["{options.reportsDirectory}"],
+      "options": {
```
Tlacenka, line 19: Unfortunately, after nx upgrade, the default targets were removed so the vitest `config` option needs to be explicitly added for each test target.

examples/plugins/src/file-size.plugin.integration.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  2,
+);
+const testJs = `
+    const str = 'Hello World'
+    const num = 42;
+    const obj = ${projectJson};
+  `;
+
+describe('create', () => {
+  const baseOptions: PluginOptions = {
```
Tlacenka, line 43: There is no need for a variable, the `directory` option can be passed directly into each test.

examples/plugins/src/file-size.plugin.unit.test.ts
```diff
@@ -0,0 +1,361 @@
...
+  2,
+);
+const testJs = `
+    const str = 'Hello World'
+    const num = 42;
+    const obj = ${projectJson};
+  `;
+const readmeMd = '# Docs';
+
+const testFile = 'test.js';
```
Tlacenka, line 43: I would use the `test.js` value directly, no need for an additional variable.

examples/plugins/src/file-size.plugin.integration.test.ts
```diff
@@ -0,0 +1,117 @@
...
+vi.mock('fs', async () => {
+  const memfs: typeof import('memfs') = await vi.importActual('memfs');
+  return memfs.fs;
+});
+vi.mock('fs/promises', async () => {
+  const memfs: typeof import('memfs') = await vi.importActual('memfs');
+  return memfs.fs.promises;
+});
+
+const outputDir = MEMFS_VOLUME;
```
Tlacenka, line 24:
```suggestion
const outputDir = '/test';
```
I think it is perfectly fine to use the value directly, even by removing the `outputDir` variable itself.

Tlacenka (commented):
Overall I think you did a great job.

Most of my feedback is regarding helper variables. Ideally, one would read the test top to bottom and understand the values used without the need to scroll up or look up the value of a variable. The main value of test comes from its readability and maintainability. As tests are independent of each other, as long as within one test the value is consistent, it is fine to use hardcoded values.

examples/plugins/src/file-size.plugin.integration.test.ts
BioPhoton: as we have gitignore settings i thought it is useful to have that folder shared. otherwise it required changing `.gitignore` too for each different folder.

examples/plugins/src/file-size.plugin.ts
```diff
@@ -1,18 +1,18 @@
...
   CategoryRef,
   Issue,
-  IssueSeverity,
   PluginConfig,
 } from '../../../packages/models/src';
+import {
+  crawlFileSystem,
+  factorOf,
+  pluralizeToken,
+} from '../../../packages/utils/src';
```
matejchalk, line 15:
```suggestion
} from '@code-pushup/models';
import {
  crawlFileSystem,
  factorOf,
  pluralizeToken,
} from '@code-pushup/utils';
```

The `@nx/enforce-module-boundaries` would've caught this. Maybe better to add an `eslint-disable-next-line` comment where needed, instead of turning the rule off for the whole project?

examples/plugins/src/file-size.plugin.integration.test.ts
matejchalk: > as we have gitignore settings i thought it is useful to have that folder shared. otherwise it required changing `.gitignore` too for each different folder.

Aren't you confusing it with `tmp` folder? That's the one we use in tests with the real file system and is in `.gitignore`. But this is a virtual memfs volume, and the name `/test` is just a convention.

testing-utils/src/lib/setup/fs.mock.ts
```diff
@@ -1,5 +1,6 @@
 import { vi } from 'vitest';
 
+console.log('setup fs mock with memfs');
```
matejchalk, line 3:
```suggestion
```

matejchalk (approved):

testing-utils/src/lib/setup/fs.mock.ts
BioPhoton: 🤦‍♂️

matejchalk (approved):

Issue #277, opened 2023-11-16
Console mocking


Tlacenka:
In most tests, the console can be mocked. The `spyOn` will be used and implemented in a shared setup for all tests.

## Implementation details
The convention is:
- We use `console.error`, `console.warn` and `console.info` for expected information into the console.
- `console.log` is used for debugging.

PR #335, opened 2023-11-29
test(cli): set up mocks, disable exit process for tests

### LINKED ISSUE #211: CLI package test refactor

### LINKED ISSUE #243: Research exit process in tests
Spike for this issue is 3h. 

The output should be documenting the result and deciding when we want to use it. 

http://yargs.js.org/docs/#api-reference-exitprocessenable

related to #211

Tlacenka:
Closes #211 

In this PR, I extract and configure mock setup for the CLI package:
- unnecessary global setup is removed from both CLI integration and unit tests
- `fs` and `fs/promises` used in almost all unit tests.
- `console.log` is now mocked for all unit tests.
- `bundleRequire` used by command objects is also extracted.
- `portal-client`, while used by two command objects, is also extracted as it could be reused.
- I also reset mocks and `memfs` volume before each test.

Additionally, for better debugging, I disabled `exitProcess` in `yargsCli` which closes #243

### DIFF
--- a/packages/cli/mocks/constants.ts
+++ b/packages/cli/mocks/constants.ts
@@ -8,4 +8,5 @@ export const DEFAULT_CLI_CONFIGURATION = {
   options,
   middlewares,
   commands,
+  noExitProcess: true, // exiting process suppresses error message
 };
--- a/packages/cli/vite.config.integration.ts
+++ b/packages/cli/vite.config.integration.ts
@@ -12,6 +12,5 @@ export default defineConfig({
     },
     environment: 'node',
     include: ['src/**/*.integration.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
-    globalSetup: ['global-setup.ts'],
   },
 });
--- a/testing-utils/src/lib/setup/console-log.mock.ts
+++ b/testing-utils/src/lib/setup/console-log.mock.ts
@@ -1,0 +1,11 @@
+import { SpyInstance, afterEach, beforeEach, vi } from 'vitest';
+
+let consoleLogSpy: SpyInstance;
+
+beforeEach(() => {
+  consoleLogSpy = vi.spyOn(console, 'log');
+});
+
+afterEach(() => {
+  consoleLogSpy.mockRestore();
+});
--- a/packages/cli/src/lib/autorun/autorun-command.unit.test.ts
+++ b/packages/cli/src/lib/autorun/autorun-command.unit.test.ts
@@ -1,39 +1,12 @@
 import { bundleRequire } from 'bundle-require';
 import { vol } from 'memfs';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import {
-  PortalUploadArgs,
-  ReportFragment,
-  uploadToPortal,
-} from '@code-pushup/portal-client';
+import { PortalUploadArgs, uploadToPortal } from '@code-pushup/portal-client';
 import { collectAndPersistReports } from '@code-pushup/core';
 import { MINIMAL_REPORT_MOCK } from '@code-pushup/testing-utils';
 import { DEFAULT_CLI_CONFIGURATION } from '../../../mocks/constants';
 import { yargsCli } from '../yargs-cli';
 import { yargsAutorunCommandObject } from './autorun-command';
-
-// This in needed to mock the API client used inside the upload function
-vi.mock('@code-pushup/portal-client', async () => {
-  const module: typeof import('@code-pushup/portal-client') =
-    await vi.importActual('@code-pushup/portal-client');
-
-  return {
-    ...module,
-    uploadToPortal: vi.fn(
-      async () => ({ packageName: '@code-pushup/cli' } as ReportFragment),
-    ),
-  };
-});
-
-// Mock file system API's
-vi.mock('fs', async () => {
-  const memfs: typeof import('memfs') = await vi.importActual('memfs');
-  return memfs.fs;
-});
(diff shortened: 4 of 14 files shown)


testing-utils/src/lib/setup/console-log.mock.ts
```diff
@@ -0,0 +1,11 @@
...
+
+let consoleLogSpy: SpyInstance;
+
+beforeEach(() => {
+  consoleLogSpy = vi.spyOn(console, 'log');
+});
+
+afterEach(() => {
+  consoleLogSpy.mockRestore();
+});
```
matejchalk, line 11: This should be changed to spy on `console.info`, `console.warn` and `console.error`, but not `console.log`. Otherwise debugging tests will be very difficult, because nothing will be printed.

But we 1st need to replace all our intentional `console.log`s to `console.info` (and turn on the `no-console` ESLint rule). I would suggest this be prioritized very soon. Perhaps even before this is merged?

packages/cli/vite.config.integration.ts
```diff
@@ -12,6 +12,5 @@ export default defineConfig({
     },
     environment: 'node',
     include: ['src/**/*.integration.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
-    globalSetup: ['global-setup.ts'],
```
matejchalk, line 15: Is the global setup not needed anymore? It takes care of cleaning `tmp` folder for tests using real file system.

matejchalk (commented):
Very good cleanup :+1:

packages/cli/vite.config.integration.ts
Tlacenka: No, it is not needed. And I intend to get rid of the `tmp` folder altogether if possible.

The only test using file system is `config-middleware` which only reads configuration files.
`yargs-cli` integration tests should not actively use a file system, only parse parameters.

testing-utils/src/lib/setup/console-log.mock.ts
Tlacenka: This will be tackled in issue #277 and it affects all packages so I wouldn't do it in this PR which only targets the CLI package. But I can prioritise that issue next 👍

matejchalk (approved):

BioPhoton (changes requested):
I wonder if we can now add better testing on errors as we should be able to see them...

packages/cli/mocks/constants.ts
```diff
@@ -8,4 +8,5 @@ export const DEFAULT_CLI_CONFIGURATION = {
   options,
   middlewares,
   commands,
+  noExitProcess: true,
```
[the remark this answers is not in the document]
Tlacenka: Comment added both in `yargs-cli` and the `DEFAULT_CLI_CONFIGURATION` constant.

Tlacenka:
> I wonder if we can now add better testing on errors as we should be able to see them...
@BioPhoton Presumably, even more so when I change `console.log` messages that belong to `console.error` in #277 .

PR #337, opened 2023-11-29
refactor: replace console.log with info, warn and error, introduce shared mock setup

### LINKED ISSUE #277: Console mocking

Tlacenka:
Closes #277 

In this PR, I go through all packages and replace all occurrences of `console.log` with `console.info`, `console.warn` or `console.error`.

Additionally, I set up all `vitest` configuration files for both unit and integration tests to use console mocks.
- Exception: `multi-progress-bars` changes the behaviour of `console` [see here](https://github.com/kamiyo/multi-progress-bars/blob/5fa31e07930d3140b5f5c1aca78caa3ad34bd435/src/virtual-console.ts#L76). Currently, none of the utils integration tests use a console so until I get to that package refactor, I simply did not set up console mocks for integration tests there. But we may have to **create a special type of a test file** for such cases. What do you think?


packages/core/src/lib/implementation/persist.integration.test.ts
```diff
@@ -56,29 +53,19 @@ const resetFiles = async () => {
...
 describe('persistReport', () => {
   beforeEach(async () => {
     resetFiles();
-    setupConsole();
   });
 
-  afterEach(() => {
-    teardownConsole();
-  });
+  afterEach(() => {});
```
matejchalk, line 63: Can now be removed :slightly_smiling_face: 

```suggestion
```

matejchalk (commented):
Excellent, great to have this in place :+1: Could also make sure the `no-console` lint rule is no longer disabled in any project? :pray:

packages/core/src/lib/implementation/persist.integration.test.ts
Tlacenka: Oops 😄

BioPhoton:
🔥

matejchalk (approved):
:tada:

PR #340, opened 2023-11-30
feat(plugin-eslint): add support for inline eslint config

### LINKED ISSUE #331: Support inline config in ESLint plugin
When configuring the Code PushUp ESLint plugin for an Nx monorepo, it is inconvenient that only an eslintrc file path can be provided to the plugin.

In a monorepo where each project has it's own eslintrc, the joined config has to be created on the fly. This currently necessitates writing the config to a temporary eslintrc file for the plugin to be able to consume it. 

```ts
const eslintConfig = // ... complex logic to create it ...

// workaround because eslintPlugin expects file
await fs.writeFile('tmp-eslintrc.json', JSON.stringify(eslintConfig));

export default {
  plugins: [
    await eslintPlugin({
      eslintrc: 'tmp-eslintrc.json', <-- only path supported here
      // ...
    }),
  ]
}
```

This pollutes the file system and is unnecessarily complex for the user. If the `eslintPlugin` function supported passing a config object, this can be avoided.

```ts
const eslintConfig = // ... complex logic to create it ...

export default {
  plugins: [
    await eslintPlugin({
      eslintrc: eslintConfig, // provide the object directly
      // ...
    }),
  ]
}
```

Another advantage of the inline config is that it simplifies standalone projects which just want to use our recommended rules, because we can do that in a single `code-pushup.config.ts` file (we could generate this as default):

```ts
export default {
  plugins: [
    await eslintPlugin({
      eslintrc: {
         extends: '@code-pushup'
      },
      patterns: 'src/**/*.js'
    }),
  ]
}
```

### Acceptance criteria
- [ ] accept `ESLint.Config` object as alternative to file path in ESLint plugin initializer
- [ ] copy or create eslintrc in temp folder (same as audit ouputs) so runner can execute it later

matejchalk:
Closes #331


vmasek (approved):

BioPhoton (approved):
love it!!

Tlacenka (approved):

PR #342, opened 2023-12-01
feat(plugin-eslint): provide helpers for Nx monorepos

### LINKED ISSUE #332: Provide helpers for using ESLint plugin with Nx projects
When configuring the Code PushUp ESLint plugin for an Nx monorepo, there is currently quite a lot of scripting required to set it up properly.

In a monorepo where each project has it's own eslintrc, the joined config has to be created on the fly, using a combination of `@nx/devkit` and `overrides` in the ESLint config.

```ts
// find Nx projects with lint target
const graph = await nx.createProjectGraphAsync({ exitOnError: true });
const projects = Object.values(
  nx.readProjectsConfigurationFromProjectGraph(graph).projects,
).filter(project => 'lint' in (project.targets ?? {}));

// create single ESLint config with override for each project
const eslintConfig: Linter.Config = {
  root: true,
  ignorePatterns: ["**/*"],
  overrides: projects.map(project => ({
    files: project.targets?.lint.options.lintFilePatterns,
    extends: `./${project.root}/.eslintrc.json`,
  })),
};

// combine lint file patterns
const patterns = projects.flatMap(
  project => project.targets?.lint.options.lintFilePatterns ?? [],
);
```

Moreover, with Nx there are 2 different kinds of monorepos to consider:
1. The above configuration will run ESLint on all projects at once, which is desired for monorepos like our CLI or portal, where the **monorepo represents a single product and has a single team working on it**.
2. But for a **monorepo where multiple teams work on multiple products**, it would be desired to only target a single Nx project (e.g. some application) in a `code-pushup.config.ts`. But this project is only an entry point and likely imports from other projects, and those should be linted as well, which again necessitates some fairly complex logic to figure out the ESLint config from Nx's dependency graph.

To simplify the configuration, we could provide helper functions to simplify both use cases:
1. For single product/team monorepos, we would find all Nx projects.
    ```ts
   import eslintPlugin, { eslintConfigFromNxProjects } from '@code-pushup/eslint-plugin';

    export default {
       plugins: [
         await eslintPlugin(
           await eslintConfigFromNxProjects()
         )
       ]
    }
    ```
2. For multi product/team monorepos, we would receive a single target project and find all its dependencies.
    ```ts
   import eslintPlugin, { eslintConfigFromNxProject } from '@code-pushup/eslint-plugin';

    export default {
       plugins: [
         await eslintPlugin(
           await eslintConfigFromNxProject('backoffice')
         )
       ]
    }
    ```

### Acceptance critera

- [ ] implement function that creates ESLint plugin parameters for all Nx projects
- [ ] implement function that creates ESLint plugin parameters from specified Nx project and its dependencies
- [ ] use inline ESLint config once support is added - see #331 
- [ ] list `@nx/devkit` as optional peer dependency in `@code-pushup/eslint-plugin` package
- [ ] use new Nx helpers to simplify CLI's `code-pushup.config.ts`

matejchalk:
Closes #332 

- implements helpers
- lots of tests
  - extended Nx monorepo in mocks for integration test scenarios 
- added docs to plugin README
- used in our CI

### DIFF
--- a/packages/plugin-eslint/package.json
+++ b/packages/plugin-eslint/package.json
@@ -3,8 +3,16 @@
   "version": "0.1.1",
   "dependencies": {
     "@code-pushup/utils": "*",
+    "@code-pushup/models": "*",
     "eslint": "~8.46.0",
-    "zod": "^3.22.4",
-    "@code-pushup/models": "*"
+    "zod": "^3.22.4"
+  },
+  "peerDependencies": {
+    "@nx/devkit": "^17.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@nx/devkit": {
+      "optional": true
+    }
   }
 }
--- a/packages/plugin-eslint/src/lib/nx.integration.test.ts
+++ b/packages/plugin-eslint/src/lib/nx.integration.test.ts
@@ -3,7 +3,7 @@ import { dirname, join } from 'path';
 import { fileURLToPath } from 'url';
 import type { SpyInstance } from 'vitest';
 import { ESLintPluginConfig } from './config';
-import { eslintConfigFromNxProjects } from './nx';
+import { eslintConfigFromNxProject, eslintConfigFromNxProjects } from './nx';
 
 describe('Nx helpers', () => {
   let cwdSpy: SpyInstance;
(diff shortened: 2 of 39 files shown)


packages/plugin-eslint/package.json
```diff
@@ -3,8 +3,16 @@
...
   "dependencies": {
     "@code-pushup/utils": "*",
+    "@code-pushup/models": "*",
     "eslint": "~8.46.0",
-    "zod": "^3.22.4",
-    "@code-pushup/models": "*"
+    "zod": "^3.22.4"
+  },
+  "peerDependencies": {
+    "@nx/devkit": "^17.0.0"
```
BioPhoton, line 11: could nx related code life in the nx plugin?

packages/plugin-eslint/package.json
```diff
@@ -3,8 +3,16 @@
...
+  },
+  "peerDependenciesMeta": {
+    "@nx/devkit": {
+      "optional": true
```
IKatsuba, line 15: @nx/devkit can not be optional. Trying to import something from the `plugin-eslint` package will throw an error like "module not found". To fix it, the package has to have an additional entry point like `@code-pushup/eslint-plugin/nx`, or `nx/devkit` has to be imported lazy.

packages/plugin-eslint/package.json
matejchalk: Good question. The code is actually both the ESLint- and Nx-related, so could live in `plugin-eslint`, `nx-plugin` or even `utils`. 

Since we need to release this feature very soon, I wouldn't put it in `nx-plugin` for the time being, simply because unlike `plugin-eslint` it's not ready for release.

I can see us moving this logic out of `plugin-eslint` at some point in the future and probably generalizing it for the CLI as a whole, but I believe it will require some time to figure out how to do that properly. In order to reach our milestone of integrating Code PushUp on our Nx projects ASAP, this is the most straightforward solution I can think of.

vmasek (approved):

packages/plugin-eslint/src/lib/nx.integration.test.ts
```diff
@@ -0,0 +1,146 @@
...
+     *    │
+     *    ▼
+     *   core
+     *    │        nx-plugin
+     *    │           │
+     *    ▼           │
+     *   utils ◄──────┘
+     */
+
+    const allProjects = ['cli', 'core', 'nx-plugin', 'utils'] as const;
```
Tlacenka, line 108:
```suggestion
    const ALL_PROJECTS = ['cli', 'core', 'nx-plugin', 'utils'] as const;
```

packages/plugin-eslint/src/lib/nx/utils.unit.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  beforeAll(() => {
+    cwdSpy = vi.spyOn(process, 'cwd').mockReturnValue(MEMFS_VOLUME);
+  });
+
+  afterAll(() => {
+    cwdSpy.mockRestore();
+  });
+
+  it('should find code-pushup.eslinrc.json', async () => {
```
Tlacenka, line 17: You could add a test case for `.json` vs `.js` preference.

Tlacenka (approved):
Love it! 🦄 🌈

PR #344, opened 2023-12-04
fix(plugin-eslint): truncate rule texts to pass models validations

matejchalk:
Ran into validation problems when trying to integrate rules from Angular ESLint. This unblocks https://github.com/code-pushup/portal/issues/151


Tlacenka (approved):
Good idea to limit the length.

PR #346, opened 2023-12-05
feat(utils): add sorting of audit issues for report md

### LINKED ISSUE #313: sorting of issues is not given `report.md` is off
Issues should be sorted by severity and then location (line number). Make sure that the sorting logic is not implemented within the scored report. This means sorted in the report .json already. 
Screenshot 2023-11-24 at 22 03 23

MishaSeredenkoPushBased:
Added sorting of audit issues for `report.md`:
- by severity (errors first), 
- by source file (alphabetical),
- by start line (ascending).

Closes #313

### DIFF
--- a/packages/utils/src/lib/report-to-md.ts
+++ b/packages/utils/src/lib/report-to-md.ts
@@ -34,6 +34,7 @@ import {
   reportHeadlineText,
   reportMetaTableHeaders,
   reportOverviewTableHeaders,
+  sortAuditIssues,
   sortAudits,
   sortCategoryAudits,
 } from './report';
@@ -216,7 +217,7 @@ function reportToAuditsSection(report: ScoredReport): string {
 
       const detailsTableData = [
         detailsTableHeaders,
-        ...audit.details.issues.map((issue: Issue) => {
+        ...audit.details.issues.sort(sortAuditIssues).map((issue: Issue) => {
           const severity = `${getSeverityIcon(issue.severity)} <i>${
             issue.severity
           }</i>`;
--- a/packages/utils/src/lib/report-to-md.ts
+++ b/packages/utils/src/lib/report-to-md.ts
@@ -21,6 +21,7 @@ import {
 import {
   FOOTER_PREFIX,
   README_LINK,
+  compareIssues,
   countCategoryAudits,
   detailsTableHeaders,
   formatReportScore,
@@ -34,7 +35,6 @@ import {
   reportHeadlineText,
   reportMetaTableHeaders,
   reportOverviewTableHeaders,
-  sortAuditIssues,
   sortAudits,
   sortCategoryAudits,
 } from './report';
@@ -217,7 +217,7 @@ function reportToAuditsSection(report: ScoredReport): string {
 
       const detailsTableData = [
         detailsTableHeaders,
-        ...audit.details.issues.sort(sortAuditIssues).map((issue: Issue) => {
+        ...audit.details.issues.sort(compareIssues).map((issue: Issue) => {
           const severity = `${getSeverityIcon(issue.severity)} <i>${
             issue.severity
           }</i>`;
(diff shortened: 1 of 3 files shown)


packages/utils/src/lib/report.ts
```diff
@@ -278,3 +279,22 @@ export function getPluginNameFromSlug(
     plugins.find(({ slug: pluginSlug }) => pluginSlug === slug)?.title || slug
   );
 }
+
+export function sortAuditIssues(a: Issue, b: Issue): number {
```
matejchalk, line 283: Since it's not technically sorting, but comparing (i.e. intended to be passed as `compareFn` to `Array.sort`), then this would be a slightly better name, I think:

```suggestion
export function compareIssues(a: Issue, b: Issue): number {
```

packages/utils/src/lib/report.unit.test.ts
```diff
@@ -223,3 +229,37 @@ describe('getPluginNameFromSlug', () => {
...
+    ] as Issue[];
+    const sortedIssues = [...mockIssues].sort(sortAuditIssues);
+    expect(sortedIssues).toEqual([
+      { severity: 'info', source: { file: 'a', position: { startLine: 2 } } },
+      { severity: 'info', source: { file: 'b', position: { startLine: 1 } } },
+      { severity: 'info', source: { file: 'b', position: { startLine: 2 } } },
+      { severity: 'info', source: { file: 'c', position: { startLine: 1 } } },
+    ]);
+  });
+});
```
matejchalk, line 265: Maybe good to also add a test case where there's a file but no position (it's optional after all), to document how we handle it (I assume missing start lines goes first).

packages/utils/src/lib/report-to-md.ts
```diff
@@ -216,7 +217,7 @@ function reportToAuditsSection(report: ScoredReport): string {
 
       const detailsTableData = [
         detailsTableHeaders,
-        ...audit.details.issues.map((issue: Issue) => {
+        ...audit.details.issues.sort(sortAuditIssues).map((issue: Issue) => {
```
BioPhoton, line 220: Can we add the sorting to an earlier stage so all generated reports benefit.

packages/utils/src/lib/report.unit.test.ts
```diff
@@ -223,3 +229,37 @@ describe('getPluginNameFromSlug', () => {
...
+    const sortedIssues = [...mockIssues].sort(sortAuditIssues);
+    expect(sortedIssues).toEqual([
+      { severity: 'error', source: { file: 'a' } },
+      { severity: 'error', source: { file: 'c' } },
+      { severity: 'warning', source: { file: 'b' } },
+      { severity: 'info', source: { file: 'b' } },
+    ]);
+  });
+
+  it('should sort issues by source file and source start line', () => {
```
BioPhoton, line 250: We could split this test into 2 as there could be issues only maintain a file but no start line

BioPhoton (changes requested):
I left comments regarding tests. Over all I like the pr a lot. Will make it easier to read the report.

packages/utils/src/lib/report.unit.test.ts
Tlacenka: Isn't that tested in the first test? L234-L248.

Tlacenka: But there could be one where some of the issues have the `position` and some don't.

packages/utils/src/lib/report.ts
MishaSeredenkoPushBased: Renamed

packages/utils/src/lib/report.unit.test.ts
MishaSeredenkoPushBased: Expanded the sorting conditions and tests to cover all cases.

Tlacenka (approved):
LGTM!

BioPhoton (approved):
LGTM

PR #349, opened 2023-12-05
docs: update special targets

BioPhoton:


Tlacenka (approved):
I didn't realise there was another place where I should change it, thank you for noticing!

PR #358, opened 2023-12-12
refactor(models): add default value for outputDir

### LINKED ISSUE #321: Specify a default location for outputDir
Extract comment from [here](https://github.com/code-pushup/cli/pull/314#discussion_r1405125782)

---

As user, I don't want to have to think about specify configuration options when sensible defaults are available. While being customize the output directory for reports can be useful, for most cases some default location is fine.

- [x] set default value for `persist.outputDir` to `'.code-pushup'`
- [x] config files without an explicit output directory should be valid

BioPhoton:
I tried to reduce the scope as much as possible. In addition I added comments in code and here to understand the changes better and have and easier review process. Hopefully this makes up the messy change.

This PR includes:

- refine options
  - tests
  - separate functions for upload and persist 
- model changes
- tests
- refactored options, config and defaults 
- moved filename into defaults in config middle ware  
  (formats follow in #360 )

**Resources:**
Overview of the target architecture can be found here: https://github.com/code-pushup/cli/wiki/Research:-Module-architecture-and-responsibilities

closes #321 

## Future refactoring

As a proper solution is outside of this PR I suggest we discuss the following change:

1. Introduce the following function to the codebase of cli package
```ts
import {Argv, Options} from "yargs";
import {readCodePushupConfig} from "@code-pushup/core";

export async function withAsyncConfig<T>(cli: Argv<T>, optionsName = 'config', options: Options = {}) {
  const args = await cli.options({[optionsName]: options}).parseAsync();
  const configJson = await readCodePushupConfig(args[optionsName] as string);
  return cli.config(configJson);
}
```

2. Make `cli` and `yargsCli` async and use the configure `yargs``

```ts
export async function yargsCli<T = unknown>(
  argv: string[],
  cfg: {
    // ...
    options?: { [key: string]: Options } & { config: Options };
  },
): Promise<Argv<T>> {
  // ...
  const cli = yargs(argv);

  const configuredCli = await withAsyncConfig(cli, 'config', options['config'])
  
  // ...
}
```


packages/models/src/lib/persist-config.ts
```diff
@@ -5,7 +5,7 @@ export const formatSchema = z.enum(['json', 'md']);
 export type Format = z.infer<typeof formatSchema>;
 
 export const persistConfigSchema = z.object({
-  outputDir: filePathSchema('Artifacts folder'),
+  outputDir: filePathSchema('Artifacts folder').default('.code-pushup'),
```
Tlacenka, line 8: I believe it would be better to add the default value directly to `yargs` which would also provide this information in the `help` command, making it more accessible for the users.

BioPhoton: done

packages/cli/src/lib/implementation/config-middleware.ts
```diff
@@ -27,9 +32,22 @@ export async function configMiddleware<
         ...importedRc.upload,
         ...cliOptions.upload,
       },
+      // we can't use a async rc file as yargs does not support it. see: https://github.com/yargs/yargs/issues/2234
```
BioPhoton, line 35: In essence this is what made the whole PR a bis messy... I have an idea on how to fix it and will bring it up next meeting.

packages/core/src/lib/collect-and-persist.unit.test.ts
```diff
@@ -25,6 +29,32 @@ vi.mock('./implementation/persist', () => ({
...
+      {
+        packageName: 'code-pushup',
+        version: '0.0.1',
+        date: expect.stringMatching(ISO_STRING_REGEXP),
+        duration: 0,
+        categories: [],
+        plugins: [],
+      },
+      normalizePersistConfig(partialConfig?.persist),
+    );
```
matejchalk, line 53: The relevant assertion regarding persist defaults (i.e. the 2nd parameter) kinda gets lost in the surrouding noise (the 1st parameter takes the focus - 8 lines vs 1 line). It's better to be more explicit about the relevant parts and less explicit about what's irrelevant to this scenario.

```suggestion
    expect(persistReport).toHaveBeenCalledWith(
      expect.any(Object),
      {
        outputDir: PERSIST_OUTPUT_DIR,
        filename: PERSIST_FILENAME,
        format: PERSIST_FORMAT,
      },
    );
```

packages/core/src/lib/upload.ts
```diff
@@ -15,18 +20,14 @@ export async function upload(
   options: UploadOptions,
   uploadFn: typeof uploadToPortal = uploadToPortal,
 ) {
-  if (options?.upload === undefined) {
-    throw new Error('upload config needs to be set');
-  }
-  const { apiKey, server, organization, project } = options.upload;
-  const { outputDir, filename } = options.persist;
+  const persist = normalizePersistConfig(options?.persist);
+  const upload = normalizeUploadConfig(options?.upload);
```
matejchalk, line 24: As mentioned above, we can't supply defaults for the upload. The original implementation should stay in place here:

```suggestion
  if (!options?.upload) {
    throw new Error('upload config needs to be set');
  }
```

matejchalk (changes requested):
Thanks for the explanations. The persist config handling now looks good :+1: But the same strategy cannot apply to the upload config, as described in comments.

packages/cli/src/lib/implementation/config-middleware.ts
matejchalk: OK, but honestly I like this implementation :grin: It's very explicit, therefore easy to understand how it works :+1:

packages/core/src/lib/types.ts
```diff
@@ -1,6 +1,33 @@
...
+} from '@code-pushup/models';
+
 export type GlobalOptions = {
   //  'Show progress bar in stdout'
   progress: boolean;
   //  Outputs additional information for a run'
   verbose: boolean;
 };
+
+export const normalizePersistConfig = (
```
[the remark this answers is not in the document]
BioPhoton: done

packages/core/src/lib/types.ts
```diff
@@ -1,6 +1,33 @@
...
+    ...cfg,
+  } as unknown as Required<PersistConfig>);
+
+export const normalizeUploadConfig = (
+  cfg?: Partial<UploadConfig>,
+): Required<UploadConfig> =>
+  ({
+    server: UPLOAD_SERVER,
+    ...cfg,
+  } as unknown as Required<UploadConfig>);
```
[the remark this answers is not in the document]
BioPhoton: deleted

packages/core/src/lib/collect-and-persist.unit.test.ts
BioPhoton: signal/noice... THX!

matejchalk (changes requested):

matejchalk (approved):

Issue #352, opened 2023-12-07
Implement report sorting for audits, groups, categories on earlier stage


MishaSeredenkoPushBased:
We should think of sorting of the report content on earlier stage to make all reports more consistent.
Currently, the sorting is implemented separately between `md` and `stdout`, reusing same methods for comparing from `report.ts` (still consistent). But if we shall add or modify sorting at any place, we need to remember about the rest.
I would propose to move the sorting into the beginning of `persistReport()` method at `persist.ts`, into the place:

```
export async function persistReport(
  report: Report,
  config: CoreConfig,
): Promise<MultipleFileResults> {
  const { persist } = config;
  const outputDir = persist.outputDir;
  const filename = persist.filename;
  const format = persist.format ?? [];

  let scoredReport = scoreReport(report); // sorting should occur here, before the reports generation

  console.info(reportToStdout(scoredReport)); // stdout report generation
  ...
  if (format.includes('md')) {
    scoredReport = scoredReport || scoreReport(report);
    const commitData = await getLatestCommit();
    if (!commitData) {
      console.warn('no commit data available');
    }
    results.push({
      format: 'md',
      content: reportToMd(scoredReport, commitData), // md report generation
    });
  }
  ...
```

PR #360, opened 2023-12-12
fix(cli): fix json format handling

### LINKED ISSUE #316: inconsistency in formats
# Wanted behaviour

The following behaviour is only relevant for the `collect`, `print-config` and `autorun` command.

The following pieces are involved:

- CLI argument parsing is defined as [yargsCoreConfigOptionsDefinition](https://github.com/code-pushup/cli/blob/5d70e4d7857be25123724457dd3fc677b6fc71cb/packages/cli/src/lib/implementation/core-config-options.ts#L5C20-L5C20) and used in [options](https://github.com/code-pushup/cli/blob/5d70e4d7857be25123724457dd3fc677b6fc71cb/packages/cli/src/lib/options.ts#L5)
- Loading remote configuration form a potential `code-pushup.config.ts` file is defined as [yargsCoreConfigOptionsDefinition](https://github.com/code-pushup/cli/blob/5d70e4d7857be25123724457dd3fc677b6fc71cb/packages/cli/src/lib/implementation/core-config-options.ts#L5) and used in [options](https://github.com/code-pushup/cli/blob/5d70e4d7857be25123724457dd3fc677b6fc71cb/packages/cli/src/lib/options.ts#L6)
- Command argument parsing located and configured in the respective command
- Command execution. The logic mostly lives in the [`cli`](https://github.com/code-pushup/cli/tree/main/packages/core) package.

Each layer can overwrite the previose one.

```mermaid
flowchart TB;

   Terminal-->CLI-Argument-Parsing;
   CLI-Argument-Parsing-->Load-Rc-Config;
   Load-Rc-Config-->Command-Argument-Parsing;
   Command-Argument-Parsing-->Command-Execution;

   click CLI-Argument-Parsing "./" "view source"

```
_[docs for the diagram](http://mermaid.js.org/syntax/flowchart.html)_

## Model

The model will default the `format` value of to 'json'.

## `print-config` and `collect` command

We will respect the user input. This is save as there is no upload done.

**Cases:**
- `undefined` in RC -> `json`
- `md,json` in RC -> `md,json`
- `md,json` in RC, `md` in terminal -> `md`

## `autorun` command

We always add `json` to the configured formats as the JSON version is needed for the upload

**Cases:**
- `undefined` in RC -> `json`
- `md,json` in RC -> `md,json`
- `md,json` in RC, `md` in terminal -> `md,json`

BioPhoton:
I added more infos to the issue: https://github.com/code-pushup/cli/issues/316 also #358 should go before

This PR includes:

- adopting model
- adjusting tests

Overview of the target architecture can be found here: https://github.com/code-pushup/cli/wiki/Research:-Module-architecture-and-responsibilities

closes #316


packages/cli/src/lib/collect/collect-command.unit.test.ts
```diff
@@ -51,6 +51,32 @@ describe('collect-command', () => {
...
+        commands: [yargsCollectCommandObject()],
+      },
+    ).parseAsync();
+
+    expect(collectAndPersistReports).toHaveBeenCalledWith(
+      expect.objectContaining({
+        verbose: true,
+        config: '/test/code-pushup.config.ts',
+        persist: expect.objectContaining({
+          format: ['md'],
```
Tlacenka, line 79:
```suggestion
```
This test is unnecessary. It tests the same properties and a subset of functions tested in the test above.

packages/cli/src/lib/implementation/config-middleware.integration.test.ts
```diff
@@ -25,6 +25,7 @@ describe('configMiddleware', () => {
         config: join(configDirPath, `code-pushup.config.${extension}`),
       });
       expect(config.config).toContain(`code-pushup.config.${extension}`);
+      expect(config.persist.format).toContain(`json`);
```
Tlacenka, line 28: This does not seem related to the test description. If you mean to test that `json` format is always included in the `persist.format`, no matter what configuration is parsed, it deserves a separate test case.

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -16,31 +17,34 @@ vi.mock('@code-pushup/core', async () => {
...
+      config: '/test/code-pushup.config.ts',
+      'persist.outputDir': '/test',
+      ...options,
+    }),
+    {
+      ...DEFAULT_CLI_CONFIGURATION,
+      commands: [yargsAutorunCommandObject()],
+    },
+  );
+
```
Tlacenka, line 34: This was removed on purpose for better readability of the tests. By having explicit values present, it is clear which arguments are passed to the test without having to scroll up and look at additional function definitions.
```suggestion
```

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -74,4 +78,18 @@ describe('autorun-command', () => {
...
+    expect(collectAndPersistReports).toHaveBeenCalledWith(
+      expect.objectContaining({
+        verbose: true,
+        config: '/test/code-pushup.config.ts',
+        persist: expect.objectContaining({
+          format: ['md', 'json'],
+        }),
+      }),
+    );
+  });
```
Tlacenka, line 94:
```suggestion
```

This logic is handled by the config-middleware, if I understand it correctly. So once a separate test case is added there for the addition of the `json` format (as commented above), it is unnecessary to create duplicate tests for this case elsewhere.

packages/core/src/lib/implementation/persist.unit.test.ts
```diff
@@ -41,31 +41,6 @@ describe('persistReport', () => {
     );
   });
 
-  it('should create a report in json format by default', async () => {
```
Tlacenka, line 44: Why has this been removed? It describes a valid use case from the core perspective - for each format the cases are tested here.

Tlacenka (dismissed):
I do not understand the changes being made to the tests. They often reintroduce the bad practices again or test logic in the wrong place (cli and core are after test refactor). I provided context in the comments below.

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -16,31 +17,34 @@ vi.mock('@code-pushup/core', async () => {
...
+      commands: [yargsAutorunCommandObject()],
+    },
+  );
+
 describe('autorun-command', () => {
   beforeEach(() => {
     vol.fromJSON(
       {
-        'my-report.json': JSON.stringify(MINIMAL_REPORT_MOCK),
+        'report.json': JSON.stringify(MINIMAL_REPORT_MOCK),
```
Tlacenka, line 39: Why is this file needed?

Tlacenka, line 39:
```suggestion
        'my-report.json': JSON.stringify(MINIMAL_REPORT_MOCK),
```
This change was here on purpose to check that the report file name is changed from the default if provided by the CLI.

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -52,7 +56,7 @@ describe('autorun-command', () => {
         verbose: true,
         config: '/test/code-pushup.config.ts',
         persist: expect.objectContaining({
-          filename: 'my-report',
+          filename: 'report',
```
Tlacenka, line 59:
```suggestion
          filename: 'my-report',
```
See above.

packages/cli/src/lib/implementation/config-middleware.integration.test.ts
BioPhoton: introduced options tests in #358 will reuse it here

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
BioPhoton: they are always the same

packages/core/src/lib/implementation/persist.unit.test.ts
BioPhoton: the logic got changed and now this is not true

packages/models/src/lib/report.ts
```diff
@@ -39,6 +39,7 @@ export const reportSchema = packageVersionSchema({
   .merge(
     z.object(
       {
+        commit: z.string().optional(),
```
BioPhoton, line 42: I added the `commit` to the model. Seem's we missed it so far as it is used only in the `md` report and `stdout` directly

testing-utils/src/lib/fixtures/configs/core-config.mock.ts
```diff
@@ -1,7 +1,5 @@
-import { CoreConfig } from '../../../../../packages/models/src';
-
 export const CORE_CONFIG_MOCK = {
-  persist: { outputDir: 'test', filename: 'output.json' },
+  persist: { outputDir: 'test' },
```
BioPhoton, line 2: must be a mistake as:
- should be `report`. `output` is used for runner output.
- should be by default `report´

packages/cli/src/lib/upload/upload-command.unit.test.ts
```diff
@@ -15,7 +15,7 @@ describe('upload-command-object', () => {
     vol.fromJSON(
       {
         'my-report.json': JSON.stringify(MINIMAL_REPORT_MOCK),
-        'code-pushup.config.ts': '', // only needs to exist for stat inside readCodePushupConfig
+        'code-pushup.config.ts': '{}', // only needs to exist for stat inside readCodePushupConfig
```
BioPhoton, line 18:
```suggestion
        'code-pushup.config.ts': '', // only needs to exist for stat inside readCodePushupConfig
```

packages/models/src/lib/report.ts
BioPhoton, line 42:
```suggestion
```

packages/core/src/lib/implementation/persist.ts
```diff
@@ -29,27 +29,34 @@ export async function persistReport(
 ): Promise<MultipleFileResults> {
   const { outputDir, filename, format } = options;
 
-  let scoredReport = scoreReport(report);
-  console.info(reportToStdout(scoredReport));
+  let scoredReport;
```
MishaSeredenkoPushBased, line 32: I just wonder about this place:
We have the task #352, which is about the report sorting on the earlier stage.
Shouldn't we keep assigning of `scoredReport` here, and sort it on the next string (in the scope of another task ofc)?
This way the sorted `scoredReport` could be passed to each necessary format.

BioPhoton: I assign it only if needed. Should be below in the code. And should be shared. No?

MishaSeredenkoPushBased (approved):

e2e/cli-e2e/tests/collect.e2e.test.ts
```diff
@@ -78,8 +78,7 @@ describe('CLI collect', () => {
     expect(stderr).toBe('');
 
     expect(stdout).toContain('Code PushUp Report');
-    expect(stdout).toContain('Generated reports');
-    expect(stdout).toContain('report.json');
+    expect(stdout).not.toContain('Generated reports');
```
matejchalk, line 81: This test scenario runs with `--persist.format=stdout`, that should be removed (no longer a valid option). And then I believe the assertions should stay the same, because JSON should be generated by default.

packages/cli/src/lib/autorun/autorun-command.ts
```diff
@@ -23,7 +23,19 @@ export function yargsAutorunCommandObject() {
...
-      await collectAndPersistReports(options);
+
+      // we need to ensure `json` is part of the formats as we want to upload
+      const optionsWithFormat: AutorunOptions = {
+        ...options,
+        persist: {
+          ...options.persist,
+          format: [
+            ...new Set([...options.persist.format, 'json']),
+          ] as AutorunOptions['persist']['format'],
```
matejchalk, line 34:
```suggestion
          format: [
            ...new Set([...options.persist.format, 'json'] satisfies Format[]),
          ],
```

or:

```suggestion
          format: [...new Set([...options.persist.format, 'json' as const])],
```

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
matejchalk: It's not a huge issue, but a recurring problem, so maybe it helps to explain the motivation better.

In general, it's better when a test contains all the context you need to know regarding what inputs are provided. Extracting setup code goes against this principle. There's a balance to be struck, of course, e.g. when the setup is long and often repeated in different tests, then extracting it may be the better option.

The old version of this test is more understandable for me, because within the `it`-block I see everything I need to understand it - e.g. "Why do the assertions expect `/test/code-pushup.config.ts`? Ah yes, because that's what argument was passed in, I see that a few lines above.". But with the new version, the relevant inputs and outputs aren't colocated, so I have to scroll up to understand the whole picture.

In fact, the new `cli` function is only used in this one test, so those no advantage with regards to reducing duplication. And I also don't understand why `persist.filename` is specified here, but e.g. `persist.outputDir` or `config` aren't, they seem to be just as relevant to the assertions.

PR #361, opened 2023-12-12
feat(cli): add autoloading of the config file

### LINKED ISSUE #307: When no `config` parameter is given, check all supported formats for a default value
As a user, I want to try out the CLI without providing any arguments. When I do so, I expect the CLI to use the default configuration, which may be in TS, JS or MJS (supported formats).

Currently, `code-pushup.config.js` is a hardcoded default value, so having `code-pushup.config.ts` will fail.

also we should allow the case fo no config present and do everything over the CLI arguments

@matejchalk will add the library.

BioPhoton:
Included in this PR is:
- split onlyPlugins middleware out of core config 
- adjust execution of core config
- simplify core config
- adopt existing docs

closes #307 

> [!NOTICE]
> Should wait until #358 and #363 is revolved

**Update:**
I moved in the second option as it cleans up the codebase without hacks. 
Will open another PR for the config refactoring.

### Problem
ATM `yargs` is not configured correctly:
- NO `config`API is used 
- `options` are without `default` values
- default handling is done in `middleware`

This setup is hard to setup, reason about maintain and extend as unintuitive and fragile.

### Suggested solution

We suggest to use the `config` API of `yargs` and implement the default values correctly.

This would be like the following:
- `config`API is used and filled with potential existing `code-pushup.config.(ts|mjs|js)` 
- `options` contain `default` values
- `middleware` gets removed completely

There are 2 possible ways to get close to the above solution:

#### Load RC config in a separate step before the CLI executes the command

As a proper solution is outside of this PR I suggest we discuss the following change:

1. Introduce the following function to the codebase of cli package
```ts
import {Argv, Options} from "yargs";
import {readCodePushupConfig} from "@code-pushup/core";

export async function withAsyncConfig<T>(unconfiguredCli: Argv<T>, optionsName = 'config', options: Options = {}) {
  // create a yargs CLI only configured to parse the config file
  const args = await unconfiguredCli.options({[optionsName]: options}).parseAsync();
  const configJson = await readCodePushupConfig(args[optionsName] as string);
  // configure cli with config
  return unconfiguredCli.config(configJson);
}
```

2. Make `cli` and `yargsCli` async and use the configure `yargs``

```ts
export async function yargsCli<T = unknown>(
  argv: string[],
  cfg: {
    // ...
    options?: { [key: string]: Options } & { config: Options };
  },
): Promise<Argv<T>> {
  // ...
  const cli = yargs(argv);

  const configuredCli = await withAsyncConfig(cli, 'config', options['config'])
  
  // ...
}
```

#### Execute the middleware before argument validation

A middleware can be added with the option `{applyBeforeValidation: boolean}`. 
We can use it and at lease reduce the hack with default values a bit. (config middle ware solves multiple problems ATM)

Still not as clean as the above when it comes to reasoning etc...


packages/core/src/lib/implementation/read-rc-file.ts
```diff
@@ -0,0 +1,51 @@
...
+    coreConfigSchema.parse,
+  );
+}
+
+export async function readRc(): Promise<CoreConfig> {
+  let ext = '';
+  // eslint-disable-next-line functional/no-loop-statements
+  for (const extension of ['ts', 'mjs', 'js']) {
+    const path = `${CONFIG_FILE_NAME}.${extension}`;
+    const exists = await fileExists(path).catch(() => false);
```
matejchalk, line 37: The `fileExists` utility already has this `catch`.

```suggestion
    const exists = await fileExists(path);
```

packages/cli/src/lib/yargs-cli.integration.test.ts
```diff
@@ -16,11 +16,11 @@ describe('yargsCli', () => {
     expect(parsedArgv.progress).toBe(true);
   });
 
-  it('should provide a correct default value for the config', async () => {
+  it('should provide no default value for the config', async () => {
```
Tlacenka, line 19:
```suggestion
```
If the logic has been moved to config-middleware, the test case here is misleading - the config does have a default prefix value but after reading this test case, it seems that one always has to provide `--config` argument. Please add a test to `config-middleware` instead.

packages/core/src/lib/implementation/read-rc-file.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (exists) {
+      ext = extension;
+      break;
+    }
+  }
+
+  return readRcByPath(
+    join(
+      process.cwd(),
+      ext ? `${CONFIG_FILE_NAME}.${ext}` : `${CONFIG_FILE_NAME}.ts`,
```
matejchalk, line 48: We already know that `ts` extension wasn't found, so I'd throw a readable error instead.

packages/core/src/lib/implementation/read-rc-file.integration.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      readRcByPath(join(configDirPath, 'code-pushup.empty.config.js')),
+    ).rejects.toThrow(`"code": "invalid_type",`);
+  });
+
+  it('should throw if the configuration is invalid', async () => {
+    await expect(
+      readRcByPath(join(configDirPath, 'code-pushup.invalid.config.ts')),
+    ).rejects.toThrow(/refs are duplicates/);
+  });
+});
```
matejchalk, line 32: It's weird not to have a happy path test case.

packages/core/src/lib/implementation/read-rc-file.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  return importEsmModule<CoreConfig>(
+    {
+      filepath,
+    },
+    coreConfigSchema.parse,
+  );
+}
+
+export async function readRc(): Promise<CoreConfig> {
```
Tlacenka, line 32: The name of the function is misleading. A better name would be `searchForRcConfig` or similar, that indicates that the function is not simply reading a file but it is scanning the root folder to find one of three expected files.

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  it('should throw if the file does not exist', async () => {
+    await expect(
+      readRcByPath(join('non-existent', 'config.file.js')),
+    ).rejects.toThrow(/Provided path .* is not valid./);
+  });
+});
+
+describe('readRc', () => {
+  it('should load ts', async () => {
```
Tlacenka, line 47:
```suggestion
  it('should prioritise a .ts configuration file', async () => {
```

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      },
+      MEMFS_VOLUME,
+    );
+
+    await expect(readRc()).resolves.toEqual(
+      expect.objectContaining({ upload: expect.any(Object) }),
+    );
+  });
+
+  it('should load mjs if no ts exists', async () => {
```
Tlacenka, line 62:
```suggestion
  it('should prioritise .mjs configuration file over .js', async () => {
```

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
```diff
@@ -0,0 +1,92 @@
...
+import { describe, expect } from 'vitest';
+import { CONFIG_FILE_NAME } from '@code-pushup/models';
+import { MEMFS_VOLUME } from '@code-pushup/testing-utils';
+import { readRc, readRcByPath } from './read-rc-file';
+
+describe('readRcByPath', () => {
+  it('should load a valid configuration file', async () => {
+    vol.fromJSON(
+      {
+        [`${CONFIG_FILE_NAME}.ts`]: '',
```
matejchalk, line 12: It's confusing that an empty file is being used as a valid config example. Maybe a comment that the content is provided by `bundle-require` mock would help.

packages/core/src/lib/implementation/read-rc-file.ts
```diff
@@ -0,0 +1,51 @@
...
+import {
+  CONFIG_FILE_NAME,
+  CoreConfig,
+  coreConfigSchema,
+} from '@code-pushup/models';
+import { fileExists, importEsmModule } from '@code-pushup/utils';
+
+export class ConfigPathError extends Error {
+  constructor(configPath: string) {
+    super(`Provided path '${configPath}' is not valid.`);
```
Tlacenka, line 11: This error is not applicable to all use cases.

If a use does not enter any configuration, the searching function is called. That internally calls the `readRcByPath` function which does not find anything. But it is not because a provided path is not valid but because the expected default configuration file does not exist. So in this case the user should see something like "Could not find any valid code-pushup.config.(ts|mjs|js) file." or similar.

packages/core/src/lib/implementation/read-code-pushup-config.integration.test.ts
```diff
@@ -1,64 +0,0 @@
-import { dirname, join } from 'path';
-import { fileURLToPath } from 'url';
-import { describe, expect } from 'vitest';
-import { readCodePushupConfig } from './read-code-pushup-config';
-
-describe('readCodePushupConfig', () => {
```
Tlacenka, line 6: By replacing the integration tests with unit tests, we are losing test coverage when using actual file system. While it is okay to test most cases using mocked file system, at least a happy path should stay without mocks.

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      },
+      MEMFS_VOLUME,
+    );
+
+    await expect(readRc()).resolves.toEqual(
+      expect.objectContaining({ upload: expect.any(Object) }),
+    );
+  });
+
+  it('should load js if no ts or mjs exists', async () => {
```
Tlacenka, line 76:
```suggestion
  it('should load a .js configuration file if no other valid extension exists', async () => {
```

Tlacenka (changes requested):
It is a great improvement overall, having the configuration extension not hardcoded.

The reason I put "request for changes" is the reduced coverage and missing documentation of the new logic.

packages/cli/src/lib/implementation/config-middleware.ts
```diff
@@ -14,7 +14,9 @@ export async function configMiddleware<
   const { config, ...cliOptions } = args as GeneralCliOptions &
     Required<CoreConfig> &
     OnlyPluginsOptions;
-  const importedRc = await readCodePushupConfig(config);
+
+  // if config path is given use it otherwise auto-load
+  const importedRc = config ? await readRcByPath(config) : await readRc();
```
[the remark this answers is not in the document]
BioPhoton: done

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
```diff
@@ -16,8 +16,8 @@ Options:
                                                        [boolean] [default: true]
       --verbose              When true creates more verbose output. This is help
                              ful when debugging.      [boolean] [default: false]
-      --config               Path the the config file, e.g. code-pushup.config.j
-                             s       [string] [default: \\"code-pushup.config.js\\"]
```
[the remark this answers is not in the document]
BioPhoton: done

packages/cli/src/lib/yargs-cli.integration.test.ts
BioPhoton: done

packages/core/src/lib/implementation/read-rc-file.ts
```diff
@@ -0,0 +1,51 @@
...
+      filepath,
+    },
+    coreConfigSchema.parse,
+  );
+}
+
+export async function readRc(): Promise<CoreConfig> {
+  let ext = '';
+  // eslint-disable-next-line functional/no-loop-statements
+  for (const extension of ['ts', 'mjs', 'js']) {
```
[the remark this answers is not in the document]
BioPhoton: done

packages/core/src/lib/implementation/read-rc-file.integration.test.ts
BioPhoton: yes reason was mocking is hard atm as it is always everywhere... and in this case filesystem as well as bundle require is used

BioPhoton: it is in now

packages/core/src/lib/implementation/read-rc-file.ts
BioPhoton: named it `autoloadRc`

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
BioPhoton: done

packages/core/src/lib/implementation/read-rc-file.ts
BioPhoton: added different error

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
BioPhoton: added comments

packages/cli/src/lib/implementation/config-middleware.unit.test.ts
```diff
@@ -0,0 +1,23 @@
...
+});
+
+describe('configMiddleware', () => {
+  it('should load code-pushup.config.(ts|mjs|js)', async () => {
+    // It is indirectly testing if code-pushup.config.(ts|mjs|ts) is used.
+    // The reason for this is a hard to setup test
+    await expect(configMiddleware({})).rejects.toThrow(
+      'No file code-pushup.config',
+    );
+  });
```
matejchalk, line 22: I think you'd have to create one these files with `memfs`'s `vol.fromJSON`. The reason it's complicated is because there's an (unneccessary, I believe) check with `fs.stat` before `bundle-require` is called (which checks the file exists then anyway), which means both have to be simulated in tests.

But since we already have integration tests set up for the config middleware, why not just add this test case there instead? Something like:

```ts
  it('should load code-pushup.config.(ts|mjs|js) by default', async () => {
    const config = await configMiddleware({});
    expect(config.config).toContain('code-pushup.config');
  });
```

BioPhoton: It would require to also mock fs in the integration tests etc ... But it is added now and checks against the root config

packages/cli/README.md
```diff
@@ -156,6 +156,9 @@ Each example is fully tested to give demonstrate best practices for plugin testi
 | **`--verbose`**  | `boolean` | `false`                 | When true creates more verbose output. This is helpful when debugging. |
 | **`--config`**   | `string`  | `code-pushup.config.js` | Path to the config file, e.g. code-pushup.config.js                    |
 
+> [!NOTE]  
+> By default, the CLI loads `code-pushup.config.(js|ts|mjs)` if no config path is provided over `--config`.
```
BioPhoton, line 160:
```suggestion
> By default, the CLI loads `code-pushup.config.(ts|mjs|js)` if no config path is provided over `--config`.
```

packages/cli/src/lib/implementation/global-options.ts
```diff
@@ -18,9 +18,9 @@ export function yargsGlobalOptionsDefinition(): Record<
       default: false,
     },
     config: {
-      describe: 'Path the the config file, e.g. code-pushup.config.js',
+      describe:
+        'Path the the config file, e.g. code-pushup.config.ts. By default it loads code-pushup.config.(js|ts|mjs).',
```
BioPhoton, line 22:
```suggestion
        'Path the the config file, e.g. code-pushup.config.ts. By default it loads code-pushup.config.(ts|mjs|js).',
```

packages/cli/README.md
```diff
@@ -18,7 +18,8 @@ The Code PushUp CLI serves to **collect audit results**, and optionally **upload
 
 It can be used locally in your repository, or integrated in your CI environment.
 
-_If you're looking for programmatic usage, then refer to the underlying [@code-pushup/core](../core/README.md) package instead._
+_If you're looking for programmatic usage, then refer to the underlying [@code-pushup/core](../core/README.md) package
+instead._
```
BioPhoton, line 22:
```suggestion
_If you're looking for programmatic usage, then refer to the underlying [@code-pushup/core](../core/README.md) package instead._
```

packages/cli/src/lib/implementation/utils.ts
```diff
@@ -23,3 +24,20 @@ export function coerceArray<T extends string>(param: T | T[] = []): T[] {
...
+  // create a yargs CLI only configured to parse the config file
+  const args = await unconfiguredCli
+    .options({ [optionsName]: options })
+    .parseAsync();
+  const configJson = transform
+    ? ((await transform(args[optionsName] as string)) as object)
+    : {};
+  // configure cli with config
+  return unconfiguredCli.config(configJson);
+}
```
matejchalk, line 43: The more I see this, the less I like it :/ It feels very hacky and unintuitive, the previous solution where it was all the config middleware's responsibility.

matejchalk (changes requested):
Honestly, I'm getting lost in this PR, I don't understand the more recent changes. 

Could you please create just for auto-loading the config file path? :pray: The `withAsyncConfig` refactor bloats the scope, and I've yet to see the benefit.

packages/cli/src/lib/implementation/utils.ts
BioPhoton, line 43:
```suggestion
```

packages/cli/src/lib/implementation/utils.ts
```diff
@@ -1,3 +1,4 @@
+import { Argv, Options } from 'yargs';
```
BioPhoton, line 1:
```suggestion
```

packages/cli/src/lib/implementation/core-config.middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+      config,
+      ...importedRc,
+      ...cliOptions,
+      upload: {
+        ...importedRc.upload,
+        ...cliOptions.upload,
+      },
+      persist: {
+        ...cliOptions?.persist,
+        ...importedRc?.persist,
```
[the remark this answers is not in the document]
BioPhoton: ouch! So the test is missing too, or wrong :D I will fix it asap.

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
```diff
@@ -1,28 +1,49 @@
...
-  'testing-utils',
-  'src',
-  'lib',
-  'fixtures',
-  'configs',
-);
+vi.mock('@code-pushup/core', async () => {
+  const core = await vi.importActual('@code-pushup/core');
+  return {
+    ...(core as object),
```
matejchalk, line 10: You can use `typeof import('...')` with `vi.importActual('...')` to get the accurate types instead of `uknown`:

```suggestion
  const core: typeof import('@code-pushup/core') = await vi.importActual('@code-pushup/core');
  return {
    ...core,
```

packages/utils/src/lib/file-system.ts
```diff
@@ -84,7 +86,7 @@ export async function importEsmModule(options: Options): Promise<unknown> {
   if (!('default' in mod)) {
     throw new NoExportError(options.filepath);
   }
-  return mod.default;
+  return mod.default as T;
 }
```
matejchalk, line 90: Please revert these changes. The original is safer, while this version makes unsafe assumptions.

Relates to how it's being used in `readRcFile`:
```ts
// `cfg: CoreConfig` ISN'T TRUE, we haven't validated it yet, it should be `unknown`
const cfg = await importEsmModule<CoreConfig>({ filepath });

return coreConfigSchema.parse(cfg); // NOW it's a CoreConfig (otherwise it'd throw) - Zod infers it for us
```

The guiding principle in TypeScript should always be to **tell the truth**.

- `as T` - probably a lie
- `as uknown as T` - gotta be a lie
- `@ts-ignore`, `any` - what is truth anymore? lies galore

matejchalk (changes requested):

packages/cli/README.md
```diff
@@ -154,20 +150,23 @@ Each example is fully tested to demonstrate best practices for plugin testing as
...
-| **`--upload.server`**       | `string`             | n/a      | URL to your portal server.                                                  |
-| **`--upload.apiKey`**       | `string`             | n/a      | API key for the portal server.                                              |
-| **`--onlyPlugins`**         | `string[]`           | `[]`     | Only run the specified plugins. Applicable to all commands except `upload`. |
+| Option                      | Type              | Default        | Description                                                                 |
+| --------------------------- | ----------------- | -------------- | --------------------------------------------------------------------------- |
+| **`--persist.outputDir`**   | `string`          | `.code-pushup` | Directory for the produced reports.                                         |
+| **`--persist.filename`**    | `string`          | `report`       | Filename for the produced reports without extension.                        |
+| **`--persist.format`**      | `('json','md')[]` | `json`         | Format(s) of the report file.                                               |
```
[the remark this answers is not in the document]
BioPhoton: done

packages/cli/src/lib/implementation/core-config.model.ts
```diff
@@ -7,13 +7,13 @@ export type PersistConfigCliOptions = {
...
 };
 
+/* eslint-disable @typescript-eslint/naming-convention */
 export type UploadConfigCliOptions = {
   'upload.organization': string;
   'upload.project': string;
   'upload.apiKey': string;
   'upload.server': string;
 };
-/* eslint-enable @typescript-eslint/naming-convention */
```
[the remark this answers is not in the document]
BioPhoton: 💩

packages/utils/src/lib/file-system.ts
BioPhoton: Special thx for the guideline! Where did you have that info from?

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    await expect(autoloadRc()).resolves.toEqual(
+      expect.objectContaining({ upload: expect.any(Object) }),
+    );
+  });
+
+  it('should load a .js configuration file if no other valid extension exists', async () => {
+    vol.fromJSON(
+      {
+        // this is just here to satisfy the file system check. the file ise served over a mock in bundleRequire
```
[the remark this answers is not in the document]
BioPhoton: its on all lines :D

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        [`${CONFIG_FILE_NAME}.js`]: '',
+        [`${CONFIG_FILE_NAME}.mjs`]: '',
+        [`${CONFIG_FILE_NAME}.ts`]: '',
+      },
+      MEMFS_VOLUME,
+    );
+
+    await expect(autoloadRc()).resolves.toEqual(
+      expect.objectContaining({ upload: expect.any(Object) }),
+    );
```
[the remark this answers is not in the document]
BioPhoton: its due to bundle require mock. I cant test specific things really...

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
matejchalk: There's a problem with the Nx lint rule: https://github.com/code-pushup/cli/actions/runs/7658472674/job/20873664503?pr=361

We encountered it in #431 as well. So in the end using `const core: object` is probably the best possible way.

packages/utils/src/lib/file-system.ts
matejchalk: Experience mostly :laughing: But if you want a tip on who to follow regarding TypeScript best practices, then Matt Pocock is a great source. The "types should tell the truth" line specifically I stole from Theo Browne, I think :grin:

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
BioPhoton: reverted to object

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: fixed

testing-utils/src/lib/setup/bundle-require.mock.ts
matejchalk, line 1: Is this merge error? @Tlacenka deleted it in #431

matejchalk (approved):
Great job :+1: The tests look really good in the end :slightly_smiling_face: Really looking forward to using this feature and deleting `--config` arguments everywhere :grin: :tada:

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        [`${CONFIG_FILE_NAME}.mjs`]: '',
+        [`${CONFIG_FILE_NAME}.ts`]: '',
+      },
+      MEMFS_VOLUME,
+    );
+
+    await expect(autoloadRc()).resolves.toEqual(
+      expect.objectContaining({
+        upload: expect.objectContaining({
+          project: 'ts',
```
Tlacenka, line 48: I think it would be better to move the comments here to explain where the value is coming from (also care, there is a 'ise' typo in all three comments right now).

packages/cli/src/lib/implementation/core-config.middleware.unit.test.ts
```diff
@@ -0,0 +1,58 @@
...
+        upload: {
+          ...CORE_CONFIG_MOCK.upload,
+          project: `cli-autoload`,
+        },
+      }),
+    ),
+  };
+});
+
+describe('coreConfigMiddleware', () => {
```
Tlacenka, line 36: The file content fetching and validation should be tested in `readRcByPath` which is either called directly or via `autoLoad` function. If this is to be a unit test, simply checking whether `autoLoad` is called or not will suffice.

packages/cli/src/lib/implementation/core-config.middleware.unit.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  it('should load code-pushup.config.(ts|mjs|js) by default', async () => {
+    const config = await coreConfigMiddleware({});
+    expect(config?.upload?.project).toBe('cli-autoload');
+  });
+
+  it.each(['ts', 'mjs', 'js'])(
```
Tlacenka, line 42: This original set of functions should go back to the integration test of `readRcByPath` (which currently only tests `.js` for testing happy paths for all supported files. Having it as a unit test makes us lose coverage and certainty that all types are correctly loaded. All config files are already prepared in `testing-utils` for this.

packages/cli/src/lib/implementation/core-config.middleware.unit.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    async extension => {
+      const config = await coreConfigMiddleware({
+        config: `code-pushup.config.${extension}`,
+      });
+      expect(config.config).toContain(`code-pushup.config.${extension}`);
+      expect(config.upload?.project).toContain(extension);
+    },
+  );
+
+  it('should throw with invalid config path', async () => {
```
Tlacenka, line 53: Since this is a unit test, fs is mocked and therefore checking invalid paths does not function normally here (unless I am mistaken). So such a test should stay as an integration test as well.

packages/core/src/lib/implementation/read-rc-file.integration.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    '..',
+    '..',
+    'testing-utils',
+    'src',
+    'lib',
+    'fixtures',
+    'configs',
+  );
+
+  it('should load the configuration', async () => {
```
Tlacenka, line 21: As a result of the comment above, this test should be replaced by the original trio of tests, one happy path per type.

Tlacenka (changes requested):
Overall a nice improvement!

Main feedback - the lost integration test coverage should be restored before merging.

testing-utils/src/lib/setup/bundle-require.mock.ts
Tlacenka: The bundle require mock should indeed stay removed from `testing-utils`.

BioPhoton: I kept it in to have a living example.

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
BioPhoton: added a comment

packages/cli/src/lib/implementation/core-config.middleware.unit.test.ts
BioPhoton: the mocked API throws if the file does not exist

packages/core/src/lib/implementation/read-rc-file.integration.test.ts
BioPhoton: as the mocked API also throws it is valid imo

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
BioPhoton: is this a blocker for this PR

matejchalk (dismissed):

packages/cli/src/lib/implementation/core-config.middleware.unit.test.ts
matejchalk: All that is needed is to un-delete `core-config.middleware.integration.test.ts`, right?

BioPhoton: undeleted

packages/cli/src/lib/implementation/core-config.middleware.unit.test.ts
BioPhoton: What can I do as a next step?

Tlacenka (approved):
Test coverage restored.

Let us please not forget about the rest of the comments in a follow-up PR.

PR #362, opened 2023-12-12
fix(cli): set terminal output to full width

### LINKED ISSUE #123: CLI help command descriptions should fit in one line
```sh
Commands:
  code-pushup               Shortcut for running collect followed by upload
                                                                       [default]

Options:
      --version              Show version number                       [boolean]
      --config               Path the the config file, e.g. code-pushup.config.j
                             s       [string] [default: \\"code-pushup.config.js\\"]
      --persist.outputDir    Directory for the produced reports         [string]
      --persist.format       Format of the report output. e.g. \`md\`, \`json\`, \`st
                             dout\`                                       [array]
      --upload.organization  Organization slug from portal              [string]
      --upload.project       Project slug from portal                   [string]
      --upload.server        URL to your portal server                  [string]
      --upload.apiKey        API key for the portal server              [string]
  -h, --help                 Show help                                 [boolean]
```

Also add `--help` command to debug section

BioPhoton:
Implements:

- [x] default screen with across the project
- [x] terminal output fits

Screenshot 2024-01-17 at 22 17 59

closes #123


packages/utils/src/lib/reports/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const TERMINAL_WIDTH = 80;
+export const TERMINAL_WIDTH = 125;
```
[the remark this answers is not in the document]
BioPhoton: added a comment and set it to 80

examples/plugins/project.json
```diff
@@ -53,6 +53,14 @@
         "^build",
         { "projects": ["cli"], "target": "build" }
       ]
+    },
+    "run-help": {
+      "command": "npx dist/packages/cli help --config=examples/plugins/code-pushup.config.ts",
```
Tlacenka, line 58: Why is there a `--config` parameter added to `help` command?
```suggestion
      "command": "npx dist/packages/cli help",
```

Tlacenka (commented):
Overall a good improvement. 🍺 

I sent you the solution for the failing E2E test due to different line wrap for longer audit names - providing a shorter string to match the `exampleAuditTile` value should solve that.

packages/utils/src/lib/reports/generate-stdout-summary.ts
```diff
@@ -52,7 +52,6 @@ function reportToDetailSection(report: ScoredReport): string {
         },
         {
           text: chalk.cyanBright(audit.displayValue || `${audit.value}`),
-          width: 10,
```
matejchalk, line 55: Keep `width: 10`, please. Without it, the middle column doesn't fill available space and a lot more of the audit titles don't fit. Look at difference in the snapshot - the original takes much better advantage of the available width.

image

packages/utils/src/lib/reports/generate-stdout-summary.ts
```diff
@@ -73,8 +72,11 @@ function reportToOverviewSection({
   plugins,
 }: ScoredReport): string {
   const table = new CliTable3({
+    /* eslint-disable no-magic-numbers */
+    colWidths: [38, 7, 8],
+    /* eslint-enable no-magic-numbers */
```
matejchalk, line 77: If the lint problem only spans one line, then `eslint-disable-next-line` is simpler:

```suggestion
    // eslint-disable-next-line no-magic-numbers
    colWidths: [38, 7, 8],
```

matejchalk (changes requested):

packages/utils/src/lib/reports/generate-stdout-summary.ts
BioPhoton: done

packages/utils/src/lib/reports/generate-stdout-summary.ts
```diff
@@ -73,8 +73,10 @@ function reportToOverviewSection({
   plugins,
 }: ScoredReport): string {
   const table = new CliTable3({
+    // eslint-disable-next-line no-magic-numbers
+    colWidths: [67, 7, 8],
     head: reportRawOverviewTableHeaders,
-    colAligns: ['left', 'right', 'right'],
+    colAligns: ['left', 'right', 'left'],
```
matejchalk, line 79: Numbers should be aligned right.

```suggestion
    colAligns: ['left', 'right', 'right'],
```

e2e/cli-e2e/tests/collect.e2e.test.ts
```diff
@@ -3,7 +3,7 @@ import { executeProcess, readJsonFile, readTextFile } from '@code-pushup/utils';
 
 describe('CLI collect', () => {
   const exampleCategoryTitle = 'Code style';
-  const exampleAuditTitle = 'Require `const` declarations for variables';
+  const exampleAuditTitle = 'Require `const`';
```
matejchalk, line 6: Why not keep the full title? :raised_eyebrow:

matejchalk (changes requested):

e2e/cli-e2e/tests/collect.e2e.test.ts
matejchalk: Oh, I see. But that should no longer be the case once you revert the width back to original, right?

BioPhoton: As it was quite tricky to understand the error, is there a better way? RegEx or so, that is more stable to line breaks

matejchalk: A regex allows you to use `\s` instead of a space. But that wouldn't handle the next column in between. A `.*` would cover it, but that's also quite likely to match something it shouldn't. 

I wouldn't complicate it. As long as we pick an audit title that fits on one line, then we're fine. This was is pretty short.

Tlacenka: When I was debugging it, `Disallow unused variables` worked for me.

packages/utils/src/lib/reports/generate-stdout-summary.ts
```diff
@@ -73,6 +73,8 @@ function reportToOverviewSection({
   plugins,
 }: ScoredReport): string {
   const table = new CliTable3({
+    // eslint-disable-next-line no-magic-numbers
+    colWidths: [67, 7, 8],
```
matejchalk, line 77: This doesn't fit in 80 characters. 67+7+8=82. Plus there are 4 column separators, so the table is stretched to 86 characters (I checked the snapshot as well).

image

If you want it to be exactly 80, then subtract 6:

```suggestion
    colWidths: [61, 7, 8],
```

If you want to calculate it relative to the `TERMINAL_WIDTH` constant, you can do something along these lines:

```suggestion
    colWidths: [TERMINAL_WIDTH - 7 - 8 - 4, 7, 8],
```

matejchalk (changes requested):

packages/utils/src/lib/reports/generate-stdout-summary.ts
BioPhoton: Thx! Now it is connected to the const! Way better!

matejchalk (approved):

PR #363, opened 2023-12-12
fix(cli): adjust persist.format options

### LINKED ISSUE #297: Comma separated formats does not seem to work e.g. `--persist.format=md,json`

BioPhoton:
closes #297


BioPhoton:
@IKatsuba I thing the reason is that i can't apply the corece over options object. I can try to read the docs for a better implementation but your [hint](https://github.com/code-pushup/cli/issues/297#issuecomment-1830519415) unfortunately did not work for me.

packages/cli/src/lib/implementation/only-plugins-utils.unit.test.ts
```diff
@@ -106,8 +108,9 @@ describe('validateOnlyPluginsOption', () => {
       [{ slug: 'plugin1' }, { slug: 'plugin2' }] as CoreConfig['plugins'],
       {
         onlyPlugins: ['plugin1'],
+        verbose: true,
       },
     );
-    expect(console.warn).not.toHaveBeenCalled();
+    expect(console.info).not.toHaveBeenCalled();
```
BioPhoton, line 114:
```suggestion
    expect(console.warn).not.toHaveBeenCalled();
```

e2e/cli-e2e/tests/print-config.e2e.test.ts
```diff
@@ -54,7 +54,70 @@ describe('print-config', () => {
...
+          ...objectToCliArgs(options),
+        ],
+      });
+
+      expect(JSON.parse(stdout)?.persist).toEqual(
+        expect.objectContaining(result),
+      );
+    },
+    120000,
+  );
```
matejchalk, line 118: E2E tests should be reserved for critical user flows. Testing many different scenarios of argument parsing that most users probably won't encounter isn't cost-effective for E2E tests. Integration (or even unit) tests in `cli` would be a much better fit.

Also, please use include the parameters in the test name when using `it.each` ([`%s`, `%j`, etc.](https://vitest.dev/api/#test-each)), otherwise it creates duplicate test names.

packages/cli/src/lib/implementation/only-plugins-utils.ts
```diff
@@ -42,21 +46,25 @@ export function filterCategoryByOnlyPluginsOption(
...
+      console.warn(
+        `${chalk.yellow(
+          '⚠',
+        )} The --onlyPlugin argument references plugins with "${missingPlugins.join(
+          '", "',
+        )}" slugs, but no such plugins are present in the configuration. Expected one of the following plugin slugs: "${plugins
+          .map(({ slug }) => slug)
+          .join('", "')}".`,
+      );
   }
```
matejchalk, line 69: Same as previous comment:

```ts
if (missingPlugins.length > 0 && verbose) {
  console.warn(...);
}
```

packages/cli/src/lib/implementation/only-plugins-utils.ts
```diff
@@ -26,13 +29,14 @@ export function filterCategoryByOnlyPluginsOption(
...
-        );
+        verbose &&
+          console.info(
+            `${chalk.yellow('⚠')} Category "${
+              category.title
+            }" is ignored because it references audits from skipped plugin "${
+              ref.slug
+            }"`,
+          );
       }
```
matejchalk, line 40: Please use `if`-statements for conditional execution, logical operators like `&&`, `||`, `!` or `? :` are intended for expressions (i.e. something that has returns a value).

In fact, you can merge the `verbose` flag into the `if`-statement above:

```ts
if (!isNotSkipped && verbose) {
  console.info(...);
}
```

packages/cli/src/lib/implementation/utils.ts
```diff
@@ -1,5 +1,7 @@
 // log error and flush stdout so that Yargs doesn't suppress it
 // related issue: https://github.com/yargs/yargs/issues/2118
+import { toArray } from '@code-pushup/utils';
+
 export function logErrorBeforeThrow<
```
matejchalk, line 5: The comment belongs to the `logErrorBeforeThrow` function, not the import.

```suggestion
import { toArray } from '@code-pushup/utils';

// log error and flush stdout so that Yargs doesn't suppress it
// related issue: https://github.com/yargs/yargs/issues/2118
export function logErrorBeforeThrow<
```

matejchalk (changes requested):

e2e/cli-e2e/tests/print-config.e2e.test.ts
BioPhoton: Thx will do both!

packages/cli/src/lib/implementation/utils.ts
BioPhoton: You are right

MishaSeredenkoPushBased (approved):
I argee with @matejchalk comments - and except it LGTM 👍

e2e/cli-e2e/tests/print-config.e2e.test.ts
BioPhoton: moved

packages/cli/src/lib/implementation/only-plugins-utils.ts
BioPhoton: done. That was a nice small improvement. Sad i missed it.

packages/cli/src/lib/implementation/only-plugins-utils.ts
BioPhoton: yes. way better!

Also thx for the additional info on logical operators and where to use them

packages/cli/src/lib/implementation/core-config-options.integration.test.ts
```diff
@@ -1,20 +1,47 @@
...
+    // persist.outputDir
+    [{ 'persist.outputDir': 'my-dir' }, { outputDir: 'my-dir' }],
+    // persist.filename
+    [{ 'persist.filename': 'my-report' }, { filename: 'my-report' }],
+    // persist.format
+    [{ 'persist.format': 'md' }, { format: ['md'] }],
+    [{ 'persist.format': ['md', 'json'] }, { format: ['md', 'json'] }],
+    // [{ 'persist.format': 'md,json' }, { format: ['md', 'json'] }], @TODO comment in when config auto-loading is implemented
+  ])(
+    'should parse persist options %s correctly as %j',
```
matejchalk, line 40: Since the first parameter is an object, I'd expect the JSON serialization to be more readable than the string serialization (`[object Object]`, I'm guessing).

```suggestion
    'should parse persist options %j correctly as %j',
```

packages/cli/src/lib/implementation/only-plugin-options.integration.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    }).parseAsync() as unknown as Promise<OnlyPluginsOptions>;
+  }
+
+  it('should fill defaults', async () => {
+    const args = await argsFromCli({});
+    expect(args).toBeDefined();
+    // "_" and "$0" are in by default
+    expect(Object.keys(args)).toHaveLength(4);
+  });
+  it.each([
```
matejchalk, line 22:
```suggestion
  });

  it.each([
```

packages/cli/src/lib/implementation/only-plugin-options.integration.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    // defaults
+    [{}, []],
+    [{ onlyPlugins: 'lighthouse' }, ['lighthouse']],
+    [{ onlyPlugins: ['lighthouse', 'eslint'] }, ['lighthouse', 'eslint']],
+    [{ onlyPlugins: 'lighthouse,eslint' }, ['lighthouse', 'eslint']],
+  ])(
+    'should parse onlyPlugins options %s correctly as %j correctly',
```
matejchalk, line 29:
```suggestion
    'should parse onlyPlugins options %j correctly as %j correctly',
```

packages/cli/src/lib/implementation/only-plugin-options.integration.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    return yargsCli<T>(objectToCliArgs(args), {
+      options: yargsOnlyPluginsOptionsDefinition(),
+    }).parseAsync() as unknown as Promise<OnlyPluginsOptions>;
+  }
+
+  it('should fill defaults', async () => {
+    const args = await argsFromCli({});
+    expect(args).toBeDefined();
+    // "_" and "$0" are in by default
+    expect(Object.keys(args)).toHaveLength(4);
```
matejchalk, line 20: Why 4? Based on the comment, I would expect 2.

e2e/cli-e2e/tests/print-config.e2e.test.ts
```diff
@@ -54,7 +54,7 @@ describe('print-config', () => {
     120000,
   );
 
-  it('should load .ts config file and overwrite it with CLI arguments', async ext => {
+  it('should load .ts config file with overloads arguments', async () => {
```
BioPhoton, line 57:
```suggestion
  it('should load .ts config file and overload it with arguments', async () => {
```

packages/cli/src/lib/implementation/core-config-options.integration.test.ts
BioPhoton: I did not know about %s is for string and %j for json :o 

Thx for the info! 

- %p - [pretty-format](https://www.npmjs.com/package/pretty-format).
- %s- String.| %d- Number. | %i - Integer.
- %f - Floating point value.| %j - JSON.| %o - Object.
- %# - Index of the test case.| %% - single percent sign ('%').

packages/cli/src/lib/implementation/only-plugin-options.integration.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  ) {
+    return yargsCli<T>(objectToCliArgs(args), {
+      options: yargsOnlyPluginsOptionsDefinition(),
+    }).parseAsync() as unknown as Promise<OnlyPluginsOptions>;
+  }
+
+  it('should fill defaults', async () => {
+    const args = await argsFromCli({});
+    expect(args).toBeDefined();
+    // "_" and "$0" are in by default
```
BioPhoton, line 19:
```suggestion
    // "_" and "$0" are in by default
    // camelCase and kebab-case of each option value
```

packages/cli/src/lib/implementation/only-plugin-options.integration.test.ts
BioPhoton: added comments

matejchalk (approved):

PR #365, opened 2023-12-13
ci: collect and upload code pushup report after release

### LINKED ISSUE #343: Run Code PushUp after release commit
Follow-up to https://github.com/code-pushup/portal/issues/159

- [ ] in [release.yml](https://github.com/code-pushup/cli/blob/main/.github/workflows/release.yml), run Code PushUp collect and upload after comitting

IKatsuba:
Close #343

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,3 +90,10 @@ jobs:
         run: |
           echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
           npx nx run-many -t=deploy --exclude=plugin-lighthouse,nx-plugin
+      - name: Collect and Upload Code PushUp report
+        run: npx dist/packages/cli --config code-pushup.config.ts autorun
+        env:
+          CP_SERVER: ${{ secrets.CP_SERVER }}
+          CP_API_KEY: ${{ secrets.CP_API_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -90,3 +90,10 @@ jobs:
         run: |
           echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
           npx nx run-many -t=deploy --exclude=plugin-lighthouse,nx-plugin
+      - name: Collect and Upload Code PushUp report
```
Tlacenka, line 93: In the [CI pipeline](https://github.com/code-pushup/cli/blob/main/.github/workflows/ci.yml#L188) where the reports are also gathered, there is an additional step for creating an artifact with the reports.
Is that not needed here? We should keep the workflows consistent.

IKatsuba: I can add it too. It is not a big problem, I think

matejchalk: Yeah, why not :slightly_smiling_face: It can be copy-pasted from the other workflow: https://github.com/code-pushup/cli/blob/main/.github/workflows/ci.yml#L188-L192

matejchalk (dismissed):

matejchalk (approved):

Issue #262, opened 2023-11-15
utils package test refactor


Tlacenka:
As a contributor to Code PushUp, I would like for the tests to be consistently structured and follow good practices.
There are several areas in which I identified room for improvement. In this issue, the following targeted changes should be made.

## Coverage
Split to #272 

## Files
- [x] Rename `transformation` to `transform` (keep it consistent that the `transform` files transform something).
- [x] The different report format functions are split across multiple files (`report.ts` is shared, `md` has both `report-to-md` and `md` folder, stdout` has `report-to-stdout`). Split the `MD` report, `stdout` report and possibly `HTML` formatting functions to different files/folders.
- [x] Consider moving `logMultipleFileResults` to `core`.
- [x] Remove unnecessary config files and console mock file. Move reusable files to `testing-utils`.

## Good practices
- [x] Transform `report-to-stdout`, `report-to-md`, `progress` into integration test files.
- [x] Define the colour ranges in one place and reuse the enum in all places where it is used. Currently there are hardcoded values in `withColor`, `getSquaredScoreMarker` and `getRoundScoreMarker`.
- [x] Replace `mockConsole` and related setup by `spyOn` for `console.log` in `logMultipleFileResults`.
- [x] Use `memfs` to test from where `bundleRequire` in `importEsmModule` takes the file. Currently, there are three one-line/empty configs which may be removed if this is done.
- [x] Split types from functions declared in `file-system` to different files.
- [x] Remove `objectToCliArgs` from tests - it reduces readability. Pass hardcoded array of arguments instead.
- [x] Name the functions based on what they do. Consider renaming `reportToStdout` to `generateStdoutReport` (otherwise implies it actually prints something) and `generateMdReport` for `md` to keep it consistent.
- [x] Use `1_234_567` formatting for long numbers, e.g. in `formatBytes` tests.
- [x] Rename `report()` and similar to reflect they are mocks.
- [x] Move the `loadReport` test near its `fs` mock setup, so that one doesn't have to scroll to see it.
- [x] Parse out colour formatting symbols in `logMultipleFileResults` tests for better readability.
- [x] In `executeProcess`, replace first `node -v` asserted value with a regexp (`v` is not very clear).
- [x] In `report-to-md`, the test uses a timestamp from the year 2000 yet the commit is 2021. Make the date difference realistic.

Test objects (`toBeContaining`) and within them, assert their properties instead of a long list of assertions. Applicable for:
- [x] `deepClone` (replace `toEqual` parts)
[...]

PR #377, opened 2023-12-21
refactor(utils): separate report sorting logic and use it once for all report outputs

### LINKED ISSUE #352: Implement report sorting for audits, groups, categories on earlier stage

MishaSeredenkoPushBased:
Now report will be sorted in the beginning of `persistReport()` function, before being passed into any output format.
This allows to remove the sorting logic from `reportToMd` and `reportToStdout`.

Closes #352


Tlacenka (changes requested):

packages/utils/src/lib/__snapshots__/sort-report.integration.test.ts.snap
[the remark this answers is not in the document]
BioPhoton: drop the snapshot and use string contains instead

packages/utils/src/lib/report-to-md.ts
```diff
@@ -95,38 +87,17 @@ function reportToCategoriesSection(report: ScoredReport): string {
...
-
-    const audits = auditsAndGroups.audits
-      .sort(sortCategoryAudits)
-      .map(audit => auditItemToCategorySection(audit, plugins))
-      .join(NEW_LINE);
-
-    const groups = auditsAndGroups.groups
-      .map(group => groupItemToCategorySection(group, plugins))
-      .join('');
+    const categoryItems = category.refs.reduce((acc, ref) => {
```
BioPhoton, line 90: The naming `categoryItems` does not tell me what is is... Is it scored refs? Is it categories?

packages/utils/src/lib/report-to-md.integration.test.ts
```diff
@@ -21,7 +22,7 @@ describe('report-to-md', () => {
       author: 'BioPhoton',
       date: 'Sat Sep 10 12:00:00 2021 +0200',
     };
-    const mdReport = reportToMd(scoreReport(report()), commit);
+    const mdReport = reportToMd(sortReport(scoreReport(report())), commit);
```
BioPhoton, line 25: We discussed it should also be stored sorted as json.

packages/utils/src/lib/sort-report.integration.test.ts
```diff
@@ -0,0 +1,10 @@
+import { report } from '@code-pushup/models/testing';
+import { scoreReport } from './scoring';
+import { sortReport } from './sort-report';
+
+describe('sortReport', () => {
+  it('should sort the report', () => {
```
BioPhoton, line 6: This description could give more insights on what sorting means.
Also a snapshot is most probably the worst testing API here.
A couple of more cases would also help to understand the test better.

BioPhoton (dismissed):
left comments

packages/utils/src/lib/report-to-md.ts
MishaSeredenkoPushBased: It's the markdown string, containing categories, which include their audits and groups of audits. I already have `categories` variable here, so I'll better rename it to `categoryMDItems`, I hope it would be clearer.

packages/utils/src/lib/__snapshots__/sort-report.integration.test.ts.snap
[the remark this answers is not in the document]
MishaSeredenkoPushBased: I'll keep the snapshot, as after the refactor of tests and changing the mocks by @Tlacenka at #262 the mock data is much more representable and compact.

packages/utils/src/lib/sort-report.ts
```diff
@@ -0,0 +1,77 @@
...
+        const bIndex = sortedAuditsAndGroups.findIndex(
+          ref => ref.slug === b.slug,
+        );
+        return aIndex - bIndex;
+      });
+
+      return [...acc, { ...category, refs: sortedRefs }];
+    },
+    [],
+  );
```
matejchalk, line 59: @MishaSeredenkoPushBased I notice you have a tendency to overuse `reduce`, when a `map`, `filter`, `flatMap`, `join` or some combination of them could be applied instead.

Take a step back and notice the pattern:

```ts
categories.reduce<ScoredCategoryConfig>(
  (acc, category) => {
    // ...
    return [...acc, { ...category, refs: sortedRefs }]
  },
  []
);
```

:top: The accumulator is an array, and each new accumulator is a copy of the previous array with an additional item at the end, which is derived from the current item. In other words, you've reinvented `Array.map`:

```ts
categories.map((category): ScoredCategoryConfig => {
  // ...
  return { ...category, refs: sortedRefs };
});
```

The `reduce` also works of course, but it's not the most appropriate method here. Using a method like `map` is a clearer statement of intent ("I'm aggregating something" :arrow_right: "I'm transforming each item"). While `reduce` is more flexible, it's also more cumbersome - more noise, explicit type annotations needed (because the accumulator could be anything), and nested `reduce`s are horrible to read (2 accumulator variables - either shadowed or have awkward names).

So I would discourage you from reaching for `reduce` as a goto solution whenever you need to loop over an array, it's more of a last resort if none of the other specialized methods fit. A good rule of thumb is if your accumulator is an array, then there's probably a better method. Some examples:
- `arr.reduce((acc, item) => [...acc, f(item)], [])` :arrow_right: `arr.map(f)`
- `arr.reduce((acc, item) => f(item) ? acc : [...acc, item], [])` :arrow_right: `arr.filter(f)`
- `arr.reduce((acc, item) => f1(item) ? acc : [...acc, f2(item)], [])` :arrow_right: `arr.filter(f1).map(f2)`
- `arr.reduce((acc, item) => [...acc, ...f(item)], [])` :arrow_right: `arr.flatMap(f)`

packages/utils/src/lib/__snapshots__/sort-report.integration.test.ts.snap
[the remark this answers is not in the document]
MishaSeredenkoPushBased: Reused `REPORT_MOCK` report mock, added detailed description for the test

matejchalk (dismissed):

packages/utils/src/lib/sort-report.integration.test.ts
MishaSeredenkoPushBased: The compare functions like `compareAudits`, `compareCategoryAudits` are already covered by unit tests. In this case I've kept the snapshot, but used the smaller mock, and left the description.

packages/utils/src/lib/sort-report.ts
MishaSeredenkoPushBased: Thanks for the great explanation! ❤️  I've used `map` instead of outer `reduce` in my case.

vmasek (dismissed):

packages/utils/src/lib/sort-report.integration.test.ts
```diff
@@ -0,0 +1,52 @@
+import { REPORT_MOCK } from '@code-pushup/testing-utils';
+import { scoreReport } from './scoring';
+import { sortReport } from './sort-report';
+
+describe('sortReport', () => {
```
Tlacenka, line 5: I am missing information of the sorting priorities. E.g. `should sort the audits and groups based on X, Y then Z and categories based on A then B`

And then for each section I would reiterate what the person should look for in the results, e.g.
`Categories (sorted based on X then Y):`

Tlacenka (approved):
Nice! 🍻

matejchalk (dismissed):

vmasek (dismissed):

Tlacenka (approved):

matejchalk (approved):

vmasek (approved):

PR #381, opened 2023-12-28
docs: improve custom plugin docs

hoebbelsB:
Slight improvements for the custom plugin section of the cli docs.

More feedback for the custom-plugin detail page:

Instead of using `MyAudit`, use something actually meaningful. Also add a minimal example in the very beginning of the docs section, where you describe that you can basically implement "anything". Maybe you can specify "anything" with an actual example.


BioPhoton:
I used myAudit is the first example that was unspecific and used file-size and lighthouse below.

Should I use audit-1?


For the hint about "anything". 

It should be mentioned in the main readme top. But always better on multiple places.

Where would you suggest to put it in addition?

Tlacenka (approved):
Nice language improvements!

PR #393, opened 2024-01-04
ci: fix labeler workflow for prs from fork

BioPhoton:
As PR's from forks fail (see #381) we need to adopt the trigger for `pull_request` to `pull_request_target`.
Docs can be found here: https://github.com/actions/labeler#permissions

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,7 +1,7 @@
 name: PR Labeler
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize]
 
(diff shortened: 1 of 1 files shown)


matejchalk (approved):

.github/workflows/pr-labeler.yml
```diff
@@ -1,7 +1,7 @@
 name: PR Labeler
 
 on:
-  pull_request:
+  pull_request_target:
```
Tlacenka, line 4: I noticed that the PR labeller check is skipped after this change. Does it mean it will not work for internal PRs anymore?

PR #397, opened 2024-01-08
refactor(examples-plugins): optimize Nx project configuration

matejchalk:
Cleaned up the `examples-plugins` Nx project, in order to improve DX and bring it more in line with Nx conventions.
- No longer imports from `.../dist/packages/utils` in `src` folder, tests and lint targets no longer depend on `build`.
- No longer ignores the `@nx/enforce-module-boundaries` rule.

Made some other related improvements:
- Added custom targets to root project for running Code PushUp command on the whole repo (e.g. `npx nx run-collect`), automatically builds dependencies using `dependsOn`. 
- ESBuild config supports non-publishable libraries - creates minimal `package.json` automatically, includes necessary metadata so imports from `dist` work correctly (e.g. references correct type declarations).
  - Removed `package.json` from `testing-utils` - it only needs to be buildable (because of Nx lint), not publishable. 
- Added a `scope:internal` Nx tag and added missing tags to Nx projects.
- Contributing docs are linked from README.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,5 @@
+# for uploading to portal
+CP_SERVER=
+CP_API_KEY=
+CP_ORGANIZATION=
+CP_PROJECT=
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -46,7 +46,11 @@
               },
               {
                 "sourceTag": "type:feature",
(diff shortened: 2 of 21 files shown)


Tlacenka (approved):
Very nice 🍓 !

vmasek (dismissed):

Tlacenka (approved):

PR #403, opened 2024-01-10
chore: remove ESLint issues from test files in core, plugin-eslint and utils

Tlacenka:
I went through the report and removed most ESLint related issues in test files and enabled rules where possible.

There are still ESLint issues in models (I will deal with those during my test refactor), nx-plugin (not relevant atm) and a few waiting for the rule to be enabled (so that I can disable the ESLint for that line)

## Stats
Before: 189 errors, 364 warnings, 10 info
After: 183 errors, 271 warnings, 10 info
Link to report displayed in portal [here](https://quality-metrics-staging.web.app/portal/code-pushup/cli/branch/fix-eslint-in-test-files/dashboard).

### DIFF
--- a/packages/core/src/lib/implementation/persist.unit.test.ts
+++ b/packages/core/src/lib/implementation/persist.unit.test.ts
@@ -1,6 +1,6 @@
-import { readFileSync } from 'fs';
 import { vol } from 'memfs';
-import { join } from 'path';
+import { readFile } from 'node:fs/promises';
+import { join } from 'node:path';
 import { beforeEach, describe, expect, it } from 'vitest';
 import { Report } from '@code-pushup/models';
 import { MEMFS_VOLUME, MINIMAL_REPORT_MOCK } from '@code-pushup/testing-utils';
@@ -41,7 +41,7 @@ describe('persistReport', () => {
     });
 
     const jsonReport: Report = JSON.parse(
-      readFileSync(join(MEMFS_VOLUME, 'report.json')).toString(),
+      await readFile(join(MEMFS_VOLUME, 'report.json'), 'utf8'),
     );
     expect(jsonReport).toEqual(
       expect.objectContaining({
@@ -50,9 +50,9 @@ describe('persistReport', () => {
       }),
     );
 
-    expect(() => readFileSync(join(MEMFS_VOLUME, 'report.md'))).toThrow(
-      'no such file or directory',
-    );
+    await expect(() =>
+      readFile(join(MEMFS_VOLUME, 'report.md')),
+    ).rejects.toThrow('no such file or directory');
   });
 
   it('should create a report in md format', async () => {
@@ -62,12 +62,12 @@ describe('persistReport', () => {
       format: ['md'],
     });
 
-    const mdReport = readFileSync(join(MEMFS_VOLUME, 'report.md')).toString();
+    const mdReport = await readFile(join(MEMFS_VOLUME, 'report.md'), 'utf8');
     expect(mdReport).toContain('Code PushUp Report');
 
-    expect(() => readFileSync(join(MEMFS_VOLUME, 'report.json'))).toThrow(
-      'no such file or directory',
-    );
+    await expect(() =>
+      readFile(join(MEMFS_VOLUME, 'report.json'), 'utf8'),
+    ).rejects.toThrow('no such file or directory');
   });
 
   it('should create a report in all formats', async () => {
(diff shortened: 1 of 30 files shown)


packages/core/src/lib/implementation/persist.unit.test.ts
```diff
@@ -41,7 +41,7 @@ describe('persistReport', () => {
     });
 
     const jsonReport: Report = JSON.parse(
-      readFileSync(join(MEMFS_VOLUME, 'report.json')).toString(),
+      await readFile(join(MEMFS_VOLUME, 'report.json'), 'utf8'),
```
matejchalk, line 44: Today I found out how to get `readFile` to resolve to a `string` directly instead of a `Buffer` :grin: 

> If no encoding is specified (using `options.encoding`), the data is returned as a `Buffer` object. Otherwise, the data will be a string.

packages/utils/src/lib/progress.integration.test.ts
```diff
@@ -27,15 +27,15 @@ describe('getSingletonMultiProgressBars', () => {
...
+const progressBuffer = bars['logger'].progressBuffer;
 
 describe('getProgressBar', () => {
-  it('should init task', async () => {
+  it('should init task', () => {
     expect(bars.getIndex(taskAName)).toBeUndefined();
     getProgressBar(taskAName);
     expect(bars.getIndex(taskAName)).toBe(0);
-    expect(progressBuffer[0].trim()).toBe(
+    expect((progressBuffer[0] as string).trim()).toBe(
```
matejchalk, line 38: It's shorter to use `!` (allowed in tests):

```suggestion
    expect(progressBuffer[0]!.trim()).toBe(
```

packages/utils/src/lib/report.unit.test.ts
```diff
@@ -55,11 +55,12 @@ describe('countWeightedRefs', () => {
 
 describe('compareIssueSeverity', () => {
   it('should order severities in logically ascending order when used as compareFn with .sort()', () => {
-    expect(
-      (['error', 'info', 'warning'] satisfies IssueSeverity[]).sort(
-        compareIssueSeverity,
-      ),
-    ).toEqual(['info', 'warning', 'error']);
+    const issueArr = ['error', 'info', 'warning'] satisfies IssueSeverity[];
+    expect([...issueArr].sort(compareIssueSeverity)).toEqual([
```
matejchalk, line 59: I would just add an ESLint disable comment here instead, the cloning is unnecessary here, the array isn't used anywhere else.

packages/utils/src/lib/report.unit.test.ts
```diff
@@ -102,17 +103,12 @@ describe('loadReport', () => {
...
-      } as unknown as PluginReport,
-    ];
-
     vol.fromJSON(
       {
-        [`report.json`]: JSON.stringify(invalidReportMock),
+        [`report.json`]: JSON.stringify({
+          ...REPORT_MOCK,
+          plugins: [{ ...REPORT_MOCK.plugins[0], slug: '-Invalid_slug' }],
+        }),
```
matejchalk, line 111: To have autocomplete (`JSON.stringify` accepts `any`):

```suggestion
        [`report.json`]: JSON.stringify({
          ...REPORT_MOCK,
          plugins: [{ ...REPORT_MOCK.plugins[0]!, slug: '-Invalid_slug' }],
        } satisfies PluginReport),
```

testing-utils/src/lib/fixtures/report.mock.ts
```diff
@@ -1,4 +1,4 @@
-import { type Report } from '../../../../packages/models/src';
+import { type Report } from '@code-pushup/models';
```
matejchalk, line 1:
```suggestion
import type { Report } from '@code-pushup/models';
```

matejchalk (approved):

packages/utils/src/lib/progress.integration.test.ts
Tlacenka: Fixed!

testing-utils/src/lib/fixtures/report.mock.ts
Tlacenka: Actually, I didn't mean to do that (circular dependencies later...). But fixed the `type` nesting.

packages/utils/src/lib/report.unit.test.ts
Tlacenka: Alright, reverting this change - it can only be disabled once the rule itself is enabled and there are still other errors in this package.

matejchalk (approved):

PR #409, opened 2024-01-12
chore(cli): fix most ESLint issues

Tlacenka:
In this PR I focused on enabling ESLint rules, so that they can be used both for production and test code:
- I fixed most ESLint issues.
- I improved implementation or notation where applicable.
- I disabled ESLint where necessary.

Before: 
image

[After](https://quality-metrics-staging.web.app/portal/code-pushup/cli/branch/eslint-cli-fixes/dashboard):
image

### DIFF
--- a/packages/cli/src/lib/implementation/model.ts
+++ b/packages/cli/src/lib/implementation/model.ts
@@ -3,6 +3,7 @@ import { Format } from '@code-pushup/models';
 
 export type GeneralCliOptions = { config: string } & GlobalOptions;
 
+/* eslint-disable @typescript-eslint/naming-convention */
 export type PersistConfigCliOptions = {
   'persist.outputDir': string;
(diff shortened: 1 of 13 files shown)


packages/cli/src/lib/implementation/model.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
matejchalk, line 1: A more future-proof way is to disable a specific block, instead of the entire file.

image

matejchalk (dismissed):
Nice :+1:

packages/cli/src/lib/implementation/model.ts
Tlacenka: Good point! ty

matejchalk (approved):

PR #410, opened 2024-01-14
chore: fix core ESLint issues, validate persist.format values in CLI

Tlacenka:
In this PR, I attempted to resolve ESLint issues in the `core` package. Currently, only one rule is turned off due to type inconsistencies encountered both in `cli` and `core` that is raised in a separate issue #411.

Notable changes:
- Since `exists` is deprecated, I had to use `stat` with an additional directory check and extracted the function to `utils`.
- Aside from ESLint issues, I added validation for `persist.format` during `yargs` parsing in order to prevent invalid values being passed as an incorrect type throughout the CLI flow.

Before:
image

[After](https://quality-metrics-staging.web.app/portal/code-pushup/cli/branch/eslint-core-fixes/dashboard):
image

### DIFF
--- a/packages/cli/src/lib/yargs-cli.ts
+++ b/packages/cli/src/lib/yargs-cli.ts
@@ -7,6 +7,7 @@ import yargs, {
   Options,
   ParserConfigurationOptions,
 } from 'yargs';
+import { PersistConfig, formatSchema } from '@code-pushup/models';
 import { logErrorBeforeThrow } from './implementation/utils';
 
 /**
@@ -42,10 +43,13 @@ export function yargsCli<T = unknown>(
     .help()
     .version(false)
     .alias('h', 'help')
+    .check(args => {
+      const persist = args['persist'] as PersistConfig | undefined;
+      return persist == null || validatePersistFormat(persist);
+    })
     .parserConfiguration({
       'strip-dashed': true,
     } satisfies Partial<ParserConfigurationOptions>)
-    .array('persist.format')
     .coerce('config', (config: string | string[]) =>
       Array.isArray(config) ? config.at(-1) : config,
     )
@@ -89,4 +93,19 @@ export function yargsCli<T = unknown>(
 
   // return CLI object
   return cli as unknown as Argv<T>;
+}
+
+function validatePersistFormat(persist: PersistConfig) {
+  try {
+    if (persist?.format != null) {
+      persist.format.forEach(format => formatSchema.parse(format));
+    }
+    return true;
+  } catch {
+    throw new Error(
+      `Invalid persist.format option. Valid options are: ${Object.values(
+        formatSchema.Values,
+      ).join(', ')}`,
+    );
+  }
 }
--- a/packages/core/src/lib/implementation/execute-plugin.ts
+++ b/packages/core/src/lib/implementation/execute-plugin.ts
@@ -117,39 +117,40 @@ export async function executePlugins(
   plugins: PluginConfig[],
   options?: { progress: boolean },
 ): Promise<PluginReport[]> {
-  const { progress = false } = options || {};
+  const { progress = false } = options ?? {};
 
-  const progressName = 'Run Plugins';
-  const progressBar = progress ? getProgressBar(progressName) : null;
+  const progressBar = progress ? getProgressBar('Run plugins') : null;
 
   const pluginsResult = await plugins.reduce(async (acc, pluginCfg) => {
-    const outputs = await acc;
-
-    progressBar?.updateTitle(`Executing  ${chalk.bold(pluginCfg.title)}`);
+    progressBar?.updateTitle(`Executing ${chalk.bold(pluginCfg.title)}`);
 
     try {
       const pluginReport = await executePlugin(pluginCfg);
       progressBar?.incrementInSteps(plugins.length);
-      return outputs.concat(Promise.resolve(pluginReport));
-    } catch (e: unknown) {
+      return [...(await acc), Promise.resolve(pluginReport)];
+    } catch (error) {
       progressBar?.incrementInSteps(plugins.length);
-      return outputs.concat(
-        Promise.reject(e instanceof Error ? e.message : String(e)),
-      );
+      return [
+        ...(await acc),
+        Promise.reject(error instanceof Error ? error.message : String(error)),
+      ];
     }
   }, Promise.resolve([] as Promise<PluginReport>[]));
 
   progressBar?.endProgress('Done running plugins');
 
-  const errorsCallback = ({ reason }: PromiseRejectedResult) =>
+  const errorsCallback = ({ reason }: PromiseRejectedResult) => {
     console.error(reason);
+  };
   const results = await Promise.allSettled(pluginsResult);
 
   logMultipleResults(results, 'Plugins', undefined, errorsCallback);
 
   const { fulfilled, rejected } = groupByStatus(results);
-  if (rejected.length) {
-    const errorMessages = rejected.map(({ reason }) => reason).join(', ');
+  if (rejected.length > 0) {
+    const errorMessages = rejected
+      .map(({ reason }) => String(reason))
+      .join(', ');
     throw new Error(
       `Plugins failed: ${rejected.length} errors: ${errorMessages}`,
     );
--- a/packages/core/src/lib/implementation/execute-plugin.unit.test.ts
+++ b/packages/core/src/lib/implementation/execute-plugin.unit.test.ts
@@ -119,7 +119,7 @@ describe('executePlugins', () => {
         { progress: false },
       ),
     ).rejects.toThrow(
-      'Plugins failed: 1 errors: Audit metadata not found for slug node-version',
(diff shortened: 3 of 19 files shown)


packages/cli/src/lib/yargs-cli.ts
```diff
@@ -42,10 +43,13 @@ export function yargsCli<T = unknown>(
...
     .version(false)
     .alias('h', 'help')
+    .check(args => {
+      const persist = args['persist'] as PersistConfig | undefined;
+      return persist == null || validatePersistFormat(persist);
+    })
     .parserConfiguration({
       'strip-dashed': true,
     } satisfies Partial<ParserConfigurationOptions>)
-    .array('persist.format')
```
matejchalk, line 48: Not sure deleting this is a good idea. It's meant to parse `--persist.format json --persist.format md` and `--persist.format json md` as `['json', 'md']`.

packages/cli/src/lib/yargs-cli.ts
```diff
@@ -90,3 +94,15 @@ export function yargsCli<T = unknown>(
...
   return cli as unknown as Argv<T>;
 }
+
+function validatePersistFormat(persist: PersistConfig) {
+  try {
+    return (
+      persist?.format == null ||
+      persist.format.map(format => formatSchema.parse(format)).length ===
+        persist.format.length
+    );
```
matejchalk, line 104: This is written in an unintuitive way, `arr.map(...).length === arr.length` is a logical tautology. The point is that none of the `formatSchema.parse` calls through, right? In that case, use `forEach`, that's what's intended for side-effects.

```suggestion
    persist.format?.forEach(format => {
      formatSchema.parse(format);
    });
    return true;
```

packages/cli/src/lib/yargs-cli.ts
```diff
@@ -90,3 +94,15 @@ export function yargsCli<T = unknown>(
...
+  } catch {
+    throw new Error('Unknown "persist.format" option. Use "md" or "json".');
```
matejchalk, line 106: Would be more future-proof to print the allowed options from the same array that our validation uses (single source of truth), instead of hardcoding `"md" or "json"`.

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -117,39 +117,39 @@ export async function executePlugins(
...
 
-  const progressName = 'Run Plugins';
-  const progressBar = progress ? getProgressBar(progressName) : null;
-
-  const pluginsResult = await plugins.reduce(async (acc, pluginCfg) => {
-    const outputs = await acc;
+  const progressBar = progress ? getProgressBar('Run plugins') : null;
 
+  const pluginsResult = plugins.map(pluginCfg => {
     progressBar?.updateTitle(`Executing  ${chalk.bold(pluginCfg.title)}`);
```
matejchalk, line 125:
```suggestion
    progressBar?.updateTitle(`Executing ${chalk.bold(pluginCfg.title)}`);
```

packages/core/src/lib/implementation/execute-plugin.ts
```diff
@@ -117,39 +117,39 @@ export async function executePlugins(
...
+    } catch (error: unknown) {
       progressBar?.incrementInSteps(plugins.length);
-      return outputs.concat(
-        Promise.reject(e instanceof Error ? e.message : String(e)),
+      return Promise.reject(
+        error instanceof Error ? error.message : String(error),
       );
     }
-  }, Promise.resolve([] as Promise<PluginReport>[]));
+  });
```
matejchalk, line 137: Revert these changes, please :pray: We execute the plugins sequentially on purpose, executing them in parallel can affect the results of some plugins (e.g. Lighthouse measurements).

packages/core/src/lib/implementation/execute-plugin.unit.test.ts
```diff
@@ -118,9 +118,7 @@ describe('executePlugins', () => {
         ] satisfies PluginConfig[],
         { progress: false },
       ),
-    ).rejects.toThrow(
-      'Plugins failed: 1 errors: Audit metadata not found for slug node-version',
-    );
+    ).rejects.toThrow('Audit metadata not found for slug node-version');
```
matejchalk, line 121: Why shorten the assertion? The `Plugins failed: 1 errors`-prefix is also interesting to check.

packages/core/src/lib/implementation/json-to-gql.ts
```diff
@@ -4,73 +4,92 @@ import {
...
-        type:
-          ref.type === 'audit'
-            ? CategoryConfigRefType.Audit
-            : CategoryConfigRefType.Group,
-        weight: ref.weight,
-        slug: ref.slug,
-      })),
-    })),
+    plugins: pluginReportsToGql(report.plugins),
+    categories: categoryConfigsToGql(toArray(report.categories)),
```
matejchalk, line 23: Yes :+1:

packages/core/src/lib/implementation/json-to-gql.ts
```diff
@@ -4,73 +4,92 @@ import {
...
+        audit.details?.issues.map(issue => ({
+          message: issue.message,
+          severity: transformSeverity(issue.severity),
+          sourceEndColumn: issue.source?.position?.endColumn,
+          sourceEndLine: issue.source?.position?.endLine,
+          sourceFilePath: issue.source?.file,
+          sourceStartColumn: issue.source?.position?.startColumn,
+          sourceStartLine: issue.source?.position?.startLine,
+          sourceType: IssueSourceType.SourceCode,
+        })) ?? [],
```
matejchalk, line 65: Could be extracted as well.

packages/core/src/lib/implementation/runner.ts
```diff
@@ -27,14 +27,12 @@ export async function executeRunnerConfig(
   });
 
   // read process output from file system and parse it
-  let audits = await readJsonFile<AuditOutputs>(
+  const rawAudits = await readJsonFile<AuditOutputs>(
```
matejchalk, line 30: Hmm, we're missing an important step here - validating the contents of the JSON file (optimistically casting as `AuditOutputs` assumes something we don't know to be true). There's needs to be `auditOutputsSchema.parse(...)` call here somewhere (and in `executeRunnerFunction` as well). Note that if `outputTransform` is passed, the result of that function should be asserted, because the JSON could contain anything in this case (`rawAudits` is a misleading name, actually, I'd go with `outputs`).

testing-utils/src/lib/utils/minimal-config.mock.ts
```diff
@@ -43,7 +44,7 @@ export const MINIMAL_PLUGIN_CONFIG_MOCK: PluginConfig = {
     },
   ],
   runner: MINIMAL_RUNNER_FUNCTION_MOCK,
-};
+} satisfies PluginConfig;
```
matejchalk, line 47: Why use `satisfies` instead of a type annotation?

matejchalk (changes requested):

packages/cli/src/lib/yargs-cli.ts
Tlacenka: The `.array()` function seems to be identical to `type: 'array'` set directly in the options which is already [in place](https://github.com/code-pushup/cli/blob/main/packages/cli/src/lib/implementation/core-config-options.ts#L34-L37).
image

packages/cli/src/lib/yargs-cli.ts
Tlacenka: I agree. However, I got no clue how to get that option from the zod error :D any ideas?

Tlacenka: Update: Got it! It was unnecessarily tedious but I got the message 🎉

packages/cli/src/lib/yargs-cli.ts
matejchalk: Cool, then it was superfluous, and we don't need it :+1:

testing-utils/src/lib/utils/minimal-config.mock.ts
Tlacenka: Good point, fixed!

packages/core/src/lib/implementation/runner.ts
Tlacenka: The parsing happens in `execute-plugin` function so I changed the type of `audits` to `unknown` within the `runner`.

matejchalk: I see, so it was just that the types were lying :sweat_smile:

packages/core/src/lib/implementation/json-to-gql.unit.test.ts
```diff
@@ -0,0 +1,36 @@
+import { describe } from 'vitest';
+import { Issue } from '@code-pushup/models';
+import { issuesToGql } from './json-to-gql';
+
+describe('issuesToGql', () => {
```
matejchalk, line 5: Why only test the issues?

matejchalk (approved):

PR #416, opened 2024-01-15
refactor(cli): adopt core-config middleware logic

BioPhoton:
part of #361


packages/cli/src/lib/implementation/core-config.model.ts
```diff
@@ -7,13 +7,13 @@ export type PersistConfigCliOptions = {
...
 };
 
+/* eslint-disable @typescript-eslint/naming-convention */
 export type UploadConfigCliOptions = {
   'upload.organization': string;
   'upload.project': string;
   'upload.apiKey': string;
   'upload.server': string;
 };
-/* eslint-enable @typescript-eslint/naming-convention */
```
matejchalk, line 16: This should be reverted. ESLint directives don't work on blocks of code, only lines. An `eslint-disable` comment affects all subsequent lines until the end of the file. To disable only a specific block, but not following blocks, it needs to be wrapped in `eslint-disable` and `eslint-enable` comments.

```ts
enabled here 

/* eslint-disable @typescript-eslint/naming-convention */
disabled here
disabled here too
/* eslint-enable @typescript-eslint/naming-convention */

enabled here
```

Your change disables the rule for the whole file (and the second comment is actually redudant):

```ts
enabled here 

/* eslint-disable @typescript-eslint/naming-convention */
disabled here
/* eslint-disable @typescript-eslint/naming-convention */
still disabled 

disable here too
```

packages/cli/src/lib/implementation/core-config.options.ts
```diff
@@ -26,14 +31,17 @@ export function yargsPersistConfigOptionsDefinition(): Record<
...
     'persist.filename': {
       describe: 'Filename for the produced reports.',
       type: 'string',
+      default: PERSIST_FILENAME,
     },
     'persist.format': {
       describe: 'Format of the report output. e.g. `md`, `json`',
       type: 'array',
+      default: PERSIST_FORMAT,
     },
```
matejchalk, line 45: I'm afraid that adding defaults to the CLI args introduces a bug - **persist options can't be customized via config file**. The precedence is supposed to be `explicit CLI arg > explicit config prop > default (if applicable)`, but setting the defaults at Yargs level means they're handled as explicit CLI args. Config middleware does `persist: {  ...importedRc.persist, ...cliOptions.persist }`, so treats explicit and implicit CLI options the exact same, and because everything under `persist` has defaults, values from config file are always overridden.

Here is an example of where the bug occurs. You'll notice that although I didn't specify `persist.outputDir` or `persist.format` in the CLI, my custom values in `code-pushup.config.ts` are overridden by defaults.
image

And you'll notice it works correctly in the main branch, so the bug is introduced in this PR.
image

matejchalk (changes requested):

packages/cli/src/lib/implementation/core-config.model.ts
BioPhoton: Ho! ok thx! will revert it!

packages/cli/src/lib/implementation/core-config.options.ts
BioPhoton: Dang! So then lets circle back to the config option of `yargs`. I would really like to use a more maintainable way to handle CIL settings.

packages/cli/src/lib/implementation/core-config.middleware.ts
```diff
@@ -1,47 +1,38 @@
...
-    Required<CoreConfig>;
-  const importedRc = await readCodePushupConfig(config);
+  const {
+    config,
+    persist: cliPersist,
+    upload: cliUpload,
+    ...cliOptions
+  } = args as GeneralCliOptions & Required<CoreConfig>;
+  const rcOptions = await readCodePushupConfig(config);
+  const { persist: rcPersist = {}, upload: rcUpload = {} } = rcOptions;
```
BioPhoton, line 21:
```suggestion
  const { persist: rcPersist, upload: rcUpload } = rcOptions;
```

packages/cli/src/lib/implementation/core-config.options.ts
BioPhoton: Update:
I refactored it back and added tests. 
Will test other options in the future.

packages/cli/src/lib/implementation/global.utils.integration.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  ])(
+    'should handle general arguments for %s correctly',
+    async (testId, configKind, cliObj, generalResult) => {
+      const argv = await cliWithGlobalOptions(
+        cliObj as GeneralCliOptions,
+      ).parseAsync();
+      expect(argv).toEqual(expect.objectContaining(generalResult));
+    },
+  );
+});
```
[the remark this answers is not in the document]
BioPhoton: refactored. thx for the typing hind.

matejchalk (approved):

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
```diff
@@ -35,3 +51,73 @@ describe('coreConfigMiddleware', () => {
...
+  cliObj: Partial<GeneralCliOptions>,
+) =>
+  yargsCli<CoreConfig>(objectToCliArgs(cliObj), {
+    options: {
+      ...yargsCoreConfigOptionsDefinition(),
+    },
+    middlewares: [{ middlewareFunction: coreConfigMiddleware }],
+  });
+
+describe('cliWithConfigOptionsAndMiddleware', () => {
```
Tlacenka, line 65: These tests concern both `yargs` CLI and `middleware`. Therefore these tests should live elsewhere, e.g. in a file `cli.<scope>-test.ts` as this file is only supposed to test the middleware function.

Additionally, the `describe` name should either name an existing production code function or a simple description of what scope is being tested, not a test-only created helper function.

Tlacenka, line 65: For better test readability and maintainability, one should be able to read the test from top to bottom and understand what it does without scrolling around or investigating other files.

This means:
- clear use case name to understand the test purpose
- minimum helper functions or variables
- clear, explicit usage of values passed to the test and expected result values

Such a test could look e.g. like this (flow taken from the last test case):

```
  it('should return persist filename from config, outputDir from CLI and format default', async () => {
    await expect(
      yargsCli<CoreConfig>(
        [
          '--persist.outputDir=tmp-cli',
          `--config=${join(
            configDirPath,
            'code-pushup.persist-only-filename.config.ts',
          )}`,
        ],
        {
          options: { ...yargsCoreConfigOptionsDefinition() },
          middlewares: [{ middlewareFunction: coreConfigMiddleware }],
        },
      ).parseAsync(),
    ).resolves.toEqual(
      expect.objectContaining({
        persist: {
          outputDir: 'tmp-cli',
          format: ['json'],
          filename: 'rc-report',
        },
      }),
    );
  });
  ```

File: packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
Comment (Tlacenka) [inline packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts:1]: I wonder, wouldn't it better to have the new feature tested by unit tests with mocked `bundle-require`? The main logic being tested is the `cli` and `configMiddleware`, not correct parsing of the config (tested elsewhere).

**Review by Tlacenka — CHANGES_REQUESTED** [review]
I looked at both the test code and the way the tests are structured.

It would be great to keep focusing on test readability introduced in the [test strategy](https://github.com/code-pushup/cli/wiki/Testing-Strategy#good-practices).
I added some tips on how to improve that below. If you have any questions, let me know and we can go through it together.

File: packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
Comment (BioPhoton) [reply packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts:1]: I tested over writes from cli -> rc -> defaults. So i guess no.

File: packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts (hunk 5 above)
Comment (BioPhoton) [reply packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts:55]: adopted the test name.

File: packages/cli/src/lib/implementation/global.utils.integration.test.ts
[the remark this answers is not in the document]
Comment (BioPhoton) [reply packages/cli/src/lib/implementation/global.utils.integration.test.ts:1]: I tested over writes and defaults. Also i checked and they are not covered. 
Pls let me know which tests you mean 🙏

File: packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts [hunk 6]
```diff
@@ -35,3 +51,73 @@ describe('coreConfigMiddleware', () => {
...
+    filename: cliPersistOptions['persist.filename'],
+  };
+  const rcResultPersistOptions: PersistConfig = {
+    outputDir: 'rc-tmp',
+    format: ['json', 'md'],
+    filename: 'rc-report',
+  };
+
+  it.each<
+    [string, CoreConfigName, Partial<CoreConfigCliOptions>, PersistConfig]
```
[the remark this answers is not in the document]
BioPhoton: done

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
BioPhoton: I will move the tests after in the future. Is this ok?

packages/cli/src/lib/implementation/global.utils.integration.test.ts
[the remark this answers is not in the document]
Tlacenka: CLI defaults for `verbose` and `progress` are covered [here](https://github.com/code-pushup/cli/blob/main/packages/cli/src/lib/yargs-cli.integration.test.ts#L11-L17).
CLI overrides are covered [here](https://github.com/code-pushup/cli/blob/main/packages/cli/src/lib/yargs-cli.integration.test.ts#L83-L122).

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
Tlacenka: I made it work!

The unit tests would need their own mocked implementation of `bundleRequire` like so:
```
vi.mock('bundle-require', () => ({
  bundleRequire: vi.fn().mockImplementation(args => {
   // mocked value options
    const filepath = args.filepath as string;
    const filenameOnly = {
      ...CORE_CONFIG_MOCK,
      persist: { filename: 'rc-filename' },
    };

    const allPersistOptions = {
      ...CORE_CONFIG_MOCK,
      persist: { filename: 'rc-filename', format: ['md'], outputDir: 'rc' },
    };
     // branching logic based on queried file name
    return {
      mod: {
        default: filepath.includes('persist-only-filename')
          ? filenameOnly
          : allPersistOptions,
      },
    };
  }),
}));
```

And then the test could look like this:
```
 beforeEach(() => {
    vol.fromJSON(
      {
        'persist-only-filename.config.ts': '', // only used for fileExists, not for value
      },
      MEMFS_VOLUME,
    );
  });

  it('should return persist filename from config, outputDir from CLI and format default', async () => {
    await expect(
      yargsCli<CoreConfig>(
        [
          '--persist.outputDir=cli-outputdir',
          '--config=./persist-only-filename.config.ts', // cwd is also mocked
        ],
        {
          options: { ...yargsCoreConfigOptionsDefinition() },
          middlewares: [{ middlewareFunction: coreConfigMiddleware }],
        },
      ).parseAsync(),
    ).resolves.toEqual(
      expect.objectContaining({
        persist: {
          filename: 'rc-filename', // from config
          outputDir: 'cli-outputdir', // from CLI
          format: ['json'],
        },
      }),
    );
  });
});
```

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
```diff
@@ -35,3 +50,80 @@ describe('coreConfigMiddleware', () => {
...
+  it('should take values for persist when provided over the rc', async () => {
+    const expectedObject = rcResultPersistOptions;
+    const argv = await cliWithConfigOptionsAndMiddleware({
+      config: configPath('persist'),
+    }).parseAsync();
+
+    expect(argv?.persist).toEqual(expect.objectContaining(expectedObject));
+  });
+
+  it('should take values for persist when provided over both, the rc config and the cli', async () => {
```
Tlacenka, line 116: It would be valuable to highlight here that if all three types of sources exist (default, config, CLI), the CLI takes precedence.

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
BioPhoton: wouldn't that require a file per test and 2 mocks per case?
Seems hart to handle :(

Tlacenka: No, the tests would become unit tests so the fs and cwd would be mocked so no config files would be needed. The additional data would be in virtual files in memfs whose name would be used in the mocked bundle require to identify the use case and return value.

The mocked config values themselves would merely derived from the default CORE_CONFIG_MOCK by adding persist option(s) (see persist-only-filename above).

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
BioPhoton: added the test

testing-utils/src/lib/utils/dynamic-mocks/config.mock.ts
```diff
@@ -36,6 +36,7 @@ export function minimalConfigMock(
   const outputFile = `${PLUGIN_1_SLUG}.${Date.now()}.json`;
 
   const cfg = coreConfigSchema.parse({
+    persist: persistConfigMock({ outputDir }),
```
Tlacenka, line 39: @BioPhoton This change I was not sure about - I wouldn't meddle with unrelated mocks in this PR. If you insist on this change and checked that it does not break existing tests, feel free to reintroduce this change of course.

packages/cli/src/lib/implementation/core-config.options-and-middleware.integration.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  return memfs.fs.promises;
+});
+
+let cwdSpy: SpyInstance;
+
+describe('parsing values from CLI and middleware', () => {
+  beforeEach(() => {
+    cwdSpy = vi.spyOn(process, 'cwd').mockReturnValue(MEMFS_VOLUME);
+
+    // value set in bundle-require mock above
```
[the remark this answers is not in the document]
BioPhoton: ok lets keep it then

Tlacenka (approved):
Nice test improvements, the `readCodePushupConfig` mock is better 😁

PR #418, opened 2024-01-15
refactor(cli): rename only-plugins utils

BioPhoton:


packages/cli/src/lib/implementation/only-plugins.utils.unit.test.ts
```diff
@@ -1,15 +1,15 @@
...
 import {
-  filterCategoryByOnlyPluginsOption,
-  filterPluginsByOnlyPluginsOption,
+  filterCategoryByPluginSlug,
+  filterPluginsBySlug,
   validateOnlyPluginsOption,
 } from './only-plugins.utils';
 
-describe('filterPluginsByOnlyPluginsOption', () => {
+describe('filterCategoryByPluginSlug', () => {
```
Tlacenka, line 9:
```suggestion
describe('filterPluginsBySlug', () => {
```

packages/cli/src/lib/implementation/only-plugins.utils.unit.test.ts
```diff
@@ -37,7 +37,7 @@ describe('filterPluginsByOnlyPluginsOption', () => {
 describe('filterCategoryByOnlyPluginsOption', () => {
```
Tlacenka, line 37:
```suggestion
describe('filterCategoryByPluginSlug', () => {
```

Tlacenka (commented):
The `describes` do not reflect the function tested within them anymore - left suggestions.

Otherwise a good naming improvement, the original names were way too long. 👍

Tlacenka (approved):

Issue #421, opened 2024-01-17
Handling plugins with no groups in `ScoredReport`


Tlacenka:
Currently, `ScoredReport` has groups as `EnrichedScoredGroup[]` which does not make sense as groups are generally optional.
This is used e.g. in `getGroupWithAudits` inside `generateMdReport` or inside `sortReport`.

## Acceptance criteria
- [ ] Groups are correctly typed and handled as optional during report scoring, sorting and generation.
- [ ] `@typescript-eslint/no-unnecessary-condition` rule is enabled or its stats are updated

PR #422, opened 2024-01-17
chore(utils): fix ESLint issues, enable ESLint rules

### LINKED ISSUE #265: lint fixes in utils (2/7) - code-style (various, simple)
- [ ] fix all lint errors and warnings in `utils` package (listed below)
- [ ] if a rule doesn't fail anymore for other packages, remove the `"off"` rule entry from last override in `.eslintrc.json`

## Current errors and warnings

Image

Image

## How to fix failing rules

- [no-magic-numbers](https://eslint.org/docs/latest/rules/no-magic-numbers)
  - move into named constants (e.g. `const MAX_SLUG_LENGTH = 128`)
- [no-console](https://eslint.org/docs/latest/rules/no-console)
  - if intended to be printed to stdout, use `console.info`, otherwise remove (`console.log` for debugging shouldn't be committed) 
- [@typescript-eslint/no-shadow](https://typescript-eslint.io/rules/no-shadow/)
  - rename variables to non-conflicting names
- [@typescript-eslint/no-unnecessary-condition](https://typescript-eslint.io/rules/no-unnecessary-condition/)
  - remove unneeded `?.` and `||` operators
- [sonarjs/no-duplicate-string](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-duplicate-string.md)
  - define a constant
 - [unicorn/import-style](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v48.0.1/docs/rules/import-style.md) 
   - e.g. `import { join } from 'path'` and `join(...)` ➡ `import path from 'node:path'` and `path.join()` 
 - [unicorn/prefer-string-replace-all](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v48.0.1/docs/rules/prefer-string-replace-all.md)
   - autofixable, but requires changing `lib` in `tsconfig.base.json` to `es2021` 
- [@typescript-eslint/dot-notation](https://typescript-eslint.io/rules/dot-notation/)
  - autofixable, but usage in `progress.spec.ts` accesses private property

### LINKED ISSUE #267: lint fixes in utils (3/7) - functional
- [ ] fix all lint errors and warnings related to `no-param-reassign`, `functional/immutable-data` and  `functional/no-let` rules
- [ ] if a rule doesn't fail anymore for other packages, remove the `"off"` rule entry from last override in `.eslintrc.json`

## Current errors and warnings

Image

Image

## How to fix failing rules

Requiring manual attention:
- [functional/immutable-data](https://github.com/eslint-functional/eslint-plugin-functional/blob/v6.0.0/docs/rules/immutable-data.md) 
  - use object and array spread (`...`) to create new objects instead of mutating the existing object 
  - e.g. `const runnerConfigMock = runnerConfig(); runnerConfigMock.outputFile = ' ';` ➡ `const runnerConfigMock = { ...runnerConfig(), outputFile: ' ' }`
- [functional/no-let](https://github.com/eslint-functional/eslint-plugin-functional/blob/v6.0.0/docs/rules/no-let.md)
  - either make it a `const` if you can initialize it one expression, or put the changed value into a different `const` variable 
- [no-param-reassign](https://eslint.org/docs/latest/rules/no-param-reassign) 
  - don't assign to parameter, introduce a new variable instead if needed

### LINKED ISSUE #268: lint fixes in utils (4/7) - complexity
- [ ] fix all lint errors and warnings related to complexity rules
- [ ] if a rule doesn't fail anymore for other packages, remove the `"off"` rule entry from last override in `.eslintrc.json`

## Current errors and warnings

Image

## How to fix failing rules

Requiring manual attention:
- warnings in `perf` folder
  - add `perf/**/*.ts` glob to test overrides (less strict)
- [max-lines-per-function](https://eslint.org/docs/latest/rules/max-lines-per-function) and [sonarjs/cognitive-complexity](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/cognitive-complexity.md)
  - break up into smaller functions

### LINKED ISSUE #269: lint fixes in utils (5/7) - promises
- [ ] fix all lint errors and warnings related to mishandled promises (listed below)
- [ ] if a rule doesn't fail anymore for other packages, remove the `"off"` rule entry from last override in `.eslintrc.json`

## Current errors and warnings

Image

## How to fix failing rules

Requiring manual attention:
- [@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises/)
  - add `await` 
- [@typescript-eslint/require-await](https://typescript-eslint.io/rules/require-await/)
  - remove `async` (if `await` not needed)

### LINKED ISSUE #271: lint fixes in utils (7/7) - type-safety
TODO

Tlacenka:
Closes #265 
Closes #267 
Closes #268 
Closes #269 
Closes #271 
In this PR, I tackle the rest of the ESLint issues in `utils`. Notable changes and findings:
- Again, I was not able to resolve `no-unnecessary-condition` due to #421
- In perf files, some of the types are wrong. For the sake of its experimental nature, however, I opted to disable ESLint rules instead. Relevant notes were added there.
- There is one case where I hesitated - replacing `pop` with `slice`. Should I leave it or not? 🙏 

Before:
image

[After](https://quality-metrics-staging.web.app/portal/code-pushup/cli/branch/eslint-utils-fixes-vol2/dashboard):
image

### DIFF
--- a/packages/utils/perf/crawl-file-system/index.ts
+++ b/packages/utils/perf/crawl-file-system/index.ts
@@ -6,19 +6,16 @@ import {
 } from '../../src/lib/file-system';
 import { crawlFileSystemOptimized0 } from './optimized0';
 
-// TODO pls double check me here... also, is pop that much faster that ESLint should be disabled instead?
-const PROCESS_ARGUMENT_TARGET_DIRECTORY = String(
+const PROCESS_ARGUMENT_TARGET_DIRECTORY =
   process.argv
     .find(arg => arg.startsWith('--directory'))
     ?.split('=')
-    .slice(0, -1)[0] ?? '',
-);
-const PROCESS_ARGUMENT_PATTERN = String(
+    .at(-1) ?? '';
+const PROCESS_ARGUMENT_PATTERN =
   process.argv
     .find(arg => arg.startsWith('--pattern'))
     ?.split('=')
-    .slice(0, -1)[0] ?? '',
-);
+    .at(-1) ?? '';
 
 const suite = new Benchmark.Suite('report-scoring');
 
@@ -44,7 +41,7 @@ const listeners = {
           2,
         )} sec`,
       );
-      console.info(`Fastest is ${suite.filter('fastest').map('name')}`);
+      console.info(`Fastest is ${String(suite.filter('fastest').map('name'))}`);
     }
   },
 };
--- a/packages/utils/perf/crawl-file-system/index.ts
+++ b/packages/utils/perf/crawl-file-system/index.ts
@@ -6,17 +6,18 @@ import {
 } from '../../src/lib/file-system';
 import { crawlFileSystemOptimized0 } from './optimized0';
 
+// TODO pls double check me here... also, is pop that much faster that ESLint should be disabled instead?
 const PROCESS_ARGUMENT_TARGET_DIRECTORY = String(
   process.argv
     .find(arg => arg.startsWith('--directory'))
     ?.split('=')
-    .pop() || '',
+    .slice(0, -1)[0] ?? '',
 );
 const PROCESS_ARGUMENT_PATTERN = String(
   process.argv
     .find(arg => arg.startsWith('--pattern'))
     ?.split('=')
-    .pop() || '',
+    .slice(0, -1)[0] ?? '',
 );
 
 const suite = new Benchmark.Suite('report-scoring');
--- a/packages/utils/perf/score-report/optimized0.ts
+++ b/packages/utils/perf/score-report/optimized0.ts
@@ -44,8 +44,6 @@ function categoryRefToScore(
           );
         }
         return group.score;
-      default:
-        throw new Error(`Type ${ref.type} is unknown`);
     }
   };
 }
--- a/packages/utils/perf/score-report/optimized1.ts
+++ b/packages/utils/perf/score-report/optimized1.ts
@@ -1,5 +1,8 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-assignment */
+// Note: The plugins of the ScoredReport are not structured correctly, hence the ESLint disables.
 import { Report } from '@code-pushup/models';
 import { ScoredReport } from '../../src';
+import { ScoredCategoryConfig } from '../../src/lib/reports/scoring';
 
 export function calculateScore<T extends { weight: number }>(
   refs: T[],
@@ -63,7 +66,7 @@ export function scoreReportOptimized1(report: Report): ScoredReport {
       });
       return categoryMap;
     },
-    new Map(),
+    new Map<string, ScoredCategoryConfig>(),
   );
 
   return {
--- a/packages/utils/perf/score-report/optimized1.ts
+++ b/packages/utils/perf/score-report/optimized1.ts
@@ -13,6 +13,7 @@ export function calculateScore<T extends { weight: number }>(
   return numerator / denominator;
 }
 
+// eslint-disable-next-line max-lines-per-function
 export function scoreReportOptimized1(report: Report): ScoredReport {
   const allScoredAuditsAndGroupsMap = new Map();
 
--- a/packages/utils/src/lib/file-system.ts
+++ b/packages/utils/src/lib/file-system.ts
@@ -12,7 +12,7 @@ export async function readTextFile(path: string): Promise<string> {
 
 export async function readJsonFile<T = unknown>(path: string): Promise<T> {
   const text = await readTextFile(path);
-  return JSON.parse(text);
+  return JSON.parse(text) as T;
 }
 
 export async function fileExists(path: string): Promise<boolean> {
@@ -76,7 +76,7 @@ export class NoExportError extends Error {
 }
 
 export async function importEsmModule(options: Options): Promise<unknown> {
-  const { mod } = await bundleRequire({
+  const { mod } = await bundleRequire<object>({
     format: 'esm',
     ...options,
   });
@@ -84,7 +84,6 @@ export async function importEsmModule(options: Options): Promise<unknown> {
   if (!('default' in mod)) {
     throw new NoExportError(options.filepath);
   }
-
   return mod.default;
 }
 
--- a/packages/utils/src/lib/reports/utils.ts
+++ b/packages/utils/src/lib/reports/utils.ts
@@ -95,8 +95,7 @@ export function getSeverityIcon(
 }
 
 export function calcDuration(start: number, stop?: number): number {
-  stop ??= performance.now();
-  return Math.floor(stop - start);
+  return Math.floor((stop ?? performance.now()) - start);
 }
 
 export function countWeightedRefs(refs: CategoryRef[]) {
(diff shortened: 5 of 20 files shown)


packages/utils/perf/crawl-file-system/index.ts
```diff
@@ -6,17 +6,18 @@ import {
...
-    .pop() || '',
+    .slice(0, -1)[0] ?? '',
 );
 const PROCESS_ARGUMENT_PATTERN = String(
   process.argv
     .find(arg => arg.startsWith('--pattern'))
     ?.split('=')
-    .pop() || '',
+    .slice(0, -1)[0] ?? '',
 );
```
matejchalk, line 21: The problem is `pop` mutates an array. However, here it doesn't really matter, because it's an intermediate array created by `split`, which is immediately mutated and never used again (can't be, since it's not assigned to a variable). Unfortunately, the ESLint rule isn't always able to distinguish these cases.

I think there is a better solution than either `.pop()` or `.slice(0, -1)[0]` - use `.at(-1)` to get last array item :slightly_smiling_face: 

(Oh and on an unrelated note - wrapping the whole thing in `String()` is almost certainly pointless. The underlying expression cannot return anything other than a string.)

packages/utils/perf/score-report/optimized0.ts
```diff
@@ -45,7 +45,7 @@ function categoryRefToScore(
         }
         return group.score;
       default:
-        throw new Error(`Type ${ref.type} is unknown`);
+        throw new Error(`Type ${String(ref.type)} is unknown`);
```
matejchalk, line 48: What is the type of `ref.type` here. If it's `never`, we should remove the `default` case. It's unreachable, and will prevent TypeScript warning us if we forget to handle some possible value here.

(Same goes for other `optimized*.ts` files.)

packages/utils/perf/score-report/optimized1.ts
```diff
@@ -1,5 +1,12 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+
+/* eslint-disable @typescript-eslint/no-unsafe-return */
+
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
matejchalk, line 5: The rules can be on one line. This is the syntax, I think:

```suggestion
/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-assignment */
```

packages/utils/src/lib/file-system.ts
```diff
@@ -12,6 +12,7 @@ export async function readTextFile(path: string): Promise<string> {
 
 export async function readJsonFile<T = unknown>(path: string): Promise<T> {
   const text = await readTextFile(path);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
   return JSON.parse(text);
```
matejchalk, line 16: This way is maybe a bit nicer:

```suggestion
  return JSON.parse(text) as T;
```

packages/utils/src/lib/file-system.ts
```diff
@@ -85,6 +87,7 @@ export async function importEsmModule(options: Options): Promise<unknown> {
     throw new NoExportError(options.filepath);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access
   return mod.default;
```
matejchalk, line 91: Weird that ESLint complains about this. We assign it to `unknown` return type, which I would consider very safe :sweat_smile:

packages/utils/src/lib/reports/utils.ts
```diff
@@ -251,10 +247,10 @@ export async function loadReport<T extends Format>(
 
   if (format === 'json') {
     const content = await readJsonFile(filePath);
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-return
     return reportSchema.parse(content) as any;
   }
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-return
   return readTextFile(filePath) as any;
```
matejchalk, line 254: This looks nasty :grimacing: Why are the `as any`s needed?

packages/utils/src/lib/file-system.ts
Tlacenka: The problem is, `mod` is `any` as returned by `bundleRequire`
image

One can type the function itself if they know what to expect in `mod`.

packages/utils/src/lib/reports/utils.ts
Tlacenka: Apparently, TypeScript does not understand how we distinguish the type and therefore screams.
image

packages/utils/perf/score-report/optimized0.ts
Tlacenka: > (Same goes for other optimized*.ts files.)
Not sure I understand. This logic is only in `optimized0`.

matejchalk: > > (Same goes for other optimized*.ts files.)
>
> Not sure I understand. This logic is only in optimized0.


My mistake, I put this in the wrong comment :man_facepalming: Was meant to be here I think: https://github.com/code-pushup/cli/pull/422#discussion_r1456001506

packages/utils/src/lib/file-system.ts
matejchalk: > One can type the function itself if they know what to expect in mod. But when I do bundleRequire<unknown> then the (!('default' in mod)) complains 😅

You're on the right track :slightly_smiling_face: Just needs some tweak so that `in` operator is safe, TS needs to know `mod` is an object. Then no ESLint disables are necessary.

- If you cast it as `object`, the error goes away.
  ```ts
  export async function importEsmModule(options: Options): Promise<unknown> {
    const { mod } = await bundleRequire<object>({
      format: 'esm',
      ...options,
    });

    if (!('default' in mod)) {
      throw new NoExportError(options.filepath);
    }

    return mod.default;
  }
  ```
- If you cast it as `unknown`, then you need to add runtime checks to verify it's an object.
  ```ts
  export async function importEsmModule(options: Options): Promise<unknown> {
    const { mod } = await bundleRequire<unknown>({
      format: 'esm',
      ...options,
    });

    if (mod == null || typeof mod !== 'object' || !('default' in mod)) {
      throw new NoExportError(options.filepath);
    }

    return mod.default;
  }
  ```

packages/utils/src/lib/reports/utils.ts
matejchalk: I see. Yeah, this is pretty common when the return type depends on a generic, some casting is needed.

But if you cast as the return type instead, you can get rid of the ESLint disable comments.

```ts
  if (format === 'json') {
    const content = await readJsonFile(filePath);
    return reportSchema.parse(content) as LoadedReportFormat<T>;
  }
  const text = await readTextFile(filePath);
  return text as LoadedReportFormat<T>;
```

matejchalk (approved):

PR #423, opened 2024-01-17
chore: clean up ESLint in plugins

Tlacenka:
I went through our plugins and remaining packages and cleaned up what I could.

However, especially `nx-plugin` still has a lot of issues in it and once we start working on it, the first thing should be enabling the ESLint rules and cleaning up what is there.

Before:
image

[After](https://quality-metrics-staging.web.app/portal/code-pushup/cli/branch/eslint-plugin-fixes/dashboard):
image


packages/plugin-eslint/src/lib/meta/rules.ts
```diff
@@ -27,7 +28,7 @@ export async function listRules(
     {
       messages: rulesIds.map(ruleId => ({ ruleId })),
       suppressedMessages: [] as Linter.SuppressedLintMessage[],
-    } as ESLint.LintResult,
+    } as unknown as ESLint.LintResult,
```
matejchalk, line 31: Really not a fan of `as unknown as T`. It completely ignores any compatibiliy, while for `as T` TS at least checks it's "similar". Maybe adding `as Linter.LintMessage[]` to `messages` would work?

packages/plugin-eslint/src/lib/meta/rules.ts
```diff
@@ -13,6 +13,7 @@ export async function listRules(
   patterns: string | string[],
 ): Promise<RuleData[]> {
   const configs = await toArray(patterns).reduce(
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
     async (acc, pattern) => [
       ...(await acc),
       await eslint.calculateConfigForFile(pattern),
```
matejchalk, line 19:
```suggestion
    async (acc, pattern) => [
      ...(await acc),
      (await eslint.calculateConfigForFile(pattern)) as Linter.Config,
```

matejchalk (approved):
Looks good overall :slightly_smiling_face:

packages/plugin-eslint/src/lib/meta/rules.ts
Tlacenka: No, it does not work. The ESLint problem is that you are using `as` as opposed to `satisfies` 🙄 . So I disabled that.

matejchalk: OK, then you picked the best solution :+1:

matejchalk (approved):

PR #428, opened 2024-01-23
fix(utils): set fixed locale for date in report.md

### LINKED ISSUE #374: [utils] integration test snapshot relies on `en-US` locale
### What happened?

When I run integration tests for `@code-pushup/utils`, the snapshot for `report-to-stdout.integration.test.ts` only works when I set my operating system to the `en-US` locale.

This is caused by using `new Date().toString` instead of `new Date().toLocaleString('en-US, { /* (...) */ })` in the `reportToAboutSection` function. Alternatively, we could add a [development container](https://containers.dev/) to the project.

Tests should work regardless of operating system locale to make it easier to contribute to the project.

### What steps did you take?

1. Change the regional formatting settings of your operating system to something other than `en-US`, in my case `da-DK`.
1. Run the `npx nx integration-test utils` command.
1. `report-to-md.integration.test.ts` fails

Windows

### Node version

18.19.0

### Relevant log output

```shell
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/lib/report-to-md.integration.test.ts > report-to-md > should contain all sections when using the fixture report
Error: Snapshot `report-to-md > should contain all sections when using the fixture report 1` mismatched

- Expected
+ Received

  ## About

- Report was created by [Code PushUp](https://github.com/flowup/quality-metrics-cli#readme) on Sat Jan 01 2000 00:00:00 GMT+0000 (Coordinated Universal Time).
+ Report was created by [Code PushUp](https://github.com/flowup/quality-metrics-cli#readme) on Sat Jan 01 2000 00:00:00 GMT+0000 (Koordineret universaltid).

❯ src/lib/report-to-md.integration.test.ts:28:22
     26|       `${commit.message} (${commit.hash.slice(0, 7)})`,
     27|     );
     28|     expect(mdReport).toMatchSnapshot();
       |                      ^
     29|   });
     30| });
```

matejchalk:
Closes #374 (cc @LayZeeDK)

### DIFF
--- a/packages/utils/src/lib/formatting.ts
+++ b/packages/utils/src/lib/formatting.ts
@@ -56,8 +56,13 @@ export function formatDate(date: Date): string {
   const locale = 'en-US'; // fixed locale to ensure consistency
   return date
     .toLocaleString(locale, {
-      dateStyle: 'medium',
-      timeStyle: 'short',
+      weekday: 'short',
+      month: 'short',
+      day: 'numeric',
+      year: 'numeric',
+      hour: 'numeric',
+      minute: '2-digit',
+      timeZoneName: 'short',
     })
     .replace(/\u202F/g, ' '); // see https://github.com/nodejs/node/issues/45171
 }
--- a/packages/utils/src/lib/formatting.ts
+++ b/packages/utils/src/lib/formatting.ts
@@ -52,6 +52,16 @@ export function formatDuration(duration: number): string {
   return `${(duration / 1000).toFixed(2)} s`;
 }
 
+export function formatDate(date: Date): string {
+  const locale = 'en-US'; // fixed locale to ensure consistency
+  return date
+    .toLocaleString(locale, {
+      dateStyle: 'medium',
+      timeStyle: 'short',
+    })
+    .replace(/\u202F/g, ' '); // see https://github.com/nodejs/node/issues/45171
+}
+
 export function truncateText(text: string, maxChars: number): string {
   if (text.length <= maxChars) {
     return text;
--- a/packages/utils/src/lib/formatting.ts
+++ b/packages/utils/src/lib/formatting.ts
@@ -53,7 +53,7 @@ export function formatDuration(duration: number): string {
 }
 
 export function formatDate(date: Date): string {
-  const locale = 'en-US'; // fixed locale to ensure consistency
+  const locale = 'en-US'; // fixed locale to ensure consistency across local defaults execution
   return date
     .toLocaleString(locale, {
       weekday: 'short',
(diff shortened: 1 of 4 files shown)


packages/utils/src/lib/reports/__snapshots__/generate-md-report.integration.test.ts.snap
```diff
@@ -421,7 +421,7 @@ ESLint rule **prefer-arrow-callback**. [📖 Docs](https://eslint.org/docs/lates
 
 ## About
 
-Report was created by [Code PushUp](https://github.com/flowup/quality-metrics-cli#readme) on Wed Sep 15 2021 00:00:00 GMT+0000 (Coordinated Universal Time).
+Report was created by [Code PushUp](https://github.com/flowup/quality-metrics-cli#readme) on Sep 15, 2021, 12:00 AM.
```
LayZeeDK, line 424: suggestion: It's useful to include the time zone or UTC offset to avoid confusion and mental overhead of jugging time zone differences

LayZeeDK (changes requested):
Suggest change

packages/utils/src/lib/reports/__snapshots__/generate-md-report.integration.test.ts.snap
matejchalk: Good call, added it :+1:

LayZeeDK (approved):

packages/utils/src/lib/reports/generate-md-report.ts
```diff
@@ -229,7 +229,8 @@ function reportToAboutSection(
   report: ScoredReport,
   commitData: CommitData | null,
 ): string {
-  const date = new Date().toString();
+  const date = formatDate(new Date());
```
Tlacenka, line 232: Question - do we then want to have the date in two formats or just one? In `core` and `execute-process` we use ISO format (see `collect.ts`, `runner.ts` and relevant tests that use `ISO_STRING_REGEXP`).

matejchalk: The human-friendly format should only be used when we're printing something for the user. But ISO is still needed when we need those dates to be parseable later (we send the dates from `report.json` to server) - `toISOString` is perfect for serializing dates so that they will be parsed correctly by pretty much anything, `toLocaleDateString` has no such guarantee.

Tlacenka (dismissed):
Code quality looks good to me, tests added as well. ✔️ 
I looked at the latest CI results and verified that the correct date format was used. ✔️

vmasek (dismissed):

packages/utils/src/lib/formatting.ts
```diff
@@ -52,6 +52,21 @@ export function formatDuration(duration: number): string {
   return `${(duration / 1000).toFixed(2)} s`;
 }
 
+export function formatDate(date: Date): string {
+  const locale = 'en-US'; // fixed locale to ensure consistency
```
BioPhoton, line 56:
```suggestion
  const locale = 'en-US'; // fixed locale to ensure consistency across local defaults execution
```

BioPhoton (commented):
Added a suggestion

vmasek (approved):

PR #429, opened 2024-01-23
feat(core): add upload timeout

matejchalk:
Long-running upload requests are unlikely to ever resolve past a certain interval. To prevent hanging CI jobs (which cost money), a default timeout of 5 minutes is set (in `@code-pushup/portal-client` - see https://github.com/code-pushup/portal/pull/230) after which upload will fail. This value is customizable via `upload.timeout` in config file.

### DIFF
--- a/packages/models/src/lib/upload-config.ts
+++ b/packages/models/src/lib/upload-config.ts
@@ -9,6 +9,11 @@ export const uploadConfigSchema = z.object({
   }),
   organization: slugSchema('Organization slug from Code PushUp portal'),
   project: slugSchema('Project slug from Code PushUp portal'),
+  timeout: z
+    .number({ description: 'Request timeout in minutes (default is 5)' })
+    .positive()
+    .int()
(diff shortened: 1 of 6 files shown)


packages/models/src/lib/upload-config.ts
```diff
@@ -9,6 +9,11 @@ export const uploadConfigSchema = z.object({
   }),
   organization: slugSchema('Organization slug from Code PushUp portal'),
   project: slugSchema('Project slug from Code PushUp portal'),
+  timeout: z
+    .number({ description: 'Request timeout in minutes (default is 5)' })
```
Tlacenka, line 13: Zod supports a default value. Or is it cleaner to only have it set on `portal-client`?

matejchalk: Yeah, I didn't want to have it set in two places.

BioPhoton (approved):
LGTM

PR #431, opened 2024-01-25
test: replace bundle-require mock by readCodePushUpConfig, update testing-utils README

### LINKED ISSUE #357: test improvements
- [x] output.json in root as test artefact => move into tmp
- [x] comments for where unexpected behaviour is given due to mocks
But in general we should not hide specific mocks for libs IMHO
Screenshot 2023-12-12 at 17 55 15

- [x] reduce console.logs

Tlacenka:
Closes #357 

In this PR, I remove `bundle-require` setup and instead include the mocks in individual files for better understanding of where the value comes from.

As a bonus, I updated the `testing-utils` README to introduce the types of utilities that can be found there.


testing-utils/README.md
```diff
@@ -1,7 +1,30 @@
...
 This library contains testing utilities, such as helper functions or fixtures.
+Any reusable mocks should live here.
 
-## Running unit tests
+## Main utilities
 
-Run `npx nx run testing-utils:unit-test` to execute the unit tests via [Vitest](https://vitest.dev).
+### Mock data
+
+Hardcoded mocks, often referring to a smaller configuration object, live in `utils`.
```
matejchalk, line 10: It's nice to have links to the folders, makes it easier to find the right one (in this case I thought you meant `packages/utils` at first :man_facepalming:).

```suggestion
Hardcoded mocks, often referring to a smaller configuration object, live in [`utils`](./src/lib/utils).
```

(Same goes for other folder and file references in this README.)

testing-utils/README.md
```diff
@@ -1,7 +1,30 @@
...
+
+Example configuration files that are to be used in integration or E2E tests live in `fixtures/configs`.
+
+### Mock setup
+
+In `setup` folder you can find all files that can be used in `vitest.config` `setupFiles`. Currently include:
+
+- console mocking
+- file system mocking
+- `portal-client` mocking
```
matejchalk, line 25: Maybe link to the file here as well?

packages/cli/src/lib/upload/upload-command.unit.test.ts
```diff
@@ -4,20 +4,32 @@ import { beforeEach, describe, expect, it } from 'vitest';
...
+// mock bundleRequire inside importEsmModule used for fetching config
+vi.mock('bundle-require', async () => {
+  const { CORE_CONFIG_MOCK }: typeof import('@code-pushup/testing-utils') =
+    await vi.importActual('@code-pushup/testing-utils');
+  return {
+    bundleRequire: vi
+      .fn()
+      .mockResolvedValue({ mod: { default: CORE_CONFIG_MOCK } }),
+  };
+});
```
matejchalk, line 23: What if we extracted this to a helper function in `testing-utils`, with `CORE_CONFIG_MOCK` as an overridable default? :thinking:

Something like this:

- default config content:
  ```ts
  import { mockBundleRequire } from `@code-pushup/testing-utils`;
  
  mockBundleRequire();
  
  describe('...', () => {
    // ...
  })
  ```;
- custom config content:

  ```ts
  import { mockBundleRequire, CORE_CONFIG_MOCK } from `@code-pushup/testing-utils`;
  
  mockBundleRequire(({ filepath }) => ({
    ...CORE_CONFIG_MOCK,
    persist: {
      outputDir: `tmp/${getExtension(filepath)}`,
    },
  }));
  
  describe('...', () => {
    // ...
  });
  ```

:top: Just a suggestion, not sure how well this can work with `vi.mock` because of hoisting. And I could also understand the counter-argument that maybe adding an abstraction layer is less transparent (depends a lot on how robust the abstraction is, I think).

matejchalk (commented):
I really like the `testing-utils` docs :slightly_smiling_face: Maybe we could also link it with the [Testing Strategy](https://github.com/code-pushup/cli/wiki/Testing-Strategy) wiki page?

testing-utils/README.md
Tlacenka: I'm dumb, I forgot I can link internal references 🤦 will do.

packages/cli/src/lib/upload/upload-command.unit.test.ts
Tlacenka: The primary reason for splitting it was to directly add the config value to the same file as the tests (while previously one had to go to `testing-utils`).

So if it was a helper function where you would pass it the value you want it to return, sure. But here you need to deal with inserting value based on `filepath` into `mod: { default: SMTH } }`... I wouldn't complicate it for now.

packages/cli/src/lib/collect/collect-command.unit.test.ts
```diff
@@ -7,6 +7,17 @@ import { DEFAULT_CLI_CONFIGURATION } from '../../../mocks/constants';
 import { yargsCli } from '../yargs-cli';
 import { yargsCollectCommandObject } from './collect-command';
 
+// mock bundleRequire inside importEsmModule used for fetching config
+vi.mock('bundle-require', async () => {
```
BioPhoton, line 11: same, we could adjust the depth of mocking to just the config loading

testing-utils/README.md
Tlacenka: Fixed everywhere.

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -8,6 +8,17 @@ import { DEFAULT_CLI_CONFIGURATION } from '../../../mocks/constants';
 import { yargsCli } from '../yargs-cli';
 import { yargsAutorunCommandObject } from './autorun-command';
 
+// mock bundleRequire inside importEsmModule used for fetching config
+vi.mock('bundle-require', async () => {
```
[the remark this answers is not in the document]
Tlacenka: Good point, done.

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
```diff
@@ -1,18 +1,23 @@
...
 import { DEFAULT_CLI_CONFIGURATION } from '../../../mocks/constants';
 import { yargsCli } from '../yargs-cli';
 import { yargsAutorunCommandObject } from './autorun-command';
 
 vi.mock('@code-pushup/core', async () => {
+  const { CORE_CONFIG_MOCK }: typeof import('@code-pushup/testing-utils') =
+    await vi.importActual('@code-pushup/testing-utils');
   const core = await vi.importActual('@code-pushup/core');
   return {
     ...(core as object),
```
matejchalk, line 18:
```suggestion
  const core: typeof import('@code-pushup/core') =
    await vi.importActual('@code-pushup/core');
  return {
    ...core,
```

(Applies to `collect-command.unit.test.ts` and `upload-command.unit.test.ts`, too.)

packages/cli/src/lib/upload/upload-command.unit.test.ts
```diff
@@ -1,23 +1,33 @@
...
+
 describe('upload-command-object', () => {
   beforeEach(() => {
     vol.fromJSON(
       {
         'my-report.json': JSON.stringify(MINIMAL_REPORT_MOCK),
-        'code-pushup.config.ts': '', // only needs to exist for stat inside readCodePushupConfig
       },
-      '/test',
+      MEMFS_VOLUME,
```
BioPhoton, line 30: Candidate for a linting rule :)

BioPhoton (dismissed):
Nice mock refactor! I like that many fs mocks are not needed anymore

packages/cli/src/lib/upload/upload-command.unit.test.ts
Tlacenka: While there are ESLint rules for hardcoded values and repeated literals, they are disabled in the test preset I believe (cc @matejchalk) as hardcoding values in test is not necessarily wrong. Here I replaced it to be consistent with other files.

packages/cli/src/lib/print-config/print-config-command.unit.test.ts
```diff
@@ -1,23 +1,17 @@
...
-      },
-      MEMFS_VOLUME,
-    );
-  });
+vi.mock('@code-pushup/core', async () => {
+  const { CORE_CONFIG_MOCK }: typeof import('@code-pushup/testing-utils') =
+    await vi.importActual('@code-pushup/testing-utils');
+  return {
+    readCodePushupConfig: vi.fn().mockResolvedValue(CORE_CONFIG_MOCK),
+  };
```
matejchalk, line 11: I'd probably preserve the other exported functions from `@code-pushup/core` here as well. Even if they're not called for now, that could change in the future and would lead to pretty strange test error.

packages/cli/src/lib/autorun/autorun-command.unit.test.ts
Tlacenka: That's how it's fixed? Brilliant, I remember we ran into this in one of the previous PRs but dismissed it. Glad it has a proper solution 😁

matejchalk: Yeah, the `const module: typeof import('<path>') = await vi.importActual('<path>')` pattern can be applied in general. The type has to annotated explicitly because I don't believe it's possible for `vi.importActual` to infer it from the path, only built-in language constructs can do that (i.e. static `import` and dynamic `import()`).

packages/cli/src/lib/upload/upload-command.unit.test.ts
matejchalk: We don't have a lint rule that covers this. Production code has [sonarjs/no-duplicate-string](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-duplicate-string.md), but that's disabled in tests for good reason (DAMP vs DRY).

matejchalk (dismissed):

matejchalk (approved):

PR #437, opened 2024-01-27
refactor: make categories optional in core config 

### LINKED ISSUE #407: Make categories optional
### User story

As a user, I want to be able to run Code PushUp without setting up any categories. This allows me to see a simplified list of audits.

### Acceptance criteria

- [ ] `CoreConfig` and `Report` accept missing or empty `categories` array.

### Implementation details

_No response_

BioPhoton:
Closes #407 

I aimed for handling the optional categories only in the cli layer and maintained the existing data structure fully in reports. I just hide the sections for empty categories.

Follow-up by @Tlacenka:
- I added tests for scoring and sorting reports.
- I allowed passing both `undefined` and `[]` by configuring categories schema to allow empty array.
- I reduced minimal mocks.
- I typed categories as a non-optional property as soon as in the only plugins middleware.


packages/cli/src/lib/implementation/only-plugins.utils.ts
```diff
@@ -14,12 +14,16 @@ export function filterPluginsBySlug(
 // skip the whole category if it has at least one skipped plugin ref
 // see https://github.com/code-pushup/cli/pull/246#discussion_r1392274281
 export function filterCategoryByPluginSlug(
-  categories: CoreConfig['categories'],
+  categories: Array<CategoryConfig>,
```
matejchalk, line 17: For consistency:

```suggestion
  categories: CategoryConfig[],
```

Pre-emptive answer: Yes, there is a lint rule for this ([array-type](https://typescript-eslint.io/rules/array-type/)) :grin: I've created an issue for adding more of these stylistic rules to our ESLint config: https://github.com/code-pushup/eslint-config/issues/4

packages/cli/src/lib/implementation/core-config.integration.test.ts
```diff
@@ -137,4 +143,16 @@ describe('parsing values from CLI and middleware', () => {
       outputDir: 'cli-outputdir',
     });
   });
+
+  it('should fill categories if not given in config file', async () => {
```
matejchalk, line 147: The word "fill" somewhat implies we have some meaningful defaults to me, I'd more explicit that it's just `undefined` :arrow_right: `[]`.

```suggestion
  it('should set empty array if categories not given in config file', async () => {
```

packages/cli/src/lib/collect/collect-command.ts
```diff
@@ -18,6 +18,12 @@ export function yargsCollectCommandObject(): CommandModule {
       console.info(chalk.bold(CLI_NAME));
       console.info(chalk.gray(`Run ${command}...`));
       await collectAndPersistReports(options);
+      if (options.categories.length === 0) {
+        console.info(
+          chalk.gray(`💡Configure categories to see the scores in an overview table.
+        See: https://github.com/code-pushup/cli/blob/main/packages/cli/README.md`),
```
matejchalk, line 24: Is the indentation here intentional? Whitespace in template strings is printed "as is" (including line breaks), so this would print:

```
💡Configure categories to see the scores in an overview table.
        See: https://github.com/code-pushup/cli/blob/main/packages/cli/README.md
```

This would print it on one line (if you want only the source code to span multiple lines, use `+` concatenation):

```suggestion
          chalk.gray('💡 Configure categories to see the scores in an overview table. See: https://github.com/code-pushup/cli/blob/main/packages/cli/README.md'),
```

packages/cli/README.md
```diff
@@ -67,11 +64,14 @@ _If you're looking for programmatic usage, then refer to the underlying [@code-p
...
 
-4. Define your custom categories.
+4. Optionally define your custom categories.
 
    ```js
    export default {
      // ...
+     plugins: [
+       // ...
+     ],
```
matejchalk, line 74: The `// ...` allows us to omit the plugins for brevity:

```suggestion
     // ...
```

packages/utils/src/lib/reports/generate-stdout-summary.ts
```diff
@@ -17,11 +17,13 @@ function addLine(line = ''): string {
 }
 
 export function generateStdoutSummary(report: ScoredReport): string {
+  const { categories = [] } = report;
+  const printCategories = categories.length > 0 ? true : false;
```
matejchalk, line 21: The `categories` aren't optional in `Report` (only in `CoreConfig`). And `<boolean expr> ? true : false` is redundant. So this can be simplified:

```suggestion
  const printCategories = report.categories.length > 0;
```

packages/utils/src/lib/reports/generate-md-report.ts
```diff
@@ -41,25 +41,29 @@ export function generateMdReport(
   report: ScoredReport,
   commitData: CommitData | null,
 ): string {
+  const { categories = [] } = report;
+  const reportWithCategories = { ...report, categories };
+  const printCategories = categories.length > 0 ? true : false;
```
matejchalk, line 46: Can be simplified for same reasons as in `generate-stdout-summary.ts`:

```suggestion
  const printCategories = categories.length > 0;
```

Then you can just use the original `report` below instead of the extra `reportWithCategories` variable.

matejchalk (commented):
We could also consider having some prompt to create categories in place of the missing category sections in `report.md` or stdout summary :thinking: But that's a different discussion and out of scope of this PR, I believe.

packages/cli/src/lib/implementation/core-config.integration.test.ts
Tlacenka: Fully agree.

packages/cli/src/lib/implementation/core-config.integration.test.ts
```diff
@@ -32,11 +32,17 @@ vi.mock('@code-pushup/core', async () => {
...
+        plugins: CORE_CONFIG_MOCK.plugins,
+        upload: CORE_CONFIG_MOCK.upload,
+      };
 
       return filepath.includes('all-persist-options')
         ? allPersistOptions
         : filepath.includes('no-persist')
         ? noPersistFilename
+        : filepath.includes('no-category')
+        ? noCategoryFilename
```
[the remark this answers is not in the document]
Tlacenka: Naming improved.

packages/cli/src/lib/collect/collect-command.ts
Tlacenka: Made it into a one-liner. Added it to both autorun and collect command.

packages/cli/README.md
Tlacenka: Good point!

packages/utils/src/lib/reports/generate-stdout-summary.ts
Tlacenka: Simplified in both generate files.

packages/utils/src/lib/reports/generate-md-report.ts
Tlacenka: Simplified.

Tlacenka:
>We could also consider having some prompt to create categories in place of the missing category sections in report.md or stdout summary 🤔

@matejchalk 
While we log a suggestion, that could get lost in the history. So I agree that having that prompt in the report itself would make sure that people know this option exists and where to find more info about it. Can you create an issue for this, please?

packages/cli/README.md
```diff
@@ -67,20 +64,21 @@ _If you're looking for programmatic usage, then refer to the underlying [@code-p
...
-         title: 'Performance',
+         slug: 'bug prevention',
+         title: 'Bug prevention',
          refs: [
+           // reference to an existing audit or group from plugins
            {
              type: 'audit',
              plugin: 'eslint',
-             slug: 'react-jsx-key',
+             slug: 'no-any',
```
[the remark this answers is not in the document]
Tlacenka: My bad, did not know that was a thing 😅 will revert.

matejchalk:
> While we log a suggestion, that could get lost in the history. So I agree that having that prompt in the report itself would make sure that people know this option exists and where to find more info about it. Can you create an issue for this, please?

Done :heavy_check_mark: 
- #445

matejchalk (approved):

BioPhoton (commented):
You could update the snippets in 'docs/custom-plugind.md'

matejchalk (approved):

PR #454, opened 2024-02-01
docs(cli): use typed config in examples

matejchalk:
Follow-up to #449 

Motivation: This way user gets autocomplete in the config file :slightly_smiling_face:


vmasek (approved):
Amazing, I was wondering why don't types autocomplete when assisting Ondrej

PR #456, opened 2024-02-01
docs(examples-plugins): update file-size.plugin.ts

BioPhoton:


JeanMeche (approved):
LGTM

examples/plugins/src/file-size/src/file-size.plugin.ts
```diff
@@ -52,7 +52,7 @@ export const recommendedRefs: CategoryRef[] = Object.values(auditsMap).map(
  * // code-pushup.config.ts
  * import {
  *   create as fileSizePlugin,
- *   recommendedRef as fileSizeRecommendedRefs
+ *   recommendedRefs as fileSizeRecommendedRefs
```
vmasek, line 55: Should it also be changed [here](https://github.com/code-pushup/cli/blob/ab047a45c018ef92f6bd7c243caa5e09554cdc09/examples/plugins/src/package-json/src/package-json.plugin.ts#L39)?

BioPhoton: Good point! Done..

vmasek (approved):

Issue #451, opened 2024-02-01
Setup the logger with @poppinss/cliui


vmasek:

BioPhoton:
@vmasek I'm missing details for the following  requests above:
- include durations of the actions
- print summary in the end (last sentence)

PR #459, opened 2024-02-03
feat(cli): add logger

### LINKED ISSUE #371: 📖 Hint to run CLI and further options
## General Hint

- You can run the command with `--verbose` to see the more detailed information about the process
- Read about custom plugins here: LINK

## Hint after `collect` run

- Use upload command to get statistics, and deeper insights in your data over the portal. 
   See: https://portal.com#getting started
- Use autorun to collect and upload at once

## Hint after `upload` run

n/a

## Hint after `autorun` run
- To get the report results in every PR use our CI integration guide:
   https://ci-integration.com

## Hint after `--help` run

-  links to the online docs: getting started, custom plugins, CI setup

BioPhoton:
This is one of many PRs to refactor our logging related code.

It includes:
- intro new lib
- use logger for info logs
- use logger for hints
  - better styling for hints
  
  closes #371 related to #451
  
  Example:
Screenshot 2024-02-03 at 13 14 22
Screenshot 2024-02-03 at 13 13 56
Screenshot 2024-02-03 at 03 30 18
Screenshot 2024-02-05 at 21 08 10


packages/cli/src/lib/autorun/autorun-command.ts
```diff
@@ -36,26 +43,14 @@ export function yargsAutorunCommandObject() {
...
-        console.info(
-          chalk.gray(
-            '💡 Configure categories to see the scores in an overview table. See: https://github.com/code-pushup/cli/blob/main/packages/cli/README.md',
-          ),
-        );
+        renderConfigureCategoriesHint(ui);
       }
 
-      if (options.upload) {
+      if (false) {
```
BioPhoton, line 49:
```suggestion
      if (options.upload) {
```

packages/utils/src/index.ts
Tlacenka, line 1: Do I understand it correctly that this logger should replace any `console.info` and `console.warn` and `console.error` instances that report the state of the CLI?

If so, I would expect the following functions to also replace console with this logger:
`verboseUtils`, `logMultipleFileResults`, `logPromiseResults`, `filterCategoryByPluginSlug` and `startLocalRegistry`.

Tlacenka: I see that that ^ is out of scope. Just read the PR intro again, my bad.
It would be valuable to include all the instances in AC of the epic/issue that this is a part of. 🙏

Tlacenka (commented):
Overall great improvement, like the formatting and upload suggestions in a block now 😁

packages/utils/src/index.ts
BioPhoton: it should be here: https://github.com/code-pushup/cli/issues/451

packages/cli/src/lib/print-config/print-config-command.unit.test.ts
```diff
@@ -25,17 +31,39 @@ describe('print-config-command', () => {
...
     );
-    expect(console.info).toHaveBeenCalledWith(
-      expect.stringContaining('"outputDir": "destinationDir"'),
+    expect(log).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          stream: 'stdout',
+          message: expect.stringContaining('"outputDir": "destinationDir"'),
+        }),
+      ]),
```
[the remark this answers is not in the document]
Tlacenka: I currently see vast majority of cases covering methods that print to `stdout` (only one test in CLI currently asserts `console.error`).

Alternatively, the helper/setup of the assertion should imo at least get rid of the array and other properties of the object, so that we are just asserting equality for `{ stream: 'stdout', message: <some string>}` and nothing else.

packages/cli/src/lib/upload/upload-command.ts
```diff
@@ -1,31 +1,34 @@
...
-          ].join('\n'),
-        );
+        renderIntegratePortalHint();
         throw new Error('Upload configuration not set');
       }
       await upload(options);
+      const commitData = await getLatestCommit();
+      if (!commitData) {
+        throw new Error('no commit data available');
+      }
```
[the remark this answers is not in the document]
BioPhoton: done. Testing will be done in #469

packages/cli/src/lib/print-config/print-config-command.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: done

packages/utils/src/lib/reports/utils.ts
```diff
@@ -304,3 +305,29 @@ export function compareIssues(a: Issue, b: Issue): number {
...
+
+export function portalCommitLink(
+  config: Pick<UploadConfig, 'project' | 'organization' | 'server'> & {
+    baseUrl?: string;
+  },
+  commit: string,
+): string {
+  const { organization, project, server } = config;
+  // https://portal-api-r6nh2xm7mq-ez.a.run.app/graphql
+  const urlObj = new URL(config.baseUrl || server);
```
matejchalk, line 317: We don't have a reliable way of creating this link without knowing the portal URL. The API URL is very likely to be different, so trying to derive the path from it will almost certainly produce a broken link.

I'd suggest that we don't interpolate the URL at all and instead print a placeholder `<YOUR_PORTAL_URL>/portal/...`, then at least the user can copy the path and they don't have the friction of a broken link.

If you think it's valuable to provide a portal link, then please create an issue for it - one approach would be to add a property to the config file, another that the server returns the portal link in the response (the 2nd would be my preference).

matejchalk (changes requested):

packages/utils/src/lib/reports/utils.ts
BioPhoton: only placed base url there and filled it with `<YOUR_PORTAL_URL>` for now

matejchalk (approved):

PR #464, opened 2024-02-05
test(utils): adapt to different process stdout behaviour

Tlacenka:
The `executeProcess` test was flaky due to environment-specific behaviour (pipelines vs local setup) where the `onStdout` spy was called once or twice (second call being with empty new line, see [here](https://github.com/code-pushup/cli/actions/runs/7785559898/job/21228444707)).

As a solution I removed the part of assertion demanding how many times the observer was called, leaving that it was called at least once and what its final value is. I think this is sufficient.


matejchalk (approved):
Thanks for the explanation :+1:

PR #469, opened 2024-02-05
feat(utils): add git helper

### LINKED ISSUE #216: History command

BioPhoton:
This PR includes utility functions related to git logic use in the future history command.

Methods included:
- `getCurrentBranchOrTag` - The history command needs to end in the same branch where it started. This method is called before we start walking the repo history to save the "initial" branch
- `statusIsClean` - To safely check out a branch the status needs to be clean 
- `guardAgainstLocalChanges` - the command should throw if the git status is not clean
- `safeCheckout` - as there might be some files generated while walking the history we need to do the following:
  - warn the user that the status is not clean (initial checkout)
  - drop all created changes after each run

partially resolves #216


packages/utils/src/lib/git.ts
```diff
@@ -17,3 +17,40 @@ export async function getLatestCommit() {
...
+  return await git.status(['-s']).then(r => r.files.length > 0);
+}
+
+export async function guardAgainstDirtyRepo(): Promise<void> {
+  const isDirty = await branchHasChanges();
+  if (isDirty) {
+    throw new Error(`
+        Repository should be clean before we you can proceed.
+        Commit your local changes or stash them.
+      `);
```
matejchalk, line 31:
```suggestion
    throw new Error('Repository should be clean before we you can proceed. Commit your local changes or stash them.');
```

packages/utils/src/lib/git.ts
```diff
@@ -17,3 +17,40 @@ export async function getLatestCommit() {
   });
   return log.latest;
 }
+
+export async function branchHasChanges(): Promise<boolean> {
+  return await git.status(['-s']).then(r => r.files.length > 0);
+}
```
matejchalk, line 23:
```suggestion
export function branchHasChanges(): Promise<boolean> {
  return git.status(['-s']).then(r => r.files.length > 0);
}
```

packages/utils/src/lib/git.ts
```diff
@@ -17,3 +17,40 @@ export async function getLatestCommit() {
...
+  }
+}
+
+export async function getCurrentBranchOrTag(): Promise<string> {
+  return (
+    (await git.branch().then(r => r.current)) ||
+    // @TODO replace with simple git
+    (await git.raw(['describe --tags --exact-match']).then(out => out.trim()))
+  );
+}
```
matejchalk, line 41: What if there is no current branch or tag?

packages/utils/src/lib/git.ts
```diff
@@ -17,3 +17,40 @@ export async function getLatestCommit() {
...
+  branchOrHash: string,
+  options: {
+    clean?: boolean;
+  } = {},
+): Promise<void> {
+  // git requires a clean history to check out a branch
+  if (options?.clean) {
+    await git.clean(['f']);
+    console.info(`branch cleaned`);
+  }
```
matejchalk, line 53: Is it necessary to remove untracked files?

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
     );
   });
 });
+
+describe('branchHasChanges', () => {
+  it('should return true if some changes are given', async () => {
+    await makeStatusDirty();
+    await expect(branchHasChanges()).resolves.toBe(true);
+    await makeStatusClean();
```
matejchalk, line 32: Don't put test clean up after the assertion, it won't be run if the test fails, which would then affect other tests (which is bad, tests should be independent of each other). Use the `afterEach` hook.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
...
+  });
+});
+
+describe('guardAgainstDirtyRepo', () => {
+  it('should throw if history is dirty', async () => {
+    await makeStatusDirty();
+    await expect(guardAgainstDirtyRepo()).rejects.toThrow(
+      'Repository should be clean before we you can proceed',
+    );
+    await makeStatusClean();
```
matejchalk, line 45: Same here, move to `afterEach`.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
...
+  });
+  it('should not throw if history is clean', async () => {
+    await expect(guardAgainstDirtyRepo()).resolves.toEqual(void 0);
```
matejchalk, line 48: `void 0` is a strange way to write `undefined` :sweat_smile: 

```suggestion
    await expect(guardAgainstDirtyRepo()).resolves.toBeUndefined();
```

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
...
+
+describe('getCurrentBranchOrTag', () => {
+  it('should log current branch', async () => {
+    await expect(getCurrentBranchOrTag()).resolves.toEqual(expect.any(String));
+  });
+});
+
+describe('safeCheckout', () => {
+  // eslint-disable-next-line vitest/require-hook
+  let initialBranch: string = '';
```
matejchalk, line 60:
```suggestion
  let initialBranch: string;
```

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
...
+    initialBranch = await getCurrentBranchOrTag();
+  });
+
+  afterEach(async () => {
+    await git.checkout(initialBranch);
+    await makeStatusClean();
+  });
+
+  it('should checkout target branch in clean state', async () => {
+    await expect(safeCheckout('main')).resolves.toBe(void 0);
```
matejchalk, line 72:
```suggestion
    await expect(safeCheckout('main')).resolves.toBeUndefined();
```

packages/utils/src/lib/git.integration.test.ts
matejchalk, line 1: Now I can't run integration tests when I have uncomitted changes :grimacing: 

image

matejchalk (changes requested):
Sorry, but requiring a clean repository state in order to run integration tests is a hard no from a DX perspective :no_good_man: Either make the tests the tests more robust, or don't test it.

packages/utils/src/lib/git.ts
```diff
@@ -17,3 +17,40 @@ export async function getLatestCommit() {
...
+export async function branchHasChanges(): Promise<boolean> {
+  return await git.status(['-s']).then(r => r.files.length > 0);
+}
+
+export async function guardAgainstDirtyRepo(): Promise<void> {
+  const isDirty = await branchHasChanges();
+  if (isDirty) {
+    throw new Error(`
+        Repository should be clean before we you can proceed.
+        Commit your local changes or stash them.
```
Tlacenka, line 30:
```suggestion
        Working directory needs to be clean before we you can proceed.
        Commit your local changes or stash them.
```

packages/utils/src/lib/git.ts
```diff
@@ -17,3 +17,40 @@ export async function getLatestCommit() {
   });
   return log.latest;
 }
+
+export async function branchHasChanges(): Promise<boolean> {
+  return await git.status(['-s']).then(r => r.files.length > 0);
+}
+
+export async function guardAgainstDirtyRepo(): Promise<void> {
```
Tlacenka, line 25: Similarly here, maybe it's just me but I would not know what a dirty repo is.

```suggestion
export async function guardAgainstLocalChanges(): Promise<void> {
```

testing-utils/src/index.ts
```diff
@@ -1,6 +1,9 @@
 export * from './lib/constants';
 export * from './lib/setup/test-folder.setup';
 export * from './lib/utils/execute-process-helper.mock';
+// utils
```
Tlacenka, line 4: Exports from lines 1-3 are also generic utils. Not sure if this comment is necessary.

packages/utils/src/lib/git.integration.test.ts
Tlacenka, line 60: This rule only triggers when one tries to instantiate the variable, not when it is only declared.
Seeing that it is assigned a value before execution of any tests, does the default value need to be there?

```suggestion
  let initialBranch: string;
```

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
...
+    await expect(getCurrentBranchOrTag()).resolves.toBe('main');
+  });
+  it('should throw if history is dirty', async () => {
```
Tlacenka, line 75:
```suggestion
  });
  
  it('should throw if history is dirty', async () => {
```

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
...
+    await makeStatusDirty();
+    await expect(safeCheckout('main')).rejects.toThrow(
+      'Repository should be clean before we you can proceed',
+    );
+  });
+
+  it('should use gitRestore option', async () => {
```
Tlacenka, line 82:
```suggestion
  it('should clean local changes and check out to main', async () => {
```

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
...
+    await makeStatusClean();
+  });
+  it('should not throw if history is clean', async () => {
+    await expect(guardAgainstDirtyRepo()).resolves.toEqual(void 0);
+  });
+});
+
+describe('getCurrentBranchOrTag', () => {
+  it('should log current branch', async () => {
+    await expect(getCurrentBranchOrTag()).resolves.toEqual(expect.any(String));
```
Tlacenka, line 54: Is there a way of knowing/setting the branch? `any(String)` is a weak assertion.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -16,3 +24,66 @@ describe('getLatestCommit', () => {
...
+
+describe('guardAgainstDirtyRepo', () => {
+  it('should throw if history is dirty', async () => {
+    await makeStatusDirty();
+    await expect(guardAgainstDirtyRepo()).rejects.toThrow(
+      'Repository should be clean before we you can proceed',
+    );
+    await makeStatusClean();
+  });
+  it('should not throw if history is clean', async () => {
```
Tlacenka, line 47:
```suggestion
  });
  
  it('should not throw if history is clean', async () => {
```

packages/utils/src/lib/git.integration.test.ts
matejchalk: **If** you did want to integration test these functions (for me it's quetionable if it pays off, the implementation is not that complex), then this could a way:
- before tests run, create a folder in `tmp` and run `git init`
- all git commands should run in this mini-repository, i.e. completely separated from our root repository
- after tests are finished, remove the folder so that the tests can be re-run reliably

packages/utils/src/lib/git.ts
BioPhoton: The repo is not clean otherwise. and it is optional.
I use it in the history command to revert repo changes.

packages/utils/src/lib/git.integration.test.ts
BioPhoton: done

packages/utils/src/lib/git.integration.test.ts
BioPhoton: done

packages/utils/src/lib/git.ts
BioPhoton: done

testing-utils/src/index.ts
BioPhoton: removed

packages/utils/src/lib/git.integration.test.ts
BioPhoton: The quickest would be to remove them :D.
What would be the best to do here?

packages/utils/src/lib/git.ts
matejchalk: No, mean what if the current commit isn't tagged. How do you handle this output?

image

packages/utils/src/lib/git.integration.test.ts
BioPhoton: min length or regex maybe?

packages/utils/src/lib/git.ts
BioPhoton: Throw?😇

I will implement a check for this case.

packages/utils/src/lib/git.integration.test.ts
matejchalk: I did something a little bit similar a while ago, it wasn't that hard in the end. In my case I used [MockGithub](https://github.com/kiegroup/mock-github?tab=readme-ov-file#mockgithub), but I'm not sure the library is really needed.

packages/utils/src/lib/git.integration.test.ts
Tlacenka: Is it not possible to determine which exact branch it is if the repo is created for tests only?

matejchalk: Yeah, that's the key factor, related to this thread - https://github.com/code-pushup/cli/pull/469#discussion_r1479469141

If it's the actual branch, then the tests shouldn't assume anything - I could have uncommitted changes, I could be in a detached state in which case there is no branch, I definitely don't want to a test to start switching my branch or deleting untracked files - that's why I don't want the test to depend on the real repo in any way.

If it is a temporary repo, then yes, the test can set any state it wants (without interfering with local development) and have way more specific assertions.

packages/utils/src/lib/git.integration.test.ts
matejchalk: So I've run into a couple of situations were I've needed this approach and it's actually pretty easy to achieve with just `simple-git` :slightly_smiling_face: - check out `git.integration.test.ts` in #503

BioPhoton: Thx for the link! that helps!

packages/utils/src/lib/git.ts
BioPhoton: I have no clue what i should do without the information. So I would go for the current implementation

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -21,12 +30,19 @@ describe('git utils', () => {
...
+
+    await git.checkout(['-b', 'feature-branch']);
   });
 
   afterAll(async () => {
     await rm(baseDir, { recursive: true, force: true });
   });
 
+  afterEach(async () => {
+    await rm(changesDir, { recursive: true, force: true });
```
Tlacenka, line 42: It would be cleaner to just delete the temporary branch that is used inside the test in the `afterEach()` instead of manual removal/addition of a folder.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -21,12 +30,19 @@ describe('git utils', () => {
 
     await git.add('README.md');
     await git.commit('Create README');
+
+    await git.checkout(['-b', 'feature-branch']);
```
Tlacenka, line 34: Does this mean the first test starts in `feature-branch` but all subsequent tests start in `master` due to the `afterEach`? That is not intuitive, please make it consistent.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -60,4 +76,86 @@ describe('git utils', () => {
...
+
+  it('safeCheckout should throw if history is dirty', async () => {
+    await mkdir(changesDir, { recursive: true });
+    await writeFile(join(changesDir, 'change.md'), '# hello-change\n');
+    await expect(safeCheckout('master', {}, git)).rejects.toThrow(
+      'Working directory needs to be clean before we you can proceed. Commit your local changes or stash them.',
+    );
+  });
+
+  it('safeCheckout should clean local changes and check out to master', async () => {
```
Tlacenka, line 122:
```suggestion
  it('safeCheckout should clean local changes and check out to feature-branch', async () => {
```

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -60,4 +76,86 @@ describe('git utils', () => {
...
+    // needs to get reset to be clean
+    await mkdir(changesDir, { recursive: true });
+    await writeFile(join(changesDir, 'change.md'), '# hello-change\n');
+    // needs to get cleaned to be clean
+    await writeFile(join(baseDir, 'README.md'), '# hello-world-2\n');
```
Tlacenka, line 127: Could be valuable to add a check that these changes are not present anymore using `git.status()`.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -60,4 +76,86 @@ describe('git utils', () => {
...
+    await git.init();
+  });
+
+  afterAll(async () => {
+    await rm(baseDir, { recursive: true, force: true });
+  });
+
+  it('getCurrentBranchOrTag should throw if no branch is given', async () => {
+    await expect(getCurrentBranchOrTag(git)).rejects.toThrow(
+      "git: 'describe --tags --exact-match' is not a git command. See 'git --help'",
```
Tlacenka, line 158: This command should return `fatal: No names found, cannot describe anything.` or similar. 

However, the asserted error message sounds like something went wrong with the command itself.

packages/utils/src/lib/git.ts
```diff
@@ -38,17 +38,60 @@ export async function toGitPath(
...
+  }
+}
+
+export async function getCurrentBranchOrTag(
+  git = simpleGit(),
+): Promise<string> {
+  return (
+    (await git.branch().then(r => r.current)) ||
+    // @TODO replace with simple git
+    (await git.raw(['describe --tags --exact-match']).then(out => out.trim()))
```
Tlacenka, line 77: Thought regarding the command error - shouldn't the arguments be separated?
```suggestion
    (await git.raw(['describe', '--tags', '--exact-match']).then(out => out.trim()))
```

matejchalk (dismissed):

Tlacenka (commented):
Not sure which methods are needed for the history command. A brief intro in the PR description explaining the goal would be helpful. 🙏 
Currently, I can only check the tests, not if the list of methods is sufficient.

Test feedback below.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -60,4 +76,86 @@ describe('git utils', () => {
...
+
+  it('statusIsClean should return false if some changes are given', async () => {
+    await mkdir(changesDir, { recursive: true });
+    await writeFile(join(changesDir, 'change.md'), '# hello-change\n');
+    await expect(statusIsClean(git)).resolves.toBe(false);
+  });
+
+  it('statusIsClean should return true if no changes are given', async () => {
+    await expect(statusIsClean(git)).resolves.toBe(true);
+  });
```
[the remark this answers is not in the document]
BioPhoton: ATM no. I will move it in and if we need it separate pull it out again

packages/utils/src/lib/git.integration.test.ts
BioPhoton: I can add a custom error. What should be the message?
(A suggestion helps as my formulations of Error are improvable always, plus i don't know git that well to fully understand all situations in which this function could potentially throw)

packages/utils/src/lib/git.integration.test.ts
BioPhoton: No it just created the branch

packages/utils/src/lib/git.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: done

packages/utils/src/lib/git.integration.test.ts
BioPhoton: done

packages/utils/src/lib/git.integration.test.ts
BioPhoton: done

packages/utils/src/lib/git.integration.test.ts
BioPhoton: done

packages/utils/src/lib/git.integration.test.ts
BioPhoton: message is no "could not checkout current branch or tag"

matejchalk (dismissed):

packages/utils/src/lib/git.integration.test.ts
Tlacenka: Did you push all the changes? It still removes the directory instead of removing the branch. I added a snippet of what I mean below.

packages/utils/src/lib/git.integration.test.ts
```diff
@@ -21,12 +29,20 @@ describe('git utils', () => {
...
   });
 
   afterAll(async () => {
     await rm(baseDir, { recursive: true, force: true });
   });
 
+  afterEach(async () => {
+    await rm(changesDir, { recursive: true, force: true });
+    await git.checkout(['master']);
+  });
```
Tlacenka, line 44:
```suggestion
  afterEach(async () => {
    await git.checkout(['master']);
    await git.deleteLocalBranch('feature-branch')
  });
```

And then in `beforeEach` e.g.
```
await git.checkout(['-b', 'feature-branch']);
```
If you want to start in that branch.

packages/utils/src/lib/git.ts
```diff
@@ -38,17 +38,67 @@ export async function toGitPath(
...
+      // @TODO use simple git
+      return await git
+        .raw(['describe', '--tags', '--exact-match'])
+        .then(out => out.trim());
+    }
+  } catch {
+    // Return a custom error message when something goes wrong
+    throw new Error('could not check out current tag or branch');
+  }
+}
```
Tlacenka, line 86: In the original comment, I was mostly pointing out that the error probably meant issues with the implementation rather than that the message was not good.

As for the current custom message, I suggest adjusting it in the following way - the command should not check out, just get ingo (unless I am missing something).

```suggestion
  } catch {
    // Return a custom error message when something goes wrong
    throw new Error('Could not get current tag or branch.');
  }
}
```

packages/utils/src/lib/git.integration.test.ts
BioPhoton: I thought I keep the checkout of the branch in the tests as there are just some of them.

BioPhoton: ok, for some reason i cant reset a branch. added a todo and made it with manual deletion so far

packages/utils/src/lib/git.integration.test.ts
BioPhoton: Now i pushed, but i was not able to remove the line. added a todo

packages/utils/src/lib/git.ts
```diff
@@ -38,17 +38,67 @@ export async function toGitPath(
...
+  branchOrHash: string,
+  options: {
+    forceCleanStatus?: true;
+  } = {},
+  git = simpleGit(),
+): Promise<void> {
+  // git requires a clean history to check out a branch
+  if (options?.forceCleanStatus) {
+    await git.raw(['reset', '--hard']);
+    await git.clean(['f', 'd']);
```
Tlacenka, line 98: Maybe this can be done to delete the branch?

Tlacenka (approved):
Looking good! 🧪 

Added one comment but can be done in a separate PR.

packages/utils/src/lib/git.ts
BioPhoton: I tried it. :(

PR #470, opened 2024-02-05
refactor: add eslint rules

BioPhoton:
dogfooding examples


packages/nx-plugin/src/utils/versions.ts
```diff
@@ -1,20 +1,26 @@
...
-  join(projectsFolder, 'models'),
+export const cpuCliVersion = (
+  await loadPackageJson(join(projectsFolder, 'models'))
 ).version;
 
-function loadPackageJson(folderPath: string) {
-  return JSON.parse(readFileSync(join(folderPath, 'package.json')).toString());
+async function loadPackageJson(folderPath: string): Promise<PkgJson> {
+  const packageJsonContent = await readFile(join(folderPath, 'package.json'));
+  return JSON.parse(packageJsonContent.toString()) as PkgJson;
```
Tlacenka, line 25: If encoding is defined while reading a file, the method returns a string instead of a buffer. Source [here](https://stackoverflow.com/questions/6456864/why-does-node-jss-fs-readfile-return-a-buffer-instead-of-string).

```suggestion
async function loadPackageJson(folderPath: string): Promise<PkgJson> {
  const packageJsonContent = await readFile(join(folderPath, 'package.json'), 'utf8');
  return JSON.parse(packageJsonContent) as PkgJson;
```

packages/nx-plugin/src/generators/init/generator.ts
```diff
@@ -18,15 +18,17 @@ import {
 } from '../../utils/versions';
 import { InitGeneratorSchema } from './schema';
 
+const nxPluginPackageName = '@code-pushup/nx-plugin';
+type PkgJson = Record<string, Record<string, string>>;
```
Tlacenka, line 22: Note: I think there is an official way of typing package.json but I we would need to introduce a new dependency. 😞

Tlacenka (dismissed):
Amazing! Now we have almost perfect score and all relevant packages as ESM!

Good job 👏

packages/nx-plugin/src/utils/versions.ts
matejchalk: We have a `readJsonFile` in `utils` you could use instead.

packages/nx-plugin/src/generators/init/generator.ts
matejchalk: Found it! 

```ts
import type { PackageJson } from 'nx/src/utils/package-json';
```

packages/nx-plugin/src/utils/versions.ts
```diff
@@ -1,20 +1,26 @@
...
-export const cpuUtilsVersion = loadPackageJson(
-  join(projectsFolder, 'utils'),
+export const cpuUtilsVersion = (
+  await loadPackageJson(join(projectsFolder, 'utils'))
 ).version;
-export const cpuCliVersion = loadPackageJson(
-  join(projectsFolder, 'models'),
+export const cpuCliVersion = (
+  await loadPackageJson(join(projectsFolder, 'models'))
 ).version;
```
matejchalk, line 21: Should use `cp` instead of `cpu` - [#branding](https://github.com/code-pushup/portal/wiki/Branding) :grin:

matejchalk (commented):
Nice cleanup :slightly_smiling_face: Very happy this was converted from CJS to ESM :tada:

packages/nx-plugin/src/generators/init/generator.ts
Tlacenka: I had a feeling there was a way when I tried it in the past but TS refused to help me 😅 Good job

packages/nx-plugin/src/utils/versions.ts
Tlacenka: Good point, forgot about that one.

BioPhoton: used the helper

packages/nx-plugin/src/utils/versions.ts
BioPhoton: renamed

packages/nx-plugin/src/generators/init/generator.ts
BioPhoton: used the type

matejchalk: > used the type

I still see `type PkgJson = Record<string, Record<string, string>>` being used here. Did you push the change?

packages/nx-plugin/src/generators/init/generator.ts
```diff
@@ -11,41 +11,52 @@ import {
...
+      ...packageJson,
+    };
 
-    if (packageJson.dependencies['@code-pushup/nx-plugin']) {
-      packageJson.devDependencies['@code-pushup/nx-plugin'] =
-        packageJson.dependencies['@code-pushup/nx-plugin'];
-      delete packageJson.dependencies['@code-pushup/nx-plugin'];
+    if (newPackageJson.dependencies?.[nxPluginPackageName] !== undefined) {
+      const { [nxPluginPackageName]: version, ...dependencies } =
+        newPackageJson.dependencies as { [nxPluginPackageName]: string };
```
matejchalk, line 49: Type-cast is not needed and also not true (object doesn't only contain `{ '@code-pushup/nx-plugin': string }`).

```suggestion
        newPackageJson.dependencies;
```

packages/nx-plugin/src/generators/init/generator.ts
```diff
@@ -11,41 +11,52 @@ import {
...
+      return {
+        ...newPackageJson,
+        dependencies,
+        devDependencies: {
+          ...newPackageJson.devDependencies,
+          [nxPluginPackageName as string]: version,
```
matejchalk, line 55:
```suggestion
          [nxPluginPackageName]: version,
```

packages/nx-plugin/src/generators/init/generator.ts
BioPhoton: done

matejchalk (approved):

Issue #90, opened 2023-10-06
Implement Lighthouse plugin


vmasek:
> As a user,
> I want a plugin that allows me to track performance degradations of any given URL,
> so that I can track and focus on adoption perf bottlenecks and opportunities to improve them.

**Acceptance Criteria:**. 
- [x] The plugin must allow users to define a single URL in the plugin options.
- [ ] Run lighthouse against that URL and get audit results in a CP report.
- [x] The plugin accepts all `CLIFlags` as well as `Config` for the official Lighthouse project
- [x] The plugin respects the following flags in relation to the CP data validation:
  - handle audits and groups filtering for `--onlyAudits` and `--onlyCategories`

BioPhoton:
this is more of an epic an should get split

PR #472, opened 2024-02-06
feat(plugin-lighthouse): add onlyAudits logic

BioPhoton:
related to #90


packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -1,4 +1,4 @@
-import { expect } from 'vitest';
+import { describe, expect, it } from 'vitest';
```
BioPhoton, line 1:
```suggestion
import { expect } from 'vitest';
```

packages/plugin-lighthouse/src/lib/constants.ts
```diff
@@ -22,7 +22,7 @@ export const GROUPS: Group[] = Object.entries(categories ?? {}).map(
   }),
 );
 
-export const AUDITS: Audit[] = await Promise.all(
+export const AUDITS = await Promise.all(
```
BioPhoton, line 25:
```suggestion
export const AUDITS: Audit[] = await Promise.all(
```

packages/utils/src/lib/transform.ts
```diff
@@ -156,3 +157,56 @@ export function toOrdinal(value: number): string {
...
+  constructor(list: { slug: string }[], auditSlugs: string[]) {
+    super(
+      `audits: "${auditSlugs
+        .filter(slug => !list.some(a => a.slug === slug))
+        .join(', ')}" not implemented`,
+    );
+  }
+}
+
+export function filterByAuditSlug<
```
Tlacenka, line 171: This is hard for me to navigate. It includes a lot of one-letter types. Does the notation need to be so complicated? I am no TypeScript expert but I wouldn't know how to debug this in the future on my own.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
```diff
@@ -1,25 +1,31 @@
-import { AuditOutputs, PluginConfig } from '@code-pushup/models';
+import { Audit, AuditOutputs, PluginConfig } from '@code-pushup/models';
+import { filterByAuditSlug, filterBySlug } from '@code-pushup/utils';
 import { AUDITS, GROUPS, LIGHTHOUSE_PLUGIN_SLUG } from './constants';
 
+type AuditSlug = (typeof AUDITS)[number]['slug'];
```
Tlacenka, line 5: If it is necessary to check that the value is a slug, it would imo be better to use a slugSchema. Otherwise, a `string` should suffice or what am I missing?

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    );
+  });
+});
+
+describe('constants', () => {
+  it.each(AUDITS.map(a => [a.slug, a]))(
+    'should parse audit "%s" correctly',
+    (_, audit) => {
+      expect(() => auditSchema.parse(audit)).not.toThrow();
+      expect(audit.description).toEqual(expect.any(String));
```
Tlacenka, line 66: It would be more valuable to check the slug value which is already extracted as a parameter to the test instead of vague "any string" in description.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
```diff
@@ -1,25 +1,31 @@
...
 export type LighthousePluginOptions = {
   url: string;
   outputPath?: string;
-  onlyAudits?: string | string[];
+  onlyAudits?: AuditSlug | AuditSlug[];
```
matejchalk, line 9: This doesn't work, unfortunately, `AuditSlug` is the same as `string` here. The `meta.id` from Lighthouse is a `string`, so we're not going to get a more specific type than that.

packages/utils/src/lib/transform.ts
```diff
@@ -156,3 +157,56 @@ export function toOrdinal(value: number): string {
...
+
+        return {
+          ...group,
+          refs: groupsRefs,
+        };
+      }) as L
+  );
+}
+
+export function filterBySlug<T extends Audit, L extends T[]>(
```
Tlacenka, line 199: Same as `filterByAuditSlug`. Please use verbose names for both types and variables and reduce the complexity of the types if possible.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  it('should filter groups by onlyAudits string "first-contentful-paint"', () => {
+    const pluginConfig = lighthousePlugin({
+      url: 'https://code-pushup-portal.com',
+      onlyAudits: 'first-contentful-paint',
+    });
+
+    expect(() => pluginConfigSchema.parse(pluginConfig)).not.toThrow();
+    expect(pluginConfig.groups).toHaveLength(1);
+
+    const refs = (pluginConfig.groups?.[0] as Group).refs;
```
matejchalk, line 48:
```suggestion
    const refs = pluginConfig.groups?.[0]?.refs;
```

packages/utils/src/lib/transform.ts
matejchalk, line 1: Not sure `transform.ts` is the best place for this new code, as its more domain-specific. How about `filter-by-slug.ts`?

packages/utils/src/lib/transform.ts
Tlacenka, line 171: I think the common logic of these two functions could be extracted into a helper filter function which would accept just an array audit slugs and filter out those that aren't in `onlyAudits`. This function could then be called on mapped array that both `filterByAuditSlug` or `filterBySlug` can provide.

The unit tests would then test this core logic instead of duplicating use cases for each function separately.

Tlacenka (changes requested):

packages/utils/src/lib/transform.ts
```diff
@@ -156,3 +157,56 @@ export function toOrdinal(value: number): string {
...
+  const slugs = toArray(auditSlugs);
+  if (slugs.length === 0) {
+    return list;
+  }
+  if (slugs.some(slug => !list.some(wS => wS.slug === slug))) {
+    throw new AuditsNotImplementedError(list, slugs);
+  }
+
+  return list.filter(({ slug }) => slugs.includes(slug)) as L;
+}
```
matejchalk, line 212: As mentioned above, the audit slug isn't union but a plain old `string`. So there's no advantage to the generic type parameters here, it just adds complexity:

```suggestion
export function filterByAuditSlug(
  groups: Group[],
  auditSlugs: string | string[],
): Group[] {
  const slugs = toArray(auditSlugs);
  if (slugs.length === 0) {
    return groups;
  }
  return (
    groups
      // filter out groups that have no audits includes from onlyAudits (avoid empty groups)
      .filter(group => group.refs.some(({ slug }) => slugs.includes(slug)))
      .map(group => {
        const groupsRefs = group.refs.filter(({ slug }) =>
          slugs.includes(slug),
        );

        return {
          ...group,
          refs: groupsRefs,
        };
      })
  );
}

export function filterBySlug(
  list: Audit[],
  auditSlugs: string | string[],
): Audit[] {
  const slugs = toArray(auditSlugs);
  if (slugs.length === 0) {
    return list;
  }
  if (slugs.some(slug => !list.some(wS => wS.slug === slug))) {
    throw new AuditsNotImplementedError(list, slugs);
  }

  return list.filter(({ slug }) => slugs.includes(slug));
}
```

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
```diff
@@ -0,0 +1,76 @@
...
+import { AUDITS, GROUPS } from './constants';
+import { lighthousePlugin } from './lighthouse-plugin';
+
+describe('lighthousePlugin-config-object', () => {
+  it('should create valid plugin config', () => {
+    const pluginConfig = lighthousePlugin({
+      url: 'https://code-pushup-portal.com',
+    });
+    expect(() => pluginConfigSchema.parse(pluginConfig)).not.toThrow();
+    expect(pluginConfig.audits).toHaveLength(168);
```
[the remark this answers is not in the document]
BioPhoton: I did not want to make a snapshot test here but basically I want to know when they change audits...
Ill go for `toBeGreaterThan`

packages/utils/src/lib/transform.ts
matejchalk: I think know what Michael was trying to do, if the audit's `slug` field as a union type then this function could have nice autocomplete. But unfortunately the slugs are strings anyway, so I believe the generics are unnecessary, as mentioned here: https://github.com/code-pushup/cli/pull/472#discussion_r1480144213

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
matejchalk: Failed attempt at inferring a union, I believe: https://github.com/code-pushup/cli/pull/472#discussion_r1480107285

packages/utils/src/lib/transform.ts
matejchalk: Not relevant anymore I believe, as mentioned elsewhere. But in general a good point, when there are multiple type parameters, then `TAudit`, `TList` etc. is often a better naming convention.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: done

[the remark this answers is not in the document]
matejchalk: Yeah, the Lighthouse audits don't change too often, I believe. And we'll only encounter this in updates, which I think is manageable.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
BioPhoton: very good point!

packages/utils/src/lib/filter-by-slug.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+  return filterSlug(list, slugs);
+}
+
+export function filterSlug<T extends { slug: string }>(
+  refs: T[],
+  slugs: string | string[],
+): T[] {
+  return refs.filter(({ slug }) => slugs.includes(slug));
+}
```
matejchalk, line 39: :lady_beetle:  Careful, `slugs` is a single string and not an array, then `includes` will match substrings, which is not what you want.

```suggestion
export function filterSlug<T extends { slug: string }>(
  refs: T[],
  slugOrSlugs: string | string[],
): T[] {
  const slugs = toArray(slugOrSlugs);
  return refs.filter(({ slug }) => slugs.includes(slug));
}
```

or

```suggestion
export function filterSlug<T extends { slug: string }>(
  refs: T[],
  slugOrSlugs: string | string[],
): T[] {
  return refs.filter(({ slug }) =>
    Array.isArray(slugOrSlugs) ? slugOrSlugs.includes(slug) : slug === slugOrSlugs,
  );
}
```

packages/plugin-lighthouse/src/lib/utils.ts
```diff
@@ -48,3 +49,22 @@ export function getLighthouseCliArguments(
...
+    super(`audits: "${auditSlugs.join(', ')}" not implemented`);
+  }
+}
+
+export function validateOnlyAudits(
+  audits: Audit[],
+  onlyAudits: string | string[],
+): audits is Audit[] {
+  const missingAudtis = toArray(onlyAudits).filter(
+    slug => !audits.some(wS => wS.slug === slug),
```
matejchalk, line 64: No clue what `wS` means.

```suggestion
    slug => !audits.some(audit => audit.slug === slug),
```

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
```diff
@@ -1,25 +1,35 @@
...
+import { filterByAuditSlug, filterBySlug } from '@code-pushup/utils';
 import { AUDITS, GROUPS, LIGHTHOUSE_PLUGIN_SLUG } from './constants';
+import { validateOnlyAudits } from './utils';
 
+type AuditSlug = (typeof AUDITS)[number]['slug'];
 export type LighthousePluginOptions = {
   url: string;
   outputPath?: string;
-  onlyAudits?: string | string[];
+  onlyAudits?: AuditSlug | AuditSlug[];
```
matejchalk, line 10: As commented before, the `AuditSlug` is an over-complicated way of inferring a `string`. Just use `string` in the public API, you'll confuse the user less.

image

```suggestion
  onlyAudits?: string | string[];
```

matejchalk (changes requested):

packages/utils/src/lib/filter-by-slug.unit.test.ts
Tlacenka, line 1: The test cases should not be identical anymore as `filterSlug` (audits by slug) has the main logic in it.
Any functions calling this function, such as the group filtering function, should only have a happy path tested or use cases only relevant for that caller function (e.g. passing more than one group).

packages/utils/src/lib/filter-by-slug.ts
```diff
@@ -0,0 +1,39 @@
+import { Audit, Group } from '@code-pushup/models';
+import { toArray } from './transform';
+
+export function filterByAuditSlug(
```
Tlacenka, line 4: It would be helpful to distinguish the name of these functions based on what the arrays contain.
```suggestion
export function filterGroupsByAuditSlug(
```

packages/utils/src/lib/filter-by-slug.ts
```diff
@@ -0,0 +1,39 @@
...
+      .map(group => ({
+        ...group,
+        refs: filterSlug(group.refs, slugs),
+      }))
+      // filter out groups that have no audits includes from onlyAudits (avoid empty groups)
+      .filter(group => group.refs.length)
+  );
+}
+
+export function filterBySlug(
```
Tlacenka, line 23: It would be helpful to distinguish the name of these functions based on what the arrays contain.
```suggestion
export function filterAuditsBySlug(
```

matejchalk (approved):

packages/utils/src/lib/filter-by-slug.unit.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    ];
+    expect(filterGroupsByAuditSlug(list, ['a'])).toEqual([
+      {
+        slug: 'g',
+        title: 'G',
+        refs: [{ slug: 'a', weight: 1 }],
+      },
+    ]);
+  });
+});
```
Tlacenka, line 129: Please remove the duplicate test cases. A happy path is enough.

Tlacenka: A relevant test for `filterGroupsByAuditSlug` was also suggested here
https://github.com/code-pushup/cli/pull/472#discussion_r1481156674

Tlacenka (approved):

PR #475, opened 2024-02-06
refactor: add lint rules to cli and core

BioPhoton:
dogfooding

unblocked by #516


packages/cli/src/lib/autorun/autorun-command.ts
```diff
@@ -48,15 +48,15 @@ export function yargsAutorunCommandObject() {
         renderConfigureCategoriesHint();
       }
 
-      if (options.upload) {
+      if (args['upload']) {
```
BioPhoton, line 51:
```suggestion
      if (!args['upload']) {
```

packages/core/src/lib/implementation/read-rc-file.unit.test.ts
```diff
@@ -14,7 +14,7 @@ vi.mock('bundle-require', async () => {
       .fn()
       .mockImplementation((options: { filepath: string }) => {
         const extension =
-          options.filepath?.split('.').at(-1) || 'no-extension-found';
+          options.filepath.split('.').at(-1) || 'no-extension-found';
```
Tlacenka, line 17: By looking at the implementation, this should never happen. The function always provides a non-empty extension to `readRcByPath` → `importEsmModule` → `bundle-require`. If the extension is not found, the function throws.

```suggestion
          options.filepath.split('.').at(-1);
```

packages/core/src/lib/implementation/report-to-gql.ts
```diff
@@ -69,6 +69,7 @@ function auditToGQL(audit: AuditReport): PortalAudit {
     formattedValue: audit.displayValue,
     ...(audit.details && {
       details: {
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
         ...(audit.details.issues && {
           issues: audit.details.issues.map(issueToGQL),
         }),
```
Tlacenka, line 75: While the `details` schema is [optional](https://github.com/code-pushup/cli/blob/main/packages/models/src/lib/audit-output.ts#L27), the `issues` property [is not](https://github.com/code-pushup/cli/blob/main/packages/models/src/lib/audit-output.ts#L8). Therefore, the optional object insertion should be removed and ESLint correctly assessed it as an unnecessary condition. Or am I missing something?

```suggestion
          issues: audit.details.issues.map(issueToGQL)
```

package.json
```diff
@@ -84,6 +84,7 @@
     "@types/benchmark": "^2.1.4",
     "@types/chalk": "^2.2.0",
     "@types/eslint": "^8.44.2",
+    "@types/globby": "^9.1.0",
```
BioPhoton, line 87:
```suggestion
```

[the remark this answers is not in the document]
BioPhoton: removed

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@code-pushup/cli-source",
-  "version": "0.22.8",
+  "version": "0.23.0",
```
BioPhoton, line 3:
```suggestion
  "version": "0.22.8",
```

Tlacenka (approved):
Finally, all our main packages have all rules enabled 🎉 🍻

PR #486, opened 2024-02-07
ci: update actions versions and Node version

### LINKED ISSUE #447: Run CI on node version 20
### User story

As a contributor, I wish to have the pipelines run on a supported version of node. Currently they run on node version 18.

### Acceptance criteria

- [ ] Update all pipelines to run on node version 20.
- [ ] All CLI functionality still works as expected.

### Implementation details

image

matejchalk:
Closes #447 

- updated to [LTS Node version](https://nodejs.org/en) - 20.11.0
  - used `node-version-file: .nvmrc` so that it's easier to change the version in one place 
- updated all 3rd party actions

Impact:
- [previous warnings - 41](https://github.com/code-pushup/cli/actions/runs/7816840860)
- [current warnings - 36](https://github.com/code-pushup/cli/actions/runs/7817234791)

There are still a bunch warnings from [Wandalen/wretry.action@master](https://github.com/Wandalen/wretry.action/tree/master/):

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: Wandalen/wretry.action@v1.3.0_js_action. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

The repo has an open issue: https://github.com/Wandalen/wretry.action/issues/126

> Unexpected input(s) 'time_out', valid inputs are ['action', 'command', 'with', 'current_path', 'env_context', 'github_context', 'job_context', 'matrix_context', 'attempt_limit', 'attempt_delay']

A ton of these warnings in testing jobs. We don't set a `time_out` property, so I think this is some internal problem of the retry action.

:top: Do we still need to use the retry action?

----

**Edit**: Removed retry from unit and integration tests. [Latest run has no warnings](https://github.com/code-pushup/cli/actions/runs/7828135953?pr=486) :astonished:


BioPhoton (dismissed):
Nice version update!

Tlacenka (dismissed):
> Do we still need to use the retry action?

Short answer: For E2E yes, in my opinion. Other suites can drop the retry.
The reason for the retry action is that:
1. Ubuntu and Windows tests take very different times to complete. Windows would sometimes time out. I did not want to set a high timeout to not leave Ubuntu error hanging for so long.
2. Verdaccio is making the E2E tests run even longer, making the baseline timeout even higher than it would normally need to be.

As for the timeout itself:
- All tests should have internal `testTimeout` property set either as a default or overriden in tests/configs.
- I would say that setting a different timeout for local vs pipeline test run is not a good idea as we would need to maintain both values. But if we were to set one, I would create a global constant for this that can be used across workflows.

Tlacenka (approved):

Issue #498, opened 2024-02-14
Support globs


Tlacenka:
As a CP user, I want to pass coverage results as a glob, e.g. `coverage/**/lcov.info`.

## Acceptance criteria
- [ ] Add a format that has absolute paths (e.g. Cobertura) and therefore allows using globs without issue links problems.
- [ ] Allow passing results paths within a glob.

Issue #430, opened 2024-01-23
Code coverage plugin


Tlacenka:
As a Code PushUp user, I want to be able to see my code coverage results in the Code PushUp report.

## Acceptance criteria
- [x] Code coverage plugin is implemented.
- [x] The plugin has sufficient documentation of its interface and usage.
- [x] The plugin is covered by tests.
- [x] Its audits as well as issues can be checked in portal.
- [x] Preset for unit tests is added to our `code-pushup.config.ts` for dogfooding.
- [x] Integration tests are added to our `code-pushup.config.ts` via nx helper for dogfooding.
- [x] The plugin is published and referenced in our README.
- [x] The plugin is fully integrated to all CLI commands (`project.json`)

## Implementation details
- [x] One can choose whether to only provide code coverage results or run a specified command first to collect results.
- [x] Paths to coverage results are provided either as an array (simple project) or deduced via nx helpers (uncovering path to project within workspace).
- [x] E2E should be implemented to test running a coverage command tool on our example repo.
- [x] Both integration and unit test results are used for dogfooding.
- [x] Audits are provided as a group for convenience.
- [x] A unified format is chosen for the expected results (currently between XML cobertura and LCOV).
- [x] Audit is a type of coverage (function, branch, condition, statement/line) for the whole project.
- [x] Issue is a finding of missing coverage (table can be filtered by folder for issues within one package).

## Additional post-MVP features
#498 

Detailed research available on demand.

PR #499, opened 2024-02-14
feat(plugin-coverage): allow passing results as strings

Tlacenka:
Related to #430 
In this PR, I allow passing result paths as strings.
However, this approach **can be only used for a single project setup** atm as LCOV needs to prepend path to root in its records with relative paths.

Once another format is supported, the full potential along with globs will be unlocked. Created #498 for this.

### DIFF
--- a/packages/plugin-coverage/README.md
+++ b/packages/plugin-coverage/README.md
@@ -36,7 +36,7 @@ Measured coverage types are mapped to Code PushUp audits in the following way
      plugins: [
        // ...
        await coveragePlugin({
-         reports: [{ resultsPath: 'coverage/lcov.info' }],
+         reports: ['coverage/lcov.info'],
          coverageToolCommand: {
            command: 'npx',
            args: ['jest', '--coverage', '--coverageReporters=lcov'],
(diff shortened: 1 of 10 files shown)


packages/plugin-coverage/README.md
```diff
@@ -119,8 +119,9 @@ It recognises the following entities:
 The plugin accepts the following parameters:
 
 - `coverageTypes`: An array of types of coverage that you wish to track. Supported values: `function`, `branch`, `line`. Defaults to all available types.
-- `reports`: Array of information about files with code coverage results - paths to results, path to project root the results belong to. LCOV format is supported for now.
-  - If you have an Nx monorepo, you can adjust our helper function `getNxCoveragePaths` to get the path information automatically.
+- `reports`: Array of information about files with code coverage results. LCOV format is supported for now.
+  - For a single project, providing paths to results as strings is enough.
+  - If you have an Nx monorepo, both path to results (`resultsPath`) and path from the root to project the results belong to (`pathToProject`) need to be provided for the LCOV format. You can adjust our helper function `getNxCoveragePaths` to get the path information automatically.
```
matejchalk, line 124:
```suggestion
  - If you have a monorepo, both path to results (`resultsPath`) and path from the root to project the results belong to (`pathToProject`) need to be provided for the LCOV format. For Nx monorepos, you can use our helper function `getNxCoveragePaths` to get the path information automatically.
```

matejchalk (dismissed):
Nice improvement :+1:

matejchalk (approved):

PR #503, opened 2024-02-14
feat(core): make all issue file paths relative to git root after executing plugin

### LINKED ISSUE #465: Fix issue file paths when run in non-root directory
When Code PushUp is run from a nested folder, the source file paths in reported issues aren't relative to root and therefore snippets and links to git repo are broken in portal.

### Example

This project has an Nx monorepo, but it all lives in a `ui` folder. So although an issue is reported for `apps/front-office/src/app/app.component.ts`, the actual path from root should be `ui/apps/front-office/src/app/app.component.ts`.

Image

### Possible solution

Extend `toUnixPath` helper (used by ESLint and Coverage plugins to format file paths in issues) to transform path so that it's relative to git repository root.

matejchalk:
Fixes #465 

Removed all usages of `toUnixPath` in plugins. Since it would be inconvenient to use the new helper function in each plugin - it's now async and we don't want to run the Git command for each issue - I transferred that responsibility to `core`. Created testing utils for making paths OS-agnostic for snapshot tests. And refactored Git tests to run against a temporary git repo (I recommend applying this approach to #469).

### DIFF
--- a/e2e/cli-e2e/tests/__snapshots__/collect.e2e.test.ts.snap
+++ b/e2e/cli-e2e/tests/__snapshots__/collect.e2e.test.ts.snap
@@ -469,7 +469,7 @@ exports[`CLI collect > should run ESLint plugin and create report.json 1`] = `
                 "message": "'loading' is assigned a value but never used.",
                 "severity": "warning",
                 "source": {
-                  "file": "src/App.jsx",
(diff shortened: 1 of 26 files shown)


Tlacenka (approved):
Great!

Checked both the code and the issue links on staging, all seems to be in order.

PR #441, opened 2024-01-27
ci(models): add generated models to docs

BioPhoton:
closes #388

Issue #388, opened 2024-01-03
Generate TS docs from Zod schemas


matejchalk:
Follow-up to research ticket:
- #242 

Decision is to generate the TypeScript code, since it's most readable.

- [ ] add script to automatically generate config docs using [zod-to-ts](https://www.npmjs.com/package/zod-to-ts)
- [ ] should be accessible from Markdown docs
- [ ] run docs script automatically so devs don't have to regenerate docs manually (git hook / CI step / on release?)

Issue #242, opened 2023-11-13
Generate config docs from Zod schemas


kristihoskova:
Spike for the research is 2h.

kristihoskova:
Spike for research 2h.

matejchalk:
The Python generator looks best (most compact), but requires Python and Pip installation to run. 

There are still some common problems in how the generated docs look:
- all the objects and arrays are defined inline, as it were, our object names (e.g. `PluginConfig`) don't make it in, only the descriptions
- `z.union`s used in `docsUrl` and `runner` aren't handled very well, the runner function gets lost completely because it has no JSON representation
- the material icon enum is very large, but unfortunately it gets inlined (could probably hack in a special case to separate it though)

The alternatives to using a 3rd party tool are:
- document them manually - not great for long-term maintenance (duplicated structure and descriptions)
- implement our own docs generator - requires a lot of implementation effort
The main advantage to doing it ourself is we can make it look just right for what we need.

matejchalk:
Came up with an alternative way to document the config automatically - generate equivalent TypeScript instead of Markdown via JSON Schema - has the advantage of being able to represent what JSON can't (e.g. functions):

----

<details>
<summary>TypeScript definition generated by <a href="https://www.npmjs.com/package/zod-to-ts">zod-to-ts</a></summary>
```ts
/** Code PushUp configuration */
type CoreConfig = {
    /** List of plugins to be used (official, community-provided, or custom) */
    plugins: {
        /** NPM package name */
        packageName?: string | undefined;
        /** NPM version of the package */
        version?: string | undefined;
        /** Descriptive name */
        title: string;
        /** Description (markdown) */
        description?: string | undefined;
        /** Plugin documentation site */
        docsUrl?: (string | undefined) | string;
        /** References plugin. ID (unique within core config) */
        slug: string;
        /** Icon from VSCode Material Icons extension */
[...]

PR #504, opened 2024-02-15
docs(models): generate docs for exported Zod schemas with zod2md

matejchalk:
Follow-up improvements to #441, #388, #242.

I created a [zod2md](https://www.npmjs.com/package/zod2md) package in my free time. Integrated it here instead of `zod-to-ts` since the docs are a lot more readable and models can be referenced individually.

Also cleaned up exports from `models` package to optimize the public interface.

- [before](https://github.com/code-pushup/cli/blob/484fde19497dc4ec677915b93670fc8f900f4e6e/packages/models/docs/core-config-types.md)
- [after](https://github.com/code-pushup/cli/blob/f778e0108d7dda601a0383db0db7a9513a131247/packages/models/docs/models-reference.md)


BioPhoton (commented):
Wow!! it looks so much better!

Thanks a lot to work on that as a side project. 🙏

PR #509, opened 2024-02-16
feat(cli): add --tsconfig option

### LINKED ISSUE #473: Add --tsconfig CLI argument for TypeScript config file
### User story

As a user who has implemented my own Code PushUp plugin in my Nx monorepo, I want to be able to use TS path aliases (`import { ... } from '@scope/my-lib'`) to import code from Nx projects in the usual way. Currently, my path aliases aren't recognized because the Code PushUp CLI doesn't use my `tsconfig.base.json` where the `paths` are configured.

### Acceptance criteria

- [x] Add an optional `--tsconfig` CLI argument which accepts a file path.
- [x] Pass  tsconfig path to `bundle-require` when importing config file.
- [x] Add docs on using `code-pushup.config.ts` file with custom TS project. 

### Implementation details

The `tsconfig` option will only be a CLI option, there's no sense to having it in the config file as well (chicken-and-egg problem).

matejchalk:
Closes #473

### DIFF
--- a/e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
+++ b/e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
@@ -16,9 +16,11 @@ Options:
                                                        [boolean] [default: true]
       --verbose              When true creates more verbose output. This is help
                              ful when debugging.      [boolean] [default: false]
-      --config               Path the the config file, e.g. code-pushup.config.t
-                             s. By default it loads code-pushup.config.(ts|mjs|j
-                             s).                                        [string]
+      --config               Path to config file, e.g. code-pushup.config.ts. By
+                              default it loads code-pushup.config.(ts|mjs|js).
+                                                                        [string]
+      --tsconfig             Path to a TypeScript config, to be used when loadin
+                             g config file.                             [string]
       --persist.outputDir    Directory for the produced reports         [string]
       --persist.filename     Filename for the produced reports.         [string]
       --persist.format       Format of the report output. e.g. \`md\`, \`json\`
--- a/packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
+++ b/packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
@@ -35,4 +35,21 @@ describe('coreConfigMiddleware', () => {
       coreConfigMiddleware({ config: 'wrong/path/to/config' }),
     ).rejects.toThrow(/Provided path .* is not valid./);
   });
+
+  it('should load config which relies on provided --tsconfig', async () => {
+    await expect(
+      coreConfigMiddleware({
+        config: join(configDirPath, 'code-pushup.needs-tsconfig.config.ts'),
+        tsconfig: 'tsconfig.base.json',
+      }),
+    ).resolves.toBeTruthy();
+  });
+
+  it('should throw if --tsconfig is missing but needed to resolve import', async () => {
+    await expect(
+      coreConfigMiddleware({
+        config: join(configDirPath, 'code-pushup.needs-tsconfig.config.ts'),
+      }),
+    ).rejects.toThrow("Cannot find package '@code-pushup/models'");
+  });
 });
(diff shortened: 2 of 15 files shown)


packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
```diff
@@ -35,4 +35,21 @@ describe('coreConfigMiddleware', () => {
...
+  it('should load config which relies on provided --tsconfig', async () => {
+    await expect(
+      coreConfigMiddleware({
+        config: join(configDirPath, 'code-pushup.needs-tsconfig.config.ts'),
+        tsconfig: 'tsconfig.base.json',
+      }),
+    ).resolves.toBeTruthy();
+  });
+
+  it('should throw if --tsconfig is missing but needed', async () => {
```
BioPhoton, line 48: But needed for what? Maybe obvious... so not sure if there is more info needed.

packages/cli/src/lib/implementation/global.options.ts
```diff
@@ -19,7 +19,12 @@ export function yargsGlobalOptionsDefinition(): Record<
     },
     config: {
       describe:
-        'Path the the config file, e.g. code-pushup.config.ts. By default it loads code-pushup.config.(ts|mjs|js).',
+        'Path to config file, e.g. code-pushup.config.ts. By default it loads code-pushup.config.(ts|mjs|js).',
+      type: 'string',
+    },
+    tsconfig: {
+      describe:
+        'Path to a TypeScript project, to be used when loading config file.',
```
BioPhoton, line 27: isn't it TypeScript config? or is it considered a project naming wise?

BioPhoton (commented):
Added some questions

packages/cli/src/lib/implementation/core-config.middleware.integration.test.ts
matejchalk: Adjusted the test description. There's more info in the `code-pushup.needs-tsconfig.config.ts` file.

packages/cli/src/lib/implementation/global.options.ts
matejchalk: Good point, [according to the docs](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html), the `tsconfig.json` file defines a TypeScript _project_, but itself is a _config_. 

I've changed the descriptions to use _config_ instead of _project_.

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
```diff
@@ -16,9 +16,11 @@ Options:
                                                        [boolean] [default: true]
       --verbose              When true creates more verbose output. This is help
                              ful when debugging.      [boolean] [default: false]
-      --config               Path the the config file, e.g. code-pushup.config.t
-                             s. By default it loads code-pushup.config.(ts|mjs|j
-                             s).                                        [string]
+      --config               Path to config file, e.g. code-pushup.config.ts. By
+                              default it loads code-pushup.config.(ts|mjs|js).
+                                                                        [string]
+      --tsconfig             Path to a TypeScript config, to be used when loadin
```
Tlacenka, line 22: Can you add this option to `print-config` E2E tests or a `yargs-cli` integration test to check that passing this argument is correctly registered by the CLI?

packages/core/src/lib/implementation/read-rc-file.integration.test.ts
```diff
@@ -37,6 +37,15 @@ describe('readRcByPath', () => {
     );
   });
 
+  it('should load the configuration using provided tsconfig', async () => {
+    await expect(
+      readRcByPath(
+        join(configDirPath, 'code-pushup.needs-tsconfig.config.ts'),
+        'tsconfig.base.json',
+      ),
+    ).resolves.toBeTruthy();
```
Tlacenka, line 46: This might still return empty places where the models have been used. I think it would be valuable to check the properties where the package was used (so `audits` and `runner`).

Tlacenka (commented):
Looking promising!

I went through the code and tests, mentioned a few minor test improvements.

Tlacenka (approved):
Nice!

Issue #488, opened 2024-02-09
Clean up example lighthouse leftover file


Tlacenka:
Currently, an integration test in example lighthouse plugin creates and leaves a new file in the root folder - `lighthouse-report.json`.
This file should not persist after the test has finished.

Issue #493, opened 2024-02-12
Investigate broken builds


matejchalk:
E2E tests have a tendency to fail with `Cannot use import statement outside a module`, e.g. here: https://github.com/code-pushup/cli/actions/runs/7855491262/job/21437256994

![Image](https://github.com/code-pushup/cli/assets/34691111/ef182132-998c-4388-8a3e-25179482c5cf)

The cause seems to be that `"type": "module"` isn't set in `package.json` build output. The is supposed to be handled by [custom plugin in `esbuild.config.js`](https://github.com/code-pushup/cli/blob/main/esbuild.config.js#L31-L81) and it works most of the time, but for some reason not always. Could perhaps be problematic when builds are in parallel? 😕 

An even worse impact is that sometimes this results in publishing broken builds to NPM - see #394. This could potentially be prevented by adding some post-build checks that would fail before releasing. But the main thing is to figure out the root cause.

PR #512, opened 2024-02-17
test(examples-plugins): convert integration tests to e2e

### LINKED ISSUE #490: Reduce lighthouse integration test duration, fix onlyAudits
Currently, the Lighthouse example integration tests take above 20 seconds each locally. This slows up the pipelines and requires the timeouts to be manually increased. As a hotfix, these tests have been temporarily ignored by using `it.skip`.

Additionally, it is not possible to run `onlyAudits` option with any other audit than `largest-contentful-paint`. Attempting to run the example plugin with any other slug results in an error.

image

## Acceptance criteria
- [ ] Fix `onlyAudits` to accept any existing audit, not only the first one.
- [ ] Verify that the tests do not take longer than 20 seconds each **in any environment** by running them multiple times in the pipelines.
   - Either the test scope may be reduced by running only one audit at a time.
   - If that does not help, convert the tests to E2E tests.

BioPhoton:
I did not change the logic of the example plugin but moved the long running tests to e2e.
I think we should put our time in the official lighthouse plugin.

closes #490
followed by #488


examples/plugins/project.json
Tlacenka, line 1: There is a dedicated folder `e2e` for E2E test projects.
In order to keep our test location consistent (see [testing strategy](https://github.com/code-pushup/cli/wiki/Testing-Strategy#end-to-end-tests), these E2E should be located in e.g. `e2e/example-plugins-e2e` project.

examples/plugins/src/lighthouse/src/lighthouse.plugin.e2e.test.ts
```diff
@@ -0,0 +1,41 @@
...
+        slug,
+        title: 'Lighthouse',
+        description: 'Chrome lighthouse CLI as code-pushup plugin',
+        duration: expect.any(Number),
+        date: expect.any(String),
+        audits: expect.any(Array),
+        groups: expect.any(Array),
+      }),
+    );
+  });
```
Tlacenka, line 25: In order to keep the E2E execution consistent, one should only interact via the CLI using `onlyPlugins` and provide a code pushup config. See ESLint E2E test [here](https://github.com/code-pushup/cli/blob/main/e2e/cli-e2e/tests/collect.e2e.test.ts#L30-L35).

examples/plugins/src/lighthouse/src/lighthouse.plugin.e2e.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  it('should use onlyAudits', async () => {
+    const pluginConfig = await create({
+      url: LIGHTHOUSE_URL,
+      outputPath: `${join('tmp', 'lighthouse-report.json')}`,
+      onlyAudits: 'largest-contentful-paint',
+    });
+    expect(pluginConfig.runner.args).toEqual(
+      expect.arrayContaining(['--onlyAudits="largest-contentful-paint"']),
+    );
```
Tlacenka, line 35: If the `onlyAudits` argument is not covered elsewhere, the `create` part + assertion can be added into the integration tests (without the `executePlugin` part).

examples/plugins/src/lighthouse/src/lighthouse.plugin.e2e.test.ts
Tlacenka, line 1: When I tried to run the E2E tests locally (also checked pipeline run) using `npx nx run examples-plugins:e2e`, I got "no tests found" message.
``` >  NX   Successfully ran target publish for 6 projects and 7 tasks they depend on

   Nx read the output from the cache instead of running the command for 5 out of 13 tasks.

(node:2508) [DEP0106] DeprecationWarning: crypto.createDecipher is deprecated.
(Use `node --trace-deprecation ...` to show where the warning was created)

 RUN  v0.32.4 /home/runner/work/cli/cli/examples/plugins

include: tests/**/*.e2e.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}
exclude:  **/node_modules/**, **/dist/**, **/cypress/**, **/.{idea,git,cache,output,temp}/**, **/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build}.config.*
watch exclude:  **/node_modules/**, **/dist/**
No test files found, exiting with code 0
```

Tlacenka (changes requested):
Good effort!

Unfortunately, the tests don't work for me. Do they work as expected when you run them locally?

examples/plugins/project.json
BioPhoton: I removed the execution as onlyAudits will throw even before execution.
also if the pluginconfig is not correct it will not execute at all.

examples/plugins/src/lighthouse/src/lighthouse.plugin.e2e.test.ts
BioPhoton: see above i removed the execution

examples/plugins/src/lighthouse/src/lighthouse.plugin.e2e.test.ts
BioPhoton: no need for e2e anymore. see above

examples/plugins/src/lighthouse/src/lighthouse.plugin.e2e.test.ts
BioPhoton: done

.github/workflows/ci.yml
```diff
@@ -130,7 +130,7 @@ jobs:
       - name: Install dependencies
         run: npm ci
       - name: E2E test affected projects
-        run: npx nx affected:e2e --parallel=3
+        run: npx nx affected:e2e --parallel=1
```
BioPhoton, line 133: @matejchalk I had to run the e2e tests in parallel to avoid a bug while publishing

examples/plugins/project.json
```diff
@@ -72,5 +72,5 @@
       ]
     }
   },
-  "tags": ["scope:internal", "type:feature"]
+  "tags": ["scope:internal", "scope:core", "type:feature"]
```
BioPhoton, line 75: I had to add that to be able to use `executePlugin` from the `core` package in the e2e tests. This helps me to test the plugin logic without execution the cli.

.github/workflows/ci.yml
Tlacenka: What was the bug? Was it related to all E2E or just the Lighthouse E2E? It would be very inconvenient to lose the parallel execution of E2E tests.

examples/plugins/project.json
Tlacenka: I am not sure I understand. Is there a bug in `onlyAudits`?

Tlacenka (changes requested):
This seems like the tests live in E2E folder but they are not E2E tests.
They also now require all E2E tests to run sequentially.

As this is an example plugin and not an official one, the cons outweigh the pros for me. So I am leaning towards skipping/removing the tests until such time when the lighthouse functionality can be executed via CLI and become a proper E2E test.

examples/plugins/project.json
BioPhoton: No but this is covered in unit tests (at least should be covered there)

.github/workflows/ci.yml
BioPhoton: I assume I have a clue now. The plugins are not executed in sequence. 

Will get back with a test case...

BioPhoton: Here a screenshot of console.logs... 

I putted logs inside the runner function (which in any case should execute sequential)

In the screen shot you see that first all runner log, then the content of the first runner logs.

Screenshot 2024-02-28 at 14 34 53

examples/plugins/project.json
BioPhoton: Otherwise it says:  
`4:1  error  A project tagged with "scope:core" can only depend on libs tagged with "scope:core", "scope:shared"  @nx/enforce-module-boundaries
`

Here graph:
Screenshot 2024-02-28 at 15 14 30

The reason:
I use `executePlugin` to e2e test the example lighthouse plugin in `e2e/examples-plugins-e2e`. This is the dependency that causes the issue.

matejchalk: > Otherwise it says:
> ```4:1  error  A project tagged with "scope:core" can only depend on libs tagged with "scope:core", "scope:shared"  @nx/enforce-module-boundaries ```

What file is that?

Tlacenka: An E2E test only accesses the application via regular user input into the CLI.
Therefore, this resulting test cannot by definition be an E2E test (a user cannot just run `executePlugin` via our CLI) but an integration test of the plugin with `core`.

So instead of going down this rabbit hole, we should skip this test until such time that it can be converted into a valid E2E test that is compatible with our testing strategy. The tested logic is an example plugin, not a critical path through the system so it is not a big deal to test it once the logic is ready in an official plugin.

.github/workflows/ci.yml
matejchalk: > I had to run the e2e tests in parallel to avoid a bug while publishing. Is this related to the bug you solved recently?

No clue what your bug is, but the bug I fixed was about `Cannot use import statement outside a module` errors in #493.

examples/plugins/project.json
BioPhoton: @Tlacenka 
I had so far a different understanding of e2e... it tests the logic end to end. Like in real life.
My test did that. The plugin is executed in the same way it would be in real life. I can delete the test, or I could just add couple of seconds for this edge case... I am happy if we just delete it too.

@matejchalk 
> I use executePlugin to e2e test the example lighthouse plugin in e2e/examples-plugins-e2e. This is the dependency that causes the issue.

In `lighthouse.e2e.test` I use `executePlugin`.

BioPhoton: Update: 

I deleted the e2e tests. If we need them I'm happy to open a new PR

Tlacenka: > I had so far a different understanding of e2e... it tests the logic end to end. Like in real life.
> My test did that. The plugin is executed in the same way it would be in real life.

Using your logic, any test would be an E2E test, even just calling one function that would normally be a unit test because that is how the function would be called in real life (all tests should test important scenarios for a given unit of code).

An E2E test should test things start to finish as it would be in real life, yes. Our E2E tests are black-box, meaning they should **only use tools at real user's disposal**. A real user will only be able to **install and run the CLI package** and pass arguments to it. They will not run `executePlugin` or any `core` functions on demand (`collect` function in `core` is debatable but no further than that).

BioPhoton: As tests are removed and scopes back to unchanged i close this here

examples/plugins/src/lighthouse/src/lighthouse.plugin.integration.test.ts
```diff
@@ -89,49 +85,19 @@ describe('lighthouse-create-export-config', () => {
...
     });
     expect(pluginConfig.runner.args).toEqual(
       expect.arrayContaining(['--onlyAudits="largest-contentful-paint"']),
     );
-    const { audits: auditOutputs } = await executePlugin(pluginConfig);
-
-    expect(auditOutputs).toHaveLength(1);
-    expect(auditOutputs[0]?.slug).toBe('largest-contentful-paint');
+    expect(pluginConfig.audits).toHaveLength(1);
+    expect(pluginConfig.audits[0]?.slug).toBe('largest-contentful-paint');
```
Tlacenka, line 99: Not formatted correctly. But using `objectContaining`, one can test it in one command 😁 
```suggestion
    expect(pluginConfig).toStrictEqual([expect.objectContaining({
      audits: [expect.object.Containing({ slug: 'largest-contentful-paint'})]
    })])
```

examples/plugins/src/lighthouse/src/lighthouse.plugin.integration.test.ts
```diff
@@ -37,7 +33,7 @@ describe('lighthouse-create-export-config', () => {
       'https://code-pushup.com',
       '--no-verbose',
       '--output="json"',
-      '--output-path="lighthouse-report.json"',
+      `--output-path="lighthouse-report.json"`,
```
Tlacenka, line 36:
```suggestion
      '--output-path="lighthouse-report.json"',
```

Tlacenka (approved):

examples/plugins/src/lighthouse/src/lighthouse.plugin.integration.test.ts
Tlacenka: Why keep backticks?

Tlacenka (approved):

matejchalk (approved):

PR #515, opened 2024-02-18
feat(plugin-lighthouse): add only filters logic for categories

BioPhoton:
lighthouse maintains a couple of CLI flags to configure the execution over process arguments. The same options are also implemented in their core logic.

This PR includes the handling of the following flags:
- `--onlyCategories` 
- partially integrates with the already existing logic for `--onlyAudits`

**Changes to review:**
- replacing utils `filterAuditsBySlug` and `filterGroupsByAuditSlug` with a more general implementation and naming `filterBy` and `filterItemRefsBy`
- align `lighthousePlugin` options with the official API (url is not in config object)
- encapsulate the linked logic for filtering audits and categories by the provided flags `--onlyAudits` and `--onlyCategories` within `filterAuditsAndGroupsByOnlyOptions`.
- added unit tests


packages/plugin-lighthouse/src/lib/utils.ts
```diff
@@ -68,3 +73,64 @@ export function validateOnlyAudits(
...
+  const { onlyAudits, onlyCategories } = options ?? {};
+
+  // category wins over audits
+  if (onlyCategories && onlyCategories.length > 0) {
+    validateOnlyCategories(groups, onlyCategories);
+
+    const categorieSlugs = new Set(onlyCategories);
+    const filteredGroups: Group[] = filterBy<Group>(groups, ({ slug }) =>
+      categorieSlugs.has(slug),
+    );
```
matejchalk, line 113: The `Group` type is inferred from the `groups` parameter:

```suggestion
    const filteredGroups = filterBy(groups, ({ slug }) =>
      categorieSlugs.has(slug),
    );
```

packages/utils/src/lib/log-results.unit.test.ts
```diff
@@ -6,7 +6,7 @@ describe('logMultipleResults', () => {
   const succeededCallbackMock = vi.fn();
   const failedCallbackMock = vi.fn();
 
-  it('should call logPromiseResults with successfull plugin result', () => {
+  it('should call logPromiseResults with successfully plugin result', () => {
```
matejchalk, line 9:
```suggestion
  it('should call logPromiseResults with successful plugin result', () => {
```

packages/utils/src/lib/filter.ts
```diff
@@ -0,0 +1,24 @@
...
+      .map(item => ({
+        ...item,
+        refs: filterBy(item.refs, refFilterFn),
+      }))
+      // remove item with empty refs
+      .filter(item => item.refs.length),
+  );
+}
+
+export function filterBy<T = unknown>(
```
matejchalk, line 16: All type parameters are `unknown` by default:

```suggestion
export function filterBy<T>(
```

packages/utils/src/lib/filter.ts
```diff
@@ -0,0 +1,24 @@
...
+  refs: T[] = [],
```
matejchalk, line 17: Why make `refs` parameter optional? What would be the point of calling `filterBy()` without any arguments? :raised_eyebrow:

matejchalk (commented):
Overall looks good :+1:

packages/utils/src/lib/filter.ts
```diff
@@ -0,0 +1,24 @@
+export function filterItemRefsBy<T extends { refs: object[] }>(
+  items: T[],
+  refFilterFn?: (item: T['refs'][number]) => boolean,
```
Tlacenka, line 3: Why is the filtering function optional? In what use case would I want to call a `filterX` function and not filter anything in it?

packages/utils/src/lib/filter.ts
Tlacenka, line 16: This function only calls `filter`. Is it necessary to have it separate from the `filterItemRefsBy`? Will it ever be called on its own instead of calling `filter`?

examples/plugins/src/lighthouse/src/lighthouse.plugin.ts
```diff
@@ -74,8 +74,10 @@ export async function create(options: PluginOptions) {
       onlyCategories: ['performance'],
       headless,
     }),
-    audits: filterAuditsBySlug(audits, onlyAudits),
-    groups: filterGroupsByAuditSlug([categoryCorePerfGroup], onlyAudits),
+    audits: filterBy(audits, ({ slug }) => onlyAudits.includes(slug)),
```
Tlacenka, line 77: Like this.
```suggestion
    audits: audits.filter(({ slug }) => onlyAudits.includes(slug)),
```

packages/utils/src/lib/filter.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function filterBy<T = unknown>(
+  refs: T[] = [],
+  filterFn?: (t: T) => boolean,
+): T[] {
+  if (filterFn === undefined || refs.length === 0) {
+    return refs;
+  }
+  return refs.filter(filterFn);
+}
```
matejchalk, line 24: I had a realization - this `filterBy` function doesn't need to exist :sweat_smile:  I thought the reason for it was could apply a filter optionally (`filterBy(arr, fn)` instead of `fn ? arr.filter(fn) : arr`), but actually you're never doing that in the production code. So actually you can just replace `filterBy(arr, fn)` with `arr.filter(fn)` and remove this wrapper function completely. It just introduces an unnecessary abstraction, and `Array.filter` is smarter with regards to type inference anyway (e.g. supports type predicates).

matejchalk (changes requested):

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -49,8 +58,364 @@ describe('validateOnlyAudits', () => {
...
+              { slug: 'd', weight: 1 },
+              { slug: 'e', weight: 1 },
+              { slug: 'f', weight: 1 },
+            ],
+          },
+        ],
+        'g2',
+      ),
+    ).toBeTruthy();
+  });
```
Tlacenka, line 94: Two things here that apply to both tests of `validateOnlyCategories`.
1. No need to add so many group references as the function does not care about them so it unnecessarily bloats the test.
2. Please use realistic values, example provided for generic categories (to be replaced by real life scenario). It is much more readable that way.

```suggestion
    expect(
      validateOnlyCategories(
        [
          {
            slug: 'performance',
            title: 'Performance',
            refs: [{ slug: 'speed-index', weight: 1 }],
          },
          {
            slug: 'coverage',
            title: 'Code coverage',
            refs: [{ slug: 'function-coverage', weight: 1 }],
          },
        ],
        'performance',
      ),
    ).toBeTruthy();
  });
```

packages/plugin-lighthouse/src/lib/utils.ts
BioPhoton: true

packages/utils/src/lib/filter.ts
BioPhoton: deleted it

packages/utils/src/lib/filter.ts
BioPhoton: i removed filterBy

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
BioPhoton: done updating ALL test data :)

matejchalk (dismissed):

packages/utils/src/lib/filter.ts
BioPhoton: filterBy is removed now

BioPhoton:
@Tlacenka I updated all test data with realistic values. Can you have a quick look please.

matejchalk (approved):

PR #516, opened 2024-02-19
refactor(cli): correct types in core config middleware

### LINKED ISSUE #411: Fix type inconsistency for CLI options
### User story

As a contributor, I want TypeScript to accurately indicate types within the CLI.

### Acceptance criteria

- [x] Correctly options provided by CLI for both `persist` and `upload` as optional (one of the main places where this happens is `config-middleware.ts`).
- [x] Enable `@typescript-eslint/no-unnecessary-condition` which now highlights cases which incorrectly infer the type.

### Implementation details

Currently, the types incorrectly represent the use case for CLI - while in `CoreConfig` provided from a code pushup config file the `upload` and `persist` options either don't exist at all or need to be fully filled out, the CLI options for `persist` and `upload` merely override individual options and therefore each suboption within their type should also be optional.

Tlacenka:
Closes #411 
In this PR, I address the following:
- Persist options are still optional in both config file and CLI. Their defaults are inserted inside the middleware so the `normalizePersistConfig` function was removed.
- Upload options are either optional in CLI or in all-or-nothing mode in core config. I fixed incorrect type assumptions, enabling for the ESLint rules to be enabled.
- I added tests for this logic.
- I fixed the last remaining ESLint issue (one-liner) and removed the temporary ESLint config.
- I narrowed down types expected and returned by middlewares.
- Created #518 as a result.

@BioPhoton 
Aside from `persist` and `upload`, there was only one remaining ESLint issue (one-liner) in `cli`. I removed it and therefore enabled all rules in `cli`. Current `core` ESLint issues are unrelated and therefore to be addressed in #475 .


packages/cli/src/lib/implementation/core-config.middleware.ts
```diff
@@ -1,23 +1,24 @@
...
   PERSIST_FORMAT,
   PERSIST_OUTPUT_DIR,
+  uploadConfigSchema,
 } from '@code-pushup/models';
+import { CoreConfigCliOptions } from './core-config.model';
 import { GeneralCliOptions } from './global.model';
 
 export async function coreConfigMiddleware<
-  T extends Partial<GeneralCliOptions & CoreConfig>,
+  T extends Partial<GeneralCliOptions & CoreConfigCliOptions>,
```
BioPhoton, line 12: Could we name it RcConfig? I is covering the type of the rc file.

packages/cli/src/lib/implementation/core-config.integration.test.ts
```diff
@@ -150,4 +153,49 @@ describe('parsing values from CLI and middleware', () => {
...
+      {
+        options: { ...yargsCoreConfigOptionsDefinition() },
+        middlewares: [{ middlewareFunction: coreConfigMiddleware }],
+      },
+    ).parseAsync();
+
+    expect(upload).toBeUndefined();
+  });
+
+  it('should throw for an incomplete upload configuration', async () => {
```
BioPhoton, line 186: I like this one!

packages/core/src/lib/upload.ts
```diff
@@ -20,13 +19,12 @@ export async function upload(
   options: UploadOptions,
   uploadFn: typeof uploadToPortal = uploadToPortal,
 ) {
-  const persist = normalizePersistConfig(options.persist);
-  if (!options.upload) {
-    throw new Error('upload config must be set');
+  if (options.upload == null) {
```
BioPhoton, line 22: I get it's same logical outcome. But the value will never be null. It will me `undefined` if ever.

Is this a best practice I don't know?

BioPhoton (commented):
LOVE THIS PR!
Thought about taking in after all my hanging ones are in... 

Left comments.

packages/core/src/lib/upload.ts
matejchalk: This is a common practice we've established at FlowUp. Not sure how widespread it is, but the fact the [ESLint core `eqeqeq` rule supports a `'always', { null: 'never' }` option](https://eslint.org/docs/latest/rules/eqeqeq#always) is some indication there's split over it in the comminuty :thinking: 

The rationale for `x == null` and `x != null` is to treat `null` and `undefined` the same way, because there's no consistent semantic distinction between them. Having two distinct nullish values is a very strange quirk of JavaScript. In practice, picking `null` or `undefined` is basically arbitrary, so checking for a nullish value with minimal syntax is a practical and defensive approach in general. The key question for me is: "If you expected `undefined` and received `null` instead, would you want to treat it any differently?"

packages/cli/src/lib/implementation/core-config.model.ts
```diff
@@ -20,6 +20,6 @@ export type ConfigCliOptions = {
   tsconfig?: string;
 };
 
-export type CoreConfigCliOptions = ConfigCliOptions &
-  PersistConfigCliOptions &
-  UploadConfigCliOptions;
+export type CoreConfigCliOptions = Omit<CoreConfig, 'upload'> & {
+  upload?: Partial<Omit<UploadConfig, 'timeout'>>;
+};
```
matejchalk, line 25: We could add an `--upload.timeout` CLI argument, so that the config and CLI upload options are 1-to-1.

packages/utils/src/lib/reports/utils.ts
```diff
@@ -237,7 +237,7 @@ export function compareIssueSeverity(
 type LoadedReportFormat<T extends Format> = T extends 'json' ? Report : string;
 
 export async function loadReport<T extends Format>(
-  options: Required<Pick<PersistConfig, 'outputDir' | 'filename'>> & {
+  options: Omit<NonNullable<Required<PersistConfig>>, 'format'> & {
```
matejchalk, line 240: The `NonNullable<...>` type transformation doesn't have any effect here, the `Required<PersistConfig>` type doesn't include `null` or `undefined`.

```suggestion
  options: Omit<Required<PersistConfig>, 'format'> & {
```

packages/cli/src/lib/implementation/global.utils.ts
```diff
@@ -40,7 +40,5 @@ export function logErrorBeforeThrow<T extends (...args: any[]) => any>(
 }
 
 export function coerceArray<T extends string>(param: T | T[] = []): T[] {
-  return [
-    ...new Set(toArray(param).flatMap((f: T) => f.split(',') as T[]) || []),
-  ];
+  return [...new Set(toArray(param).flatMap((f: T) => f.split(',') as T[]))];
 }
```
matejchalk, line 44: There's no advantage to `coerceArray` accepting a type parameter - we don't call it ourselves, we pass it to Yargs' `coerce` option which is loosely typed. And the type assumptions are incorrect anyway - if you call `.split(',')` on `T`, the result won't be a `T[]`.

```ts
export function coerceArray(param: string | string[] = []): string[] {
  return [...new Set(toArray(param).flatMap(f => f.split(',')))];
}
```

packages/cli/src/lib/implementation/core-config.model.ts
Tlacenka: Seems like a very specific argument that most people will not come across. 🤔 But you can create an issue for this and we can discuss it on the refinement.

packages/cli/src/lib/implementation/core-config.middleware.ts
Tlacenka: The way I understand it, this is the type for CLI arguments only that are passed to the function, not the config file which is loaded inside of this function and returned (hence why I added the return type as `CoreConfig` in my latest commit).

So I am not sure what you mean by this, can you explain?

packages/cli/src/lib/implementation/global.utils.ts
Tlacenka: Good point

packages/cli/src/lib/implementation/core-config.model.ts
matejchalk: Now that I think about it, I'm not sure it helps much. Lets skip it until some user needs it.

matejchalk (approved):

PR #522, opened 2024-02-22
fix(plugin-eslint): handle rules which emit column 0

matejchalk:
Fixed validation error coming from [unicorn/no-abusive-eslint-disable](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-abusive-eslint-disable.md) rule - column 0 wasn't handled and failed validation.

#### CI error

image

#### Debugging

image

image


vmasek (approved):

Tlacenka (approved):
As long as we don't highlight column range, this should be fine.

Issue #293, opened 2023-11-20
AuditOutput value should be float


BioPhoton:
As a user I want to also provide floating numbers as audit value.

- value should accept
  - int
  - float

kristihoskova:
@BioPhoton create issue for the lighthouse plugin

matejchalk:
Fixed in:
- #693
- #725

PR #487, opened 2024-02-07
refactor(utils): use cliui as central logger

BioPhoton:
**This PR includes:**
- changes the return type of log helpers to fit the new logger
- replacement parts of 'isaac-cliui' with new logger
- replacement of all console log,info,warning,error
- extending the new logger with a 'row' function to encapsulate '3table' 
   opened https://github.com/poppinss/cliui/issues/14 to get rid of 3table and use `cliui` helpers instead
   opened #513 
- refactor the stdo of the report 
- added hack to clear logs for testing  
   issue opened https://github.com/poppinss/cliui/issues/15
   PR opened https://github.com/poppinss/cliui/issues/16

> [!NOTE]
> I will move #513 into a new PR and implement in when the PR in cliui is in if I reopen ever.
> created #552 

**Testing:**
I tried to use hooks for the logger tests to switch to `raw` mode and cleanup the logs `beforeEach` run.

I was only able to use this setup in `utils` which imports directly the file:

Working code:  
```ts
import {beforeEach, vi} from "vitest";
import {ui} from "../../src/lib/logging";

vi.mock('../../src/lib/logging', async () => {
  const module = await vi.importActual('../../src/lib/logging');

  module.ui().switchMode('raw')
  return module;
});

beforeEach(() => {
   ui().logger.flushLogs();
})
```

However in `core` and `cli` i was only able to have the mode switching in the setup script and hat to move the logger cleanup in the test files. If I import `ui` from `@code-pushup/utils` many other tests fail.

Working reduced script:  
```ts
import {ui} from "@code-pushup/utils";

vi.mock('@code-pushup/utils', async () => {
  const module = await vi.importActual('@code-pushup/utils');

  module.ui().switchMode('raw')
  return module;
});
```

Update:

I was able to implement the beforeEach hook with async imports:

```ts

beforeEach(async () => {
  const {ui} = await vi.importActual<{ ui: () => CliUi }>(
    '@code-pushup/utils',
  );
  ui().logger.flushLogs();
});
```

closes #502
related #513

PR #523, opened 2024-02-24
feat(plugin-lighthouse): implement basic audit parsing

### LINKED ISSUE #90: Implement Lighthouse plugin

BioPhoton:
**This PR includes:**
- logic to parse LH audits to CP audits
- tests 
- test data for all unimplemented details

Supported detail types:
- [Table](https://github.com/GoogleChrome/lighthouse/blob/8662fbe9a2c705c7ed35ef901dad1be7b31be604/types/lhr/audit-details.d.ts#L106)
- [Opportunity](https://github.com/GoogleChrome/lighthouse/blob/8662fbe9a2c705c7ed35ef901dad1be7b31be604/types/lhr/audit-details.d.ts#L75)

**Further considerations:**
As many of the lighthouse audits have hard to parse details we should discuss #293 again.

partially resolves #90


packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -54,3 +59,265 @@ describe('validateOnlyAudits', () => {
...
+      {
+        displayValue: undefined,
+        score: 1,
+        slug: 'audit',
+        value: 0,
+      },
+    ]);
+  });
+
+  it('should not parse lhr details debugdata', () => {
```
Tlacenka, line 85: The tests should verify the currently implemented logic which is that a lot of the types are not supported yet and some are not implemented yet.

So the same description edit should go to all other types that are not implemented yet.

```suggestion
  it('should inform that debugdata type is not supported yet', () => {
```

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -54,3 +59,265 @@ describe('validateOnlyAudits', () => {
...
+          message: 'sourceLocation, description',
+          severity: 'info',
+        } satisfies Issue,
+      ],
+    });
+  });
+});
+
+describe('opportunityToDetails', () => {
+  it('should parse valid lhr details', () => {
```
Tlacenka, line 290: I like the description distinction you did in the table that distinguishes valid/invalid and empty/non-empty.

```suggestion
  it('should parse a valid non-empty lhr opportunity type', () => {
```

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -54,3 +59,265 @@ describe('validateOnlyAudits', () => {
...
+        },
+      },
+    });
+
+    expect(outputs[0]?.details?.issues[0]).toEqual(
+      expect.objectContaining({
+        message: 'Parsing details from type filmstrip is not implemented.',
+        severity: 'info',
+      }),
+    );
```
Tlacenka, line 138: Instead of accessing possibly undefined properties, you can check relevant areas of the whole object using `objectContaining` as well.

```suggestion
    expect(outputs).toStrictEqual([
      expect.objectContaining({
        details: {
          issues: [
            {
              message:
                'Parsing details from type criticalrequestchain is not implemented.',
              severity: 'info',
            },
          ],
        },
      }),
    ]);
```

packages/plugin-lighthouse/src/lib/utils.ts
```diff
@@ -68,3 +75,89 @@ export function validateOnlyAudits(
...
+  | Details.CriticalRequestChain
+  | Details.List
+  | Details.TreemapData
+  | Details.Screenshot
+  | Details.Filmstrip
+  | Details.DebugData;
+type UnsupportedDetailTypes = UnsupportedDetail['type'];
+
+export function toAuditOutputs(
+  lhrAudits: Record<string, Result>,
```
Tlacenka, line 89: Will the key ever be relevant for this function? If not, wouldn't it be easier to just pass the values to this function?

```suggestion
  lhrAudits: Result[],
```

Tlacenka (commented):
Nice small increment!

It would really help me to see the Lighthouse plugin criteria for additional context.
If you plan on working on the official plugin more, could we go through the epic on a refinement?

packages/plugin-lighthouse/src/lib/utils.ts
```diff
@@ -68,3 +75,89 @@ export function validateOnlyAudits(
...
+              ...auditOutput,
+              details: {
+                issues: [
+                  {
+                    message: `Parsing details from type ${unsupportedType} is not implemented.`,
+                    severity: 'info',
+                  },
+                ],
+              },
+            };
```
matejchalk, line 131: I don't think this is a good way of dealing with unsupported features. I'd log a warning to the console instead of creating issues in the report.

packages/plugin-lighthouse/src/lib/utils.ts
```diff
@@ -68,3 +75,89 @@ export function validateOnlyAudits(
...
+      numericValue: value = 0, // not every audit has a numericValue
+      details,
+      displayValue,
+    }: Result) => {
+      const auditOutput: AuditOutput = {
+        slug,
+        score: score ?? 1, // score can be null
+        value,
+        displayValue,
+      };
```
matejchalk, line 104: Interesting scenario. Not sure if `score: 1, value: 0` is the best way to represent such audits, filtering them out could also be an option :thinking: Which Lighthouse audits does this concern?

packages/plugin-lighthouse/src/lib/utils.ts
```diff
@@ -68,3 +75,89 @@ export function validateOnlyAudits(
...
+  | Details.TreemapData
+  | Details.Screenshot
+  | Details.Filmstrip
+  | Details.DebugData;
+type UnsupportedDetailTypes = UnsupportedDetail['type'];
+
+export function toAuditOutputs(
+  lhrAudits: Result[],
+): AuditOutputs {
+  return Object.values(lhrAudits).map(
```
matejchalk, line 91: It's already an array:

```suggestion
  return lhrAudits.map(
```

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -54,3 +59,272 @@ describe('validateOnlyAudits', () => {
...
+      },
+    });
+
+    expect(outputs[0]?.details?.issues[0]).toEqual(
+      expect.objectContaining({
+        message: 'Parsing details from type screenshot is not implemented.',
+        severity: 'info',
+      }),
+    );
+  });
```
matejchalk, line 171: The test name says `should parse valid XYZ details`, but the assertion expects a `Parsing details from type XYZ is not implemented` message. Something's not right here :sweat_smile: (Same goes for next 2 test cases.)

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -54,3 +59,272 @@ describe('validateOnlyAudits', () => {
...
+    ).toStrictEqual({
+      issues: [
+        {
+          message: 'url, responseTime',
+          severity: 'info',
+        } satisfies Issue,
+      ],
+    });
+  });
+});
```
matejchalk, line 330: Yes, this test case actually uses some realistic looking data. This enables us to evaluate the behaviour meaningfully.

Look at the input data and look at the output data. Would you say we converted the Lighthouse data into something useful? The Lighthouse data tells me that what the response time was for a specific URL. But the resulting issue tells me none of this useful information. Instead it cryptically tells me what heading keys are. How is that useful in any way? :confounded: 

However, it seems like the table and opportunities data structures could map to our issues array fairly well, if you want to go down that route.
- A table with static columns can be seen as a fancy list. So we could create an issue for each row (`items` array).
- Our issue's don't support arbitray object data, but a string concatenation of the key=value pairs could serve as the `message`. E.g. to generate something like "`URL: "https://staging.code-pushup.dev/login", Time Spent: 449.292_000_000_000_03`":
  ```ts
  items.map((item): Issue => ({ 
    message: headings.map(({ key, label }) => `${label}: ${JSON.stringify(item[key])}`,
    severity: 'info',
  })
  ```

But the best solution is to refine an issue for extending the `AuditDetails` model to support tabular data.

matejchalk (changes requested):

packages/plugin-lighthouse/src/lib/utils.ts
```diff
@@ -68,3 +75,89 @@ export function validateOnlyAudits(
...
+                    message: `Parsing details from type ${unsupportedType} is not implemented.`,
+                    severity: 'info',
+                  },
+                ],
+              },
+            };
+        }
+      }
+
+      return auditOutput;
```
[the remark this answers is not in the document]
BioPhoton: fixed

packages/plugin-lighthouse/src/lib/utils.ts
BioPhoton: Returned the audit output without details and added a `console.info`.

packages/plugin-lighthouse/src/lib/utils.ts
BioPhoton: There are multiple scenarios that a was able to track down:
- The provided URL is not accessible. In this case all audits have a score of `null` but are still present in the result.
- If the URL is valid, the following `scoreDisplayMode` have a score of `null`:
  -  `informative`
  - `notApplicable`
  - `manual`
  
  Here a screenshot:
  
Screenshot 2024-02-28 at 16 16 32

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
BioPhoton: adjusted the test description and adopted the types

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
BioPhoton: Yes, I opened the PR while you where gone and was hoping to get more attention on the tabular data issue.

I did the hack above to have "something" working besides "not implemented" messages for all detail data.

If you like I can turn all `tables` into issues, or I remove the implementation of detail data parsing for now and add it back later when we have tabular data supported.

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: I added more real data wherever I could.

BioPhoton:
@Tlacenka 
> could we go through the epic on a refinement?

Yes, but I assume a quick sync between us is good for now.

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
matejchalk: I'm fine with either, but leaning towards skipping audit details for now.

packages/plugin-lighthouse/src/lib/utils.ts
matejchalk: (I'm fine with this being a separate issue, BTW :top:)

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -54,3 +59,266 @@ describe('validateOnlyAudits', () => {
...
+      {
+        displayValue: '2.8 s',
+        score: 0.55,
+        slug: 'first-contentful-paint',
+        value: 2838.974,
+      },
+    ]);
+  });
+
+  it('should parse lhr audits with score null to score 1', () => {
```
Tlacenka, line 89:
```suggestion
  it('should convert null score to 1', () => {
```

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -54,3 +59,266 @@ describe('validateOnlyAudits', () => {
...
+          type: 'debugdata',
+          items: [
+            {
+              cumulativeLayoutShiftMainFrame: 0.000_350_978_852_728_593_95,
+            },
+          ],
+        },
+      },
+    ]);
+    expect(outputs[0]?.details).toBeUndefined();
```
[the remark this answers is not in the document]
BioPhoton: I skipped this part as I wait for #487

packages/plugin-lighthouse/src/lib/utils.ts
BioPhoton: Removed all code+tests related to details and opened #530

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
BioPhoton: Done 

also see #530

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: done

matejchalk (dismissed):

matejchalk (approved):

matejchalk (approved):

Issue #497, opened 2024-02-13
Handling crashed plugins


Tlacenka:
### User story

As a CP user, I wish to be able to understand what is going on, especially if one of the plugins crash. Currently, when a plugin crashes, the CLI continues. However, as the plugins are not filtered properly during scoring, the report is not created and I get a cryptic "invalid group reference" error message.

### Acceptance criteria

- [ ] Crashed plugins are filtered out during scoring and sorting, so that a valid report may be created.
- [x] The user is sufficiently informed about what went wrong in the failed plugin, e.g. by printing `stdout` and `stderr` (not just process code 1).

### Implementation details

Error message example:
![image](https://github.com/code-pushup/cli/assets/6306671/012b34fb-ce22-4f49-82cd-1daca44adda5)

vmasek:
Conclusion: we want to let the "broken" run fail with no report collected as we want users to notice failed collection and don't want them to consume broken category scores.

PR #528, opened 2024-02-28
fix: stop CLI when a plugin fails

Tlacenka:
Related to #497 
Until now, the plugin failure was silent and ended up by failing during scoring the incomplete report (no Report parsing is done atm).

In this PR, I removed the `catch` which caused this behaviour in this PR so now the plugins fail after they are run.

I did not delve into the user experience of the logs (they are printed twice or three times still) in this PR, I only fixed the plugin fail behaviour.


matejchalk (approved):
Great catch :+1: (pun intended :wink:)

Issue #426, opened 2024-01-19
ESLint plugin not recognising rules matching a folder


Tlacenka:
### What happened?

I received different ESLint results locally and in Code PushUp report.

### What would you expect to happen?

I expected the `no-magic-numbers` rule to be turned off for the `perf` folder in the `utils` package.

### What steps did you take?

1. I added an exception for some ESLint rules in the `perf` folder and added it to the root `.eslintrc.json`.
```
 {
      "files": ["perf/**/*.ts"],
      "rules": {
        "no-magic-numbers": "off",
        "sonarjs/no-duplicate-string": "off"
      }
 }
```
2. I checked that ESLint no longer showed warnings e.g. for `no-magic-numbers` in these files.
3. I ran Code PushUp and displayed the results in portal.
4. There I saw these rules being treated as enabled.
![image](https://github.com/code-pushup/cli/assets/6306671/073ab060-3960-4aaf-a4a1-e0f99461ef72)

### What operation system are you on?

Windows

vmasek:
I think the culprit is that you have to turn off the `/* eslint-disable @typescript-eslint/no-magic-numbers */` not the `/* eslint-disable no-magic-numbers */` as we migrated our preset to use the rule from typescript that supports much more clever assessments of what is and isn't magic number

PR #534, opened 2024-03-04
refactor(utils): treat groups as optional

### LINKED ISSUE #421: Handling plugins with no groups in `ScoredReport`

Tlacenka:
Closes #421 
In this PR, I refactor the scoring and sorting logic to treat groups as optional.

By doing so, I am also able to remove the last `code-pushup.eslintrc.json` file. 🎉 
Technically, we should be at perfect ESLint score now. The only remaining ESLint issues are a result of a bug or insufficiently excluded ESLint location:
- `perf` folder which turns off some rules but the plugin counts them in #426 
- `plugin-eslint/mocks/fixtures/nx-monorepo` which should probably be excluded from our ESLint and only included in the ESLint that is run within the tests. WDYT @matejchalk


packages/utils/src/lib/reports/scoring.ts
```diff
@@ -19,18 +19,18 @@ export function scoreReport(report: Report): ScoredReport {
   const allScoredAuditsAndGroups = new Map<string, AuditReport | ScoredGroup>();
 
   const scoredPlugins = report.plugins.map(plugin => {
-    const { slug, audits, groups } = plugin;
+    const { groups, ...pluginProperties } = plugin;
```
matejchalk, line 22: "Props" is a common abbreviation for "properties" in code.

```suggestion
    const { groups, ...pluginProps } = plugin;
```

packages/utils/src/lib/reports/utils.ts
```diff
@@ -71,7 +71,7 @@ export function countCategoryAudits(
   // Create lookup object for groups within each plugin
   const groupLookup = plugins.reduce<Record<string, Record<string, Group>>>(
     (lookup, plugin) => {
-      if (plugin.groups.length === 0) {
+      if (plugin.groups == null || plugin.groups.length === 0) {
```
matejchalk, line 74: Not saying this is better, but it is shorter. Feel free to ignore :grin: 

```suggestion
      if (!plugin.groups?.length) {
```

Tlacenka: I like it the original way more as it explicitly states there are two valid cases in which it should end early.

PR #535, opened 2024-03-04
fix(plugin-eslint): pass arguments via config file instead of argv

### LINKED ISSUE #529: Bug Eslint Plugin: The command line is too long
### What happened?

I executed the eslint plugin inside nx monorepo
`await eslintPlugin(await eslintConfigFromNxProject('...')),`

I get an error that the command line is too long.
Digging into the plugin I found that eslint is called with a long list of arguments "'angular-eslint-prefer-standalone-component,angular-eslint-prefer-on-push-component-change-detection,...,,angular-eslint-template-eqeqeq,....." (around 9000-10000 characters)

The code pushup audit should run and not throw an error. I would have expected that eslint is called with a file and not with a list of command line arguments.

### What steps did you take?

I cannot provide the details of the project but can provide some general stats that could help to reproduce the issue

The eslintrc.json in the root folder contains for ["*.ts", "*.tsx"]
"plugin:@nx/typescript", "@code-pushup/eslint-config/typescript" as plugins
around 60 lines of nx boundaries configuration and around 40 lines of turining off eslint checks.
The project itself contains in addition the angular eslint plugin

0.22.8

Windows

### Node version

v20.11.1

### Relevant log output

```shell
Plugins failed:
The command line is too long.

Error: Plugins failed: 1 errors: The command line is too long.
 at executePlugins ... @code-pushup/cli/index.js:1679:11

or traced to 
async function executeRunnerConfig(cfg, onProgress) {
  const { args, command, outputFile, outputTransform } = cfg;
  const { duration, date } = await executeProcess({
    command,
    args,
    observer: { onStdout: onProgress }
  });
```

Tlacenka:
**WIP**
Closes #529 

In this PR, I rework the ESLint plugin to pass arguments to the runner via a configuration file.
When running the ESLint plugin locally now, the correct output is produced instead of `The command line is too long.`

<details>
  <summary> Part of the output </summary>

  image
</details>

### DIFF
--- a/packages/plugin-eslint/src/lib/runner/index.ts
+++ b/packages/plugin-eslint/src/lib/runner/index.ts
@@ -1,28 +1,28 @@
-import { mkdir, writeFile } from 'node:fs/promises';
-import { platform } from 'node:os';
+import { writeFile } from 'node:fs/promises';
 import { dirname, join } from 'node:path';
 import type { Audit, AuditOutput, RunnerConfig } from '@code-pushup/models';
-import { pluginWorkDir, readJsonFile, toArray } from '@code-pushup/utils';
+import {
+  ensureDirectoryExists,
+  pluginWorkDir,
+  readJsonFile,
+  toArray,
+} from '@code-pushup/utils';
+import { ESLintPluginRunnerConfig } from '../config';
 import { lint } from './lint';
 import { lintResultsToAudits } from './transform';
 
 export const WORKDIR = pluginWorkDir('eslint');
 export const RUNNER_OUTPUT_PATH = join(WORKDIR, 'runner-output.json');
 export const ESLINTRC_PATH = join(process.cwd(), WORKDIR, '.eslintrc.json');
+export const PLUGIN_CONFIG_PATH = join(
+  process.cwd(),
+  WORKDIR,
+  'plugin-config.json',
+);
 
-const AUDIT_SLUGS_SEP = ',';
-
-export async function executeRunner(argv = process.argv): Promise<void> {
-  const [slugs, eslintrc, ...patterns] = argv.slice(2);
-  if (!slugs) {
-    throw new Error('Invalid runner args - missing slugs argument');
-  }
-  if (!eslintrc) {
-    throw new Error('Invalid runner args - missing eslintrc argument');
-  }
-  if (patterns.length === 0) {
-    throw new Error('Invalid runner args - missing patterns argument');
-  }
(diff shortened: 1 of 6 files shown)


packages/plugin-eslint/src/lib/runner/index.ts
```diff
@@ -47,22 +47,23 @@ export async function executeRunner(argv = process.argv): Promise<void> {
...
   eslintrc: string,
   patterns: string | string[],
-): RunnerConfig {
+): Promise<RunnerConfig> {
+  const config: ESLintPluginRunnerConfig = {
+    eslintrc,
+    slugs: audits.map(audit => audit.slug),
+    patterns: toArray(patterns),
+  };
+  await ensureDirectoryExists(WORKDIR);
```
matejchalk, line 61: This way it'll infer what directory needs to be created from `PLUGIN_CONFIG_PATH` (used like this in `executeRunner`):

```suggestion
  await mkdir(dirname(PLUGIN_CONFIG_PATH), { recursive: true });
```

matejchalk (approved):
Looks promising :+1:

packages/plugin-eslint/src/lib/runner/index.ts
Tlacenka: Why not keep `ensureDirectoryExists`? It does the same thing plus it does not throw when the directory is already created.

matejchalk: > It does the same thing plus it does not throw when the directory is already created.

But neither does `mkdir('...', { recursive: true })`, AFAIK.

> Why not keep `ensureDirectoryExists` for the new path?

Although I don't actually see the point of the `ensureDirectoryExists` wrapper, the main suggestion is to use `dirname(PLUGIN_CONFIG_PATH)` instead of `WORKDIR`. That way it's more directly linked to the file you want to create there. And if we decide to not put the `plugin-config.json` directly under `WORKDIR`, we can change without touching this logic. 

You could even extract those two statements into a helper function:
```ts
await mkdir(dirname(filePath), { recursive: true });
await writeFile(filePath, content);
```

matejchalk (approved):

packages/plugin-eslint/src/lib/runner/index.ts
Tlacenka: I see. Thought using helpers consistently is the goal. Perhaps we don't need the helper then. But that is a story for another time.

PR #541, opened 2024-03-05
feat(core): add history logic

BioPhoton:
This PR is related to the history command and part of #216 and adds the history logic to `core`.

**It includes the following logic:**  
- `prepareHashes` - logic to get the commit hash from a list of logs
- `getHashes` - logic to get a list of hashes based on options `from`, `to`, `maxCount`
- `history` - the history logic that checks out all commits and runs `collect` and optionally `upload`

**Testing:**  
- Added unit tests and integration tests. For the integration tests I used a separate git folder to avoid the flaky test. I thing we should document that somewhere?


packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,110 @@
...
+  GlobalOptions &
+  HistoryOnlyOptions;
+
+export async function history(
+  config: HistoryOptions,
+  commits: string[],
+): Promise<string[]> {
+  const initialBranch: string = await getCurrentBranchOrTag();
+
+  const { uploadReports = true } = config as unknown as HistoryOptions;
```
matejchalk, line 33:
```suggestion
  const { uploadReports = true } = config;
```

packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export type HistoryOnlyOptions = {
+  targetBranch?: string;
+  uploadReports?: boolean;
+  forceCleanStatus?: true;
+};
+export type HistoryOptions = Required<
+  Pick<CoreConfig, 'plugins' | 'categories'> & {
+    persist: Required<PersistConfig>;
+  } & { upload: Required<UploadConfig> }
```
matejchalk, line 22:
```suggestion
  Pick<CoreConfig, 'plugins' | 'categories'> & {
    persist: Required<PersistConfig>;
    upload: Required<UploadConfig>;
  }
```

packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,110 @@
...
+} from '@code-pushup/models';
+import { getCurrentBranchOrTag, safeCheckout } from '@code-pushup/utils';
+import { collectAndPersistReports } from './collect-and-persist';
+import { GlobalOptions } from './types';
+import { upload as uploadCommandLogic } from './upload';
+
+export type HistoryOnlyOptions = {
+  targetBranch?: string;
+  uploadReports?: boolean;
+  forceCleanStatus?: true;
```
matejchalk, line 17:
```suggestion
  forceCleanStatus?: boolean;
```

packages/core/src/lib/history.ts
matejchalk, line 33: I usually define optional boolean parameters so it's off by default, then I can treat all falsy values the same. You can always invert the parameter meaning, e.g. this would become `skipUploads`.

packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,110 @@
...
+  // eslint-disable-next-line functional/no-loop-statements
+  for (const commit of commits) {
+    console.info(`Collect ${commit}`);
+    await safeCheckout(commit, { forceCleanStatus: config.forceCleanStatus });
+
+    const currentConfig: HistoryOptions = {
+      ...config,
+      persist: {
+        ...config.persist,
+        format: ['json' as Format],
```
matejchalk, line 48:
```suggestion
        format: ['json'],
```

packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,110 @@
...
+  CoreConfig,
+  Format,
+  PersistConfig,
+  UploadConfig,
+  uploadConfigSchema,
+} from '@code-pushup/models';
+import { getCurrentBranchOrTag, safeCheckout } from '@code-pushup/utils';
+import { collectAndPersistReports } from './collect-and-persist';
+import { GlobalOptions } from './types';
+import { upload as uploadCommandLogic } from './upload';
```
matejchalk, line 12: Why rename it?

packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,110 @@
...
+    await collectAndPersistReports(currentConfig);
+
+    if (uploadReports) {
+      const result = uploadConfigSchema.safeParse(currentConfig.upload);
+      if (result.success) {
+        await uploadCommandLogic({ ...currentConfig, upload: result.data });
+      } else {
+        console.error(`Collecting ${commit} failed.`);
+        console.error(result.error);
+      }
```
matejchalk, line 62: The error message doesn't mention the cause - invalid upload config. Can that even happen, though? If this is to be triggered from the command-line, then it will have already been parsed in the CLI middleware. So probably the `safeParse` can be ommitted here.

```suggestion
      await uploadCommandLogic(currentConfig);
```

packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,110 @@
...
+  const { from, to, maxCount } = options;
+
+  if (from || to) {
+    // validate from & to
+    if (from === undefined || from === '') {
+      throw new Error('from has to be defined');
+    }
+    if (to === undefined || to === '') {
+      throw new Error('to has to be defined');
+    }
```
matejchalk, line 93: Why require both? The `git log` command with only `--since` or only `--until`.

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+vi.mock('./upload', () => ({
+  upload: vi.fn(),
+}));
+
+describe('history', () => {
+  it('should check out all passed commits and reset to initial branch or tag', async () => {
+    const historyOptions: HistoryOptions = {
+      ...(MINIMAL_CONFIG_MOCK as HistoryOptions),
+    };
```
matejchalk, line 32:
```suggestion
    const historyOptions = MINIMAL_CONFIG_MOCK as HistoryOptions;
```

There are several more occurrences in the file.

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    expect(safeCheckout).toHaveBeenNthCalledWith(1, 'abc', {
+      forceCleanStatus: undefined,
+    });
+    expect(safeCheckout).toHaveBeenNthCalledWith(2, 'def', {
+      forceCleanStatus: undefined,
+    });
+    // reset
+    expect(safeCheckout).toHaveBeenNthCalledWith(3, 'main', {
+      forceCleanStatus: undefined,
+    });
```
matejchalk, line 50: The Vitest equality checks don't differentiate between missing props and props with `undefined` value.

```suggestion
    expect(safeCheckout).toHaveBeenNthCalledWith(1, 'abc', {});
    expect(safeCheckout).toHaveBeenNthCalledWith(2, 'def', {});
    // reset
    expect(safeCheckout).toHaveBeenNthCalledWith(3, 'main', {});
```

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  const utils: object = await vi.importActual('@code-pushup/utils');
+  let currentBranchOrTag = 'main';
+  return {
+    ...utils,
+    safeCheckout: vi.fn().mockImplementation((branch: string) => {
+      currentBranchOrTag = branch;
+    }),
+    getCurrentBranchOrTag: vi.fn().mockImplementation(() => currentBranchOrTag),
+  };
+});
```
Tlacenka, line 18: All tests are using `main` branch as the current branch.
Therefore, there is no need to implement additional logic for a potential different branch.

```suggestion
vi.mock('@code-pushup/utils', async () => {
  const utils: object = await vi.importActual('@code-pushup/utils');
  return {
    ...utils,
    safeCheckout: vi.fn(),
    getCurrentBranchOrTag: vi.fn().mockReturnValue('main'),
  };
});
});
```

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+describe('history', () => {
+  it('should check out all passed commits and reset to initial branch or tag', async () => {
+    const historyOptions: HistoryOptions = {
+      ...(MINIMAL_CONFIG_MOCK as HistoryOptions),
+    };
+
+    await history(historyOptions, ['abc', 'def']);
+
+    expect(getCurrentBranchOrTag).toHaveBeenCalledTimes(1);
+    expect(getCurrentBranchOrTag).toHaveReturnedWith('main');
```
Tlacenka, line 37: Once the `utils` mocks are simplified, this is no longer needed (it would only be checking the `jest.fn()` functionality).
```suggestion
```

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  });
+
+  it('should return correct number of results', async () => {
+    const historyOptions: HistoryOptions = {
+      ...(MINIMAL_CONFIG_MOCK as HistoryOptions),
+    };
+
+    const results = await history(historyOptions, ['abc', 'def']);
+
+    expect(results).toHaveLength(2);
```
Tlacenka, line 60: Since the array is fully verified in the subsequent command, there is no need to check the length in a separate command
```suggestion
```

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      ...(MINIMAL_CONFIG_MOCK as HistoryOptions),
+    };
+    await history(historyOptions, ['abc']);
+
+    expect(upload).not.toHaveBeenCalled();
+  });
+});
+
+describe('prepareHashes', () => {
+  it('should get all commits from log if no option is passed', () => {
```
Tlacenka, line 132: What options?
```suggestion
  it('should return commit hashes in reverse order, () => {
```

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    expect(
+      prepareHashes({ all: [{ hash: '1' }, { hash: '5' }] }),
```
Tlacenka, line 134: This makes TS unhappy but not sure how to solve that without `as unknown as LogResult`. Any ideas? cc @matejchalk

packages/core/src/lib/history.integration.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  describe('with a branch and commits clean', () => {
+    const commits: string[] = [];
+    beforeAll(async () => {
+      await writeFile(join(baseDir, 'README.md'), '# hello-world\n');
+      await emptyGit.add('README.md');
+      await emptyGit.commit('Create README');
+      // eslint-disable-next-line functional/immutable-data
+      commits.push((await emptyGit.log()).latest?.hash);
```
Tlacenka, line 40: TS is unhappy about this.
I would almost say it is okay to use `!` here because the commit happened in the previous command. WDYT? cc @matejchalk 

Applicable to all places where this happens.
```suggestion
      commits.push((await emptyGit.log()).latest!.hash);
```

packages/core/src/lib/history.integration.test.ts
```diff
@@ -0,0 +1,94 @@
...
+      await emptyGit.deleteLocalBranch('feature-branch');
+    });
+
+    it('getHashes should get all commits from log if no option is passed', async () => {
+      await expect(getHashes({}, emptyGit)).resolves.toStrictEqual(commits);
+    });
+
+    it('getHashes should get last 2 commits from log if maxCount is set to 2', async () => {
+      await expect(getHashes({ maxCount: 2 }, emptyGit)).resolves.toStrictEqual(
+        [commits[1], commits[2]],
```
Tlacenka, line 69: This does not require knowing the length of the commits.
```suggestion
        [commits.at(-2), commits.at(-1)],
```

packages/core/src/lib/history.integration.test.ts
```diff
@@ -0,0 +1,94 @@
+import { mkdir, rm, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { type SimpleGit, simpleGit } from 'simple-git';
+import { afterAll, beforeAll, describe, expect } from 'vitest';
+import { getHashes } from './history';
+
+// we need a separate folder that is not cleaned in `global-setup.ts`, otherwise the tests can't execute in parallel
+const gitTestFolder = 'core-history-git-test';
+describe('git utils in a git repo', () => {
```
Tlacenka, line 9: The only tested function is `getHashes`
```suggestion
describe('getHashes', () => {
```

packages/core/src/lib/history.integration.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    await emptyGit.addConfig('user.name', 'John Doe');
+    await emptyGit.addConfig('user.email', 'john.doe@example.com');
+  });
+
+  afterAll(async () => {
+    await rm(baseDir, { recursive: true, force: true });
+  });
+
+  describe('without a branch and commits', () => {
+    it('getHashes should throw', async () => {
```
Tlacenka, line 26: Once it is clear this is the only function tested, its name does not need to be included in all the tests.

This should be modified in all other test cases as well.
```suggestion
    it('should throw', async () => {
```

Tlacenka (commented):
Has potential!

packages/core/src/lib/history.integration.test.ts
matejchalk: > I would almost say it is okay to use ! here because the commit happened in the previous command. WDYT? cc

Totally agree. The `no-non-null-assertion` rule is turned off in tests files for this reason.

packages/core/src/lib/history.ts
BioPhoton: I thought from-to is a range 🤷‍♂️

packages/core/src/lib/history.integration.test.ts
```diff
@@ -0,0 +1,94 @@
+import { mkdir, rm, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { type SimpleGit, simpleGit } from 'simple-git';
+import { afterAll, beforeAll, describe, expect } from 'vitest';
+import { getHashes } from './history';
+
+// we need a separate folder that is not cleaned in `global-setup.ts`, otherwise the tests can't execute in parallel
+const gitTestFolder = 'core-history-git-test';
```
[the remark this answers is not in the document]
Tlacenka: Once this is merged, please move the folder under `tmp` as well.
#547

packages/core/src/lib/history.ts
BioPhoton: The bad stuff about `as unknown as HistoryOptions` is it hardly break in the future :D 

Thx for spotting!

packages/core/src/lib/history.ts
BioPhoton: Even if the typing was on porpose I understand the handling of booleans now different and would also always stick to the above.

packages/core/src/lib/history.ts
BioPhoton: good input! Changed it already!

packages/core/src/lib/history.ts
BioPhoton: It's a leftover as I had the upload property of the core config destructed. Thx for catching it!

packages/core/src/lib/history.ts
BioPhoton: upload it self could be undefined as it is optional in core config. I will adjust the typing of `HistoryCliOptions` to be in line with CoreConfig.

BioPhoton: Done! I also adjusted the log message accordingly.

packages/core/src/lib/history.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: Done!

packages/core/src/lib/history.ts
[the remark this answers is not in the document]
BioPhoton: You are right! I also spotted it and commented here: https://github.com/code-pushup/cli/pull/541#discussion_r1516678297

packages/core/src/lib/history.unit.test.ts
BioPhoton: I used real data and implemented the full object

packages/core/src/lib/history.ts
matejchalk: It is a range, it's just that you don't need specify both ends.

image

BioPhoton: I `Pick<LogOption, 'maxCount' | 'from' | 'to'>` and will move validation into cli.

BioPhoton: Thx for the examples! did not know about [--since](https://git-scm.com/docs/git-log#Documentation/git-log.txt---sinceltdategt)

packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export async function getHashes(
+  options: LogOptions,
+  git = simpleGit(),
+): Promise<string[]> {
+  const { from, to } = options;
+
+  // validate from & to
+  if (to && (from === '' || from == null)) {
```
matejchalk, line 72:
```suggestion
  if (to && !from) {
```

packages/core/src/lib/history.unit.test.ts
matejchalk: @BioPhoton 

> There are several more occurrences in the file.

:top: I don't want to pollute PRs by commenting on every single occurrence.

packages/core/src/lib/history.ts
matejchalk: Ah, I misunderstood what `from` and `to` mean in `simple-git`'s `log` method. They're not equivalent to `since` and `until`, but they're the positional args in `git log` (`git log <from>`, `git log <from>..<to>`) :man_facepalming:

In that case, the updated validation makes sense :+1:

matejchalk (approved):

packages/core/src/lib/history.ts
Tlacenka: Shouldn't both parameters be optional?

packages/core/src/lib/history.ts
```diff
@@ -0,0 +1,87 @@
...
+    throw new Error('from has to be defined if to is defined');
```
Tlacenka, line 73:
```suggestion
    throw new Error('git log command needs the "from" option defined to accept the "to" option.');
```

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    };
+    await history(historyOptions, ['abc']);
+
+    expect(upload).toHaveBeenCalledTimes(1);
+    expect(upload).toHaveBeenNthCalledWith(
+      1,
+      expect.objectContaining({
+        persist: expect.objectContaining({ filename: 'abc-report' }),
+      }),
+    );
```
Tlacenka, line 88: When it is already established there was only one call, the Nth is unnecessary.
```suggestion
    expect(upload).toHaveBeenCalledTimes(1);
    expect(upload).toHaveBeenCalledWith(
      expect.objectContaining({
        persist: expect.objectContaining({ filename: 'abc-report' }),
      }),
    );
```

packages/core/src/lib/history.unit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        total: 2,
+        latest: {
+          hash: '22287eb716a84f82b5d59e7238ffcae7147f707a',
+          date: 'Thu Mar 7 20:13:33 2024 +0100',
+          message:
+            'test: replace default (buggy on Windows) with basic reporter',
+          refs: 'string',
+          body: 'string',
+          author_name: 'string',
+          author_email: 'string',
```
Tlacenka, line 152: The test data could use some love. The aim is to simulate realistic scenarios.

```suggestion
            hash: '22287eb716a84f82b5d59e7238ffcae7147f707a',
            date: 'Thu Mar 7 20:13:33 2024 +0100',
            message:
              'test: change test reported to basic in order to work on Windows',
            refs: 'string',
            body: '',
            author_name: 'John Doe',
            author_email: 'john.doe@gmail.com',
          },
          {
            hash: '111b284e48ddf464a498dcf22426a9ce65e2c01c',
            date: 'Thu Mar 7 20:13:34 2024 +0100',
            message: 'chore: exclude fixtures from ESLint',
            refs: 'string',
            body: '',
            author_name: 'Jane Doe',
            author_email: 'jane.doe@gmail.com',
          },
        ],
        total: 2,
        latest: {
          hash: '22287eb716a84f82b5d59e7238ffcae7147f707a',
          date: 'Thu Mar 7 20:13:33 2024 +0100',
          message:
            'test: change test reported to basic in order to work on Windows',
          refs: 'string',
          body: '',
          author_name: 'John Doe',
          author_email: 'john.doe@gmail.com',
```

packages/core/src/lib/history.integration.test.ts
```diff
@@ -0,0 +1,99 @@
+import { mkdir, rm, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { type SimpleGit, simpleGit } from 'simple-git';
+import { afterAll, beforeAll, describe, expect } from 'vitest';
+import { getHashes } from './history';
+
+describe('getHashes', () => {
+  const baseDir = join(process.cwd(), 'tmp', 'core-history-git-test');
```
Tlacenka, line 8: Thank you 🙏

Tlacenka (commented):
A few minor suggestions, otherwise looking good!

packages/core/src/lib/history.unit.test.ts
BioPhoton: Thx for the love ❤️

packages/core/src/lib/history.unit.test.ts
BioPhoton: SRY! Resolved now. I also created a better mock object

matejchalk (dismissed):

matejchalk (approved):

PR #543, opened 2024-03-06
refactor(utils): 🐩 fix lint rules in packages/utils/perf

BioPhoton:
This is a dogfooding PR to fix lint errors in the perf folder.

It includes:
- lint rule fixes in `utils/perf`
- removing of turned off lint rules

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -100,10 +100,7 @@
     },
     {
       "files": ["perf/**/*.ts"],
-      "rules": {
-        "no-magic-numbers": "off",
-        "sonarjs/no-duplicate-string": "off"
-      }
+      "rules": {}
     }
   ]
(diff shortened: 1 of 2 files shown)


Tlacenka (changes requested):
While there are issues in the portal, the IDE acts as if there are no ESLint problems and the `lint` is passing.

That is because this is a bug of the ESLint plugin reported in #426 

In general, a `perf` folder is likely to break more rules due to its experimental nature and therefore rather than fixing these rules or having to disable ESLint everywhere, I added the most frequently broken rules in our `.eslintrc` instead.

BioPhoton:
Should we set up a eslint for perf folder with the exceptions? This folder is potentially present in any project and is "standardised".

Issue #482, opened 2024-02-07
CLI compare command (2 report.json files => reports-diff.json)


matejchalk:
Part of epic #478

### Acceptance criteria

Core function:
- [ ] `compare` function is implemented in `@code-pushup/core`
- [ ] accepts two reports objects (source and target)
- [ ] returns serializable object containing audits and categories which have changed
  - [ ] categories and audits are identified by `slug`
  - [ ] category diff includes `score` change
  - [ ] audit diff includes `score` change and `value` change (along with both display values)
CLI wrapper:
- [ ] `compare` command is implemented in `@code-pushup/cli`
- [ ] command accepts 2 key arguments, both are required and specific to this command - path to source `report.json` and path to target `report.json`
- [ ] command handler calls `compare` function from `@code-pushup/core`
- [ ] command handler writes returned object into a `reports-diff.json` file
- [ ] command handler prints message to stdout so user knows where file was created (a `reports-diff.md` will be added in follow-up issue)

PR #544, opened 2024-03-06
feat(cli): include commit info in report.json

matejchalk:
Prerequisite for #482 

Instead of fetching the latest commit for `upload` command and `report.md` formatting, the commit is fetched and included in `report.json` by `collect`.


Tlacenka (approved):
💎 I went through the code. Nice clean-up!
 🧪 I looked at the tests, both unit and integration tests are nicely modified.
Wondering if there is a point in including it in an E2E as it will include variable data - probably enough to have it in the `sorting` integration test report.

PR #548, opened 2024-03-07
test: add type notation in tests

Tlacenka:
Upon discovering one can use a type argument in `toEqual` and `toStrictEqual`, I added these arguments in our tests. They replace the need for `satisfies` in expected test output.

Note: The type argument cannot be used when the output from `expect` is a promise that requires a `resolves` first.

### DIFF
--- a/packages/utils/src/lib/reports/utils.unit.test.ts
+++ b/packages/utils/src/lib/reports/utils.unit.test.ts
@@ -79,7 +79,7 @@ describe('getSortableAuditByRef', () => {
           },
         ],
       ),
-    ).toStrictEqual({
+    ).toStrictEqual<SortableAuditReport>({
       slug: 'function-coverage',
       title: 'Function coverage',
       score: 1,
@@ -172,7 +172,7 @@ describe('getSortableGroupByRef', () => {
           },
         ],
       ),
-    ).toStrictEqual({
+    ).toStrictEqual<SortableGroup>({
       slug: 'code-coverage',
       title: 'Code coverage',
       score: 0.66,
@@ -311,11 +311,9 @@ describe('countWeightedRefs', () => {
 describe('compareIssueSeverity', () => {
   it('should order severities in logically ascending order when used as compareFn with .sort()', () => {
     const severityArr = ['error', 'info', 'warning'] satisfies IssueSeverity[];
-    expect([...severityArr].sort(compareIssueSeverity)).toEqual([
-      'info',
-      'warning',
-      'error',
-    ]);
+    expect([...severityArr].sort(compareIssueSeverity)).toEqual<
+      IssueSeverity[]
+    >(['info', 'warning', 'error']);
   });
 });
 
--- a/packages/utils/src/lib/reports/utils.unit.test.ts
+++ b/packages/utils/src/lib/reports/utils.unit.test.ts
@@ -310,7 +310,7 @@ describe('countWeightedRefs', () => {
 
 describe('compareIssueSeverity', () => {
   it('should order severities in logically ascending order when used as compareFn with .sort()', () => {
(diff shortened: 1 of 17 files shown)


packages/utils/src/lib/reports/utils.unit.test.ts
```diff
@@ -311,11 +311,9 @@ describe('countWeightedRefs', () => {
 describe('compareIssueSeverity', () => {
   it('should order severities in logically ascending order when used as compareFn with .sort()', () => {
     const severityArr = ['error', 'info', 'warning'] satisfies IssueSeverity[];
```
matejchalk, line 313: Not really related, but still :grin: 

```suggestion
    const severityArr: IssueSeverity[] = ['error', 'info', 'warning'];
```

matejchalk (approved):

BioPhoton:
Could that be a linting rule?

packages/utils/src/lib/reports/utils.unit.test.ts
BioPhoton: could that be a linting rule?

BioPhoton (dismissed):
Nice improvmenet!

packages/utils/src/lib/reports/utils.unit.test.ts
Tlacenka: It probably could be but did not find any existing rule.

matejchalk:
@BioPhoton 

> Could that be a linting rule?

In general, lots of things _could_ be linting rules. The more practical question is: _Are_ there any linting rules available in the ecosystem? If not, we would have to implement that lint rule ourselves, in which case the cost-benefit ratio becomes highly questionable.

The current [@code-pushup/eslint-config](https://github.com/code-pushup/eslint-config) contains all the available rules I could find which I thought were beneficial. It's possible I missed something, of course. But mostly I would say that if it hasn't been added already, there's good chance it's because it doesn't exist or I didn't think it was beneficial (an inaccurate linting rule is worse than no linting rule, IMO).

Also, keep in mind that lint rules should be something that's _universally_ a good thing. When it comes to practices that are good in some cases but in other's it's pointless or even counter-productive, then it's very difficult to automate accurately with a lint rule.

Regarding your specific requests:

- a rule to enforce `.toEqual<T>(...)` instead of untyped `.toEqual(...)`:
  - _Is it available?_ - Well, it would have to be specific to Vitest, so the place to look is [eslint-plugin-vitest](https://github.com/veritem/eslint-plugin-vitest?tab=readme-ov-file#rules). I don't see any such rule there (I don't think any of their rules are TS-specific, actually).
  - _Is it universal?_ - I'm not sure. The typed `toEqual` is mostly beneficial because of autocomplete, but the correctness checking is already take care of by the runtime assertion.
- a rule to enforce `const x: T = ...` instead of `const x = ... satisfies T`:
  - _Is it available?_ - Since it's a TypeScript feature, the place to look is [typescript-eslint](https://typescript-eslint.io/rules/). I can't see any rules regarding `satisfies` usage at all (could be because it's still a fairly new feature :man_shrugging:).
  - _Is it universal?_ - Potentially. The reason `satisfies` exists is to be able to do `const x: T1 = ... satisfies T2`, but I can't think of a practical use case for needing to do `const x = ... satisfies T`.

So in both cases, we would have to implement our own lint rule. I don't think either are important enough to prioritize such work though.

packages/utils/src/lib/reports/utils.unit.test.ts
matejchalk: Responded here: https://github.com/code-pushup/cli/pull/548#issuecomment-1985379373

matejchalk (approved):

PR #549, opened 2024-03-08
feat(plugin-lighthouse): add lighthouse runner

BioPhoton:
This PR wraps lighthouse core functionality exported as ESM with a `PluginRunner` for the `@code-pushup/lighthouse-plugin` package.

It aims to leverage all functionality directly over reusing their ESM exports instead of using the CLI.
This is less code to maintain as well as faster to execute and more scalable in terms of process arguments.

Changes included in this PR:
- plugin options mirroring [`runLighthouse`](https://github.com/GoogleChrome/lighthouse/blob/5854c54d5c1bb5ea69a863b82fc65bd08fc1367b/cli/run.js#L215), the core functionality of lighthouse. The following arguments are defined:  `url`, `flags` and `config`.
- a `getRunner` is implemented to get a lighthouse runner receiving the `url`, `flags` and `config` arguments
  - Parts of the functionality of lighthouse CLI are excluded: all commands that don't produce a report, latern data logic. 
    The user is informed about missing implementations if by accident provided over the CLI flags or config object
- additional helper to generate and validate the used `flags` and `config` object
- added `CONTRIBUTING.md` to document setup and debugging of Lighthouse
- setup CI for Lighthouse path configuration on windows

Follow up PR:
- I this refactoring into a new PR https://github.com/code-pushup/cli/issues/552

> [!note]
> There is one thing left, I want to have the env variable only set for Windows in the CI.

**References:**
- I created a docs to describe [debugging of Lighthouse](https://github.com/code-pushup/cli/wiki/Lighthouse-Testing-Guide)

related to #90


packages/plugin-lighthouse/src/lib/constants.ts
```diff
@@ -63,3 +64,21 @@ async function loadLighthouseAudit(
...
 }
+
+export const DEFAULT_CLI_FLAGS: Partial<CliFlags> = {
+  // default values extracted from
+  // https://github.com/GoogleChrome/lighthouse/blob/7d80178c37a1b600ea8f092fc0b098029799a659/cli/cli-flags.js#L80
+  verbose: false,
+  quiet: false,
+  saveAssets: false,
+  // needed to pass CI on linux and windows (locally it works without headless too)
+  chromeFlags: '--headless=new',
```
ChristopherPHolder, line 75: This should be `--headless=shell` 

https://github.com/puppeteer/puppeteer/blob/puppeteer-v22.1.0/docs/api/puppeteer.browserlaunchargumentoptions.md

This was checked by checking the version of lighthouse you have installed and the version of puppeteer they use. 

As users may use a different version of puppeteer it my be different for them. 

The may only have the option to have the boolean or the `new`.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -10,3 +10,15 @@ describe('lighthousePlugin', () => {
     expect(pluginConfig.groups).toHaveLength(5);
   });
 });
+
+describe('getRunner', () => {
+  it('should create and execute runner correctly', async () => {
+    // onlyAudits is used to reduce test time
+    const runner = getRunner('https://example.com', {
+      onlyAudits: ['is-on-https'],
+    });
```
ChristopherPHolder, line 19: It is worth noting that there is a log being hidden from the test because of the console mocking: 

image


As to what this log means, I don't know, but since the typing is not actually extracted from LH itself it my mean that its not accepting something there or there is some config error.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -10,3 +10,15 @@ describe('lighthousePlugin', () => {
...
+    await expect(runner(() => void 0)).resolves.toEqual([
+      expect.objectContaining({ slug: 'is-on-https' }),
+    ]);
+  });
+}, 70_000);
```
matejchalk, line 24: Do we really need such a long timeout? :open_mouth:

BioPhoton: Only on windows, but I'm researching tricks to spin up chromium with a cached website and taking away internet delay or flakiness. 🕵️

My plan is to ship a solution with the e2e test PR that is missing.

BioPhoton: I will close this here and open a issue on the current research and document it. 
I tried to get it into this PR but it seemed not possible to deliver a good solution in this PR.

Feel free to reopen it.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
```diff
@@ -23,6 +50,53 @@ export function lighthousePlugin(
...
+    const flagsWithDefaults = {
+      ...parsedFlags,
+      budgets: budgetsJson,
+    };
+
+    if (precomputedLanternDataPath) {
+      // eslint-disable-next-line no-console
+      console.log(
+        `The parsing precomputedLanternDataPath ${precomputedLanternDataPath} is skipped as not implemented.`,
+      );
```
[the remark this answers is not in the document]
BioPhoton: Nice catch! Seems like I did not merge master recently!

testing/test-setup/src/lib/chrome-path.setup.ts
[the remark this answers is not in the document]
BioPhoton: Nice catch!! This is a left over from my research on getting the testing time down and I missed to delete it.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
BioPhoton: How can I reproduce it?

packages/plugin-lighthouse/README.md
```diff
@@ -1,3 +1,110 @@
...
-TODO: docs
+[![npm](https://img.shields.io/npm/v/%40code-pushup%2Flighthouse-plugin.svg)](https://www.npmjs.com/package/@code-pushup/lighthouse-plugin)
+[![downloads](https://img.shields.io/npm/dm/%40code-pushup%2Flighthouse-plugin)](https://npmtrends.com/@code-pushup/lighthouse-plugin)
+[![dependencies](https://img.shields.io/librariesio/release/npm/%40code-pushup/lighthouse-plugin)](https://www.npmjs.com/package/@code-pushup/lighthouse-plugin?activeTab=dependencies)
+
+🕵️ **Code PushUp plugin for running lighthouse performance tests.** 🔥
+
+---
+
+The plugin parses your lighthouse configuration and lints all audits of the official [lighthouse](https://github.com/GoogleChrome/lighthouse/blob/main/readme.md#lighthouse-------) CLI.
```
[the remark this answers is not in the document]
BioPhoton: got it! Will write Lighthouse correctly

packages/plugin-lighthouse/README.md
```diff
@@ -1,3 +1,110 @@
...
+|               | channel                           |                                                                                                                                                                                                                                                                                                                                                                                                                                  | `string`  | 'cli'       |
+|               | chrome-ignore-default-flags       |                                                                                                                                                                                                                                                                                                                                                                                                                                  | `boolean` | `false`     |
```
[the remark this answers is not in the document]
BioPhoton: Ok!

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -10,3 +10,15 @@ describe('lighthousePlugin', () => {
...
 });
+
+describe('getRunner', () => {
+  it('should create and execute runner correctly', async () => {
+    // onlyAudits is used to reduce test time
+    const runner = getRunner('https://example.com', {
+      onlyAudits: ['is-on-https'],
+    });
+    await expect(runner(() => void 0)).resolves.toEqual([
+      expect.objectContaining({ slug: 'is-on-https' }),
```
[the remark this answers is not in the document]
BioPhoton: Changed report and added numbers. 

BTW, i `used is-on-https` because I thought the test is faster. But it doesn't feel much slower with the new audit...

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
```diff
@@ -1,18 +1,142 @@
...
+        ? undefined
+        : {
+            flags,
+            config,
+            lhr: {
+              audits: {
+                ['cumulative-layout-shift']: {
+                  id: 'cumulative-layout-shift',
+                  title: 'title',
+                  description: 'description',
```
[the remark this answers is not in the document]
BioPhoton: done

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
```diff
@@ -1,18 +1,142 @@
...
+  return {
+    ...actual,
+    runLighthouse: mockRunLighthouse, // Mock the default export if 'lighthouse' is imported as default
+  };
+});
+
+describe('getRunner', () => {
+  it('should return AuditOutputs if executed correctly', async () => {
+    const runner = getRunner('https://localhost:8080');
+    await expect(runner(() => void 0)).resolves.toEqual(
```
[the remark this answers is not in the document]
BioPhoton: It is red in my IDE sry..

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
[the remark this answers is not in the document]
BioPhoton: A added a related taks for the test helper to deal with ASCII color in addition:
https://github.com/code-pushup/cli/issues/552

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
matejchalk: If it's only needed for Windows, then setting the timeout based on OS could be a practical workaround - `platform() === 'win32' ? 70_000 : <something smaller>` :thinking:

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: removed all occurrences

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
```diff
@@ -1,18 +1,142 @@
...
+    );
+
+    expect(setLogLevel).toHaveBeenCalledWith(
+      expect.objectContaining({ verbose: false, quiet: false }),
+    );
+    expect(getBudgets).not.toHaveBeenCalled();
+    expect(getConfig).toHaveBeenCalledWith(expect.objectContaining({}));
+  });
+
+  it('should return consider verbose and quiet flags for logging', async () => {
```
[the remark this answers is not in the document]
BioPhoton: Fixed

packages/plugin-lighthouse/README.md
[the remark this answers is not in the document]
BioPhoton: done. Thanks!

BioPhoton:
I believe I know why the timeout is so high in windows. Get back in a day

packages/plugin-lighthouse/package.json
```diff
@@ -5,6 +5,8 @@
   "dependencies": {
     "@code-pushup/models": "*",
     "lighthouse": "^11.0.0",
-    "@code-pushup/utils": "*"
+    "@code-pushup/utils": "*",
+    "lighthouse-logger": "2.0.1",
```
BioPhoton, line 9: It is possible but might be quite tricky to get rid of this dependency so I keep it in for now.

packages/plugin-lighthouse/README.md
```diff
@@ -1,3 +1,118 @@
...
+[![downloads](https://img.shields.io/npm/dm/%40code-pushup%2Flighthouse-plugin)](https://npmtrends.com/@code-pushup/lighthouse-plugin)
+[![dependencies](https://img.shields.io/librariesio/release/npm/%40code-pushup/lighthouse-plugin)](https://www.npmjs.com/package/@code-pushup/lighthouse-plugin?activeTab=dependencies)
+
+🕵️ **Code PushUp plugin for measuring web performance and quality with Lighthouse.** 🔥
+
+---
+
+The plugin parses your Lighthouse configuration and lints all audits of the official [Lighthouse](https://github.com/GoogleChrome/lighthouse/blob/main/readme.md#lighthouse-------) CLI.
+
+Detected Lighthouse audits are mapped to Code PushUp audits. Audit reports are calculated based on the [original implementation](https://googlechrome.github.io/lighthouse/scorecalc/).
```
matejchalk, line 13: Maybe also good to mention that Lighthouse categories are mapped to Code PushUp groups.

packages/plugin-lighthouse/README.md
```diff
@@ -1,3 +1,118 @@
...
+   yarn add --dev @code-pushup/lighthouse-plugin
+   ```
+
+   ```sh
+   pnpm add --save-dev @code-pushup/lighthouse-plugin
+   ```
+
+3. Add this plugin to the `plugins` array in your Code PushUp CLI config file (e.g. `code-pushup.config.ts`).
+
+   Pass in the path to your ESLint config file, along with glob patterns for which files you wish to target (relative to `process.cwd()`).
```
matejchalk, line 37: This is copy-pasted from ESLint plugin, needs to be replaced.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
```diff
@@ -23,6 +52,58 @@ export function lighthousePlugin(
...
+
+    const flagsWithDefaults = {
+      ...parsedFlags,
+      budgets: budgetsJson,
+      outputPath,
+    };
+
+    if (precomputedLanternDataPath) {
+      ui().logger.info(
+        `The parsing precomputedLanternDataPath "${precomputedLanternDataPath}" is skipped as not implemented.`,
```
matejchalk, line 93:
```suggestion
        `Parsing precomputedLanternDataPath "${precomputedLanternDataPath}" is skipped as not implemented.`,
```

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -587,3 +615,176 @@ describe('toAuditOutputs', () => {
...
+    ).resolves.toBeUndefined();
+  });
+});
+
+describe('getBudgets', () => {
+  it('should return and empty array if no path is specified', async () => {
+    await expect(getBudgets()).resolves.toStrictEqual([]);
+  });
+
+  it('should load budgets from path is specified', async () => {
```
Tlacenka, line 691: 🔡
```suggestion
  it('should load budgets from specified path', async () => {
```

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -587,3 +615,176 @@ describe('toAuditOutputs', () => {
     expect(outputs[0]?.details).toBeUndefined();
   });
 });
+
+describe('getConfig', () => {
+  it('should return undefined if no path is specified', async () => {
+    await expect(getConfig()).resolves.toBeUndefined();
+  });
+
+  it('should load config from lighthouse preset if preset is specified', async () => {
```
Tlacenka, line 624: 🧪 Testing principles: The same function should not be called in one test multiple times.

Suggestion: Splitting it into two tests, one of which could potentially be dynamic.
```
  it.each([
    ['desktop', { formFactor: 'desktop' }],
    ['perf', { onlyCategories: ['performance'] }],
  ] satisfies readonly ['desktop' | 'perf' | 'experimental', object][])(
    'should load config from lighthouse preset if %s preset is specified',
    async (preset, settings) => {
      await expect(getConfig({ preset })).resolves.toEqual(
        expect.objectContaining({
          settings: expect.objectContaining(settings),
        }),
      );
    },
  );
```

Tlacenka (changes requested):
Nice setup.

Unfortunately, I am unable to run integration tests locally. Wondering if it could be related to the hardcoded chrome path.

packages/plugin-lighthouse/README.md
BioPhoton: Very good point!!! Love it!

packages/plugin-lighthouse/README.md
BioPhoton: Damn!! Sry for not reading over more detailed.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -1,12 +1,48 @@
...
+      outputPath:
+        'tmp/plugin-lighthouse/get-runner/no-latern-data/lh-report.json',
+      chromeFlags: ['--headless=shell'],
+    });
+    await expect(runner()).resolves.toBeTruthy();
+    expect(getLogMessages(ui().logger).at(0)).toMatch(
+      'The parsing precomputedLanternDataPath "/path/to/latern-data" is skipped as not implemented.',
+    );
+  });
+}, 30_000);
```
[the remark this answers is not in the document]
BioPhoton: Very true!! Thx!

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -1,12 +1,48 @@
...
+    const pluginConfig = lighthousePlugin('https://www.google.com/');
     expect(() => pluginConfigSchema.parse(pluginConfig)).not.toThrow();
     expect(pluginConfig.audits.length).toBeGreaterThan(100);
     expect(pluginConfig.groups).toHaveLength(5);
   });
 });
+
+describe('getRunner', () => {
+  it('should create and execute runner correctly', async () => {
+    const runner = getRunner('https://www.google.com/', {
```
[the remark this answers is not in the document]
BioPhoton: I believe this is a bug we have in models. Will get back.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: Can u try to run `lighthouse <URL>` directly?

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -587,3 +615,176 @@ describe('toAuditOutputs', () => {
...
+  it('should forward supported entries', () => {
+    expect(validateFlags({ verbose: true })).toStrictEqual({ verbose: true });
+  });
+
+  it('should remove unsupported entries and log', () => {
+    expect(
+      validateFlags({ 'list-all-audits': true, verbose: true } as Flags),
+    ).toStrictEqual({ verbose: true });
+    expect(getLogMessages(ui().logger).at(0)).toBe(
+      '[ blue(info) ] ⚠ The following used flags are not supported: list-all-audits',
```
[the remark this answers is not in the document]
BioPhoton: Good question 😂

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
Tlacenka: When I run `npx lighthouse https://www.google.com` I get a report in a HTML format.
image

packages/plugin-lighthouse/README.md
BioPhoton: done

packages/plugin-lighthouse/README.md
BioPhoton: Done!

packages/plugin-lighthouse/src/index.ts
```diff
@@ -1,3 +1,9 @@
 import { lighthousePlugin } from './lib/lighthouse-plugin';
 
 export default lighthousePlugin;
+export {
+  LIGHTHOUSE_REPORT_NAME,
+  LIGHTHOUSE_PLUGIN_SLUG,
+  AUDITS,
+  GROUPS,
```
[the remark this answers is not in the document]
BioPhoton: Done!

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: Done

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: The `getRunner` returns a `RunnerFunction` which is correct.

```ts
export function getRunner(
  urlUnderTest: string,
  flags: Flags = {},
): RunnerFunction { ... }
```

The problem highlighted in the IDE is related to the `zod` schema `runnerFunctionSchema`. To be more specific to the `onProgressSchema`. For some reason even if we defined the this argument as optional, the IDE highlights is. 

```ts

export const onProgressSchema = z
  .function()
  .args(z.unknown())
  .returns(z.void());
export type OnProgress = z.infer<typeof onProgressSchema>;

export const runnerFunctionSchema = z
  .function()
  .args(onProgressSchema.optional())
```

However as the parsing of the `lighthousePlugin` `PluginConfig` does not throw the code should be correct.

Please let me know if you share that opinion and what we should do to go on.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: Can you try to use `--headless` instead of `--headless=shell`. 

Also we can ask @ChristopherPHolder to support in debugging as he has quite some experience with lighthouse and also owns a Windows machine. I will ping him.

(Sry for the vague suggestion, I know it is time consuming and does not follow a structured approach but more of a trial and error way. Unfrotunately I have no Windows laptop with me)

packages/plugin-lighthouse/src/lib/lighthouse-plugin.ts
```diff
@@ -1,17 +1,46 @@
...
+import {
+  filterAuditsAndGroupsByOnlyOptions,
+  getBudgets,
+  getConfig,
+  setLogLevel,
+  toAuditOutputs,
+  validateFlags,
+} from './utils';
+
+export type Flags = Partial<Omit<LighthouseFlags, 'enableErrorReporting'>>;
```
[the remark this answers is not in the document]
BioPhoton: Good point! renamed it!

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
```diff
@@ -1,18 +1,138 @@
...
+  });
+
+  it('should return configPath', async () => {
+    await getRunner('https://localhost:8080', { configPath: 'lh-config.js' })();
+    expect(getConfig).toHaveBeenCalledWith(
+      expect.objectContaining({ configPath: 'lh-config.js' }),
+    );
+  });
+
+  it('should return budgets', async () => {
```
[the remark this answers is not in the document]
BioPhoton: I renamed the test to:

- should return budgets from the budgets object directly
- should return budgets maintained in the file specified over budgetPath

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
BioPhoton: Done

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: Can I change something to have a more intuitive read of the test?

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
ChristopherPHolder: @Tlacenka the error you have is common, that is happening is that the puppeteer launcher cannot find the chromium or chrome launcher. 

You need to tell lighthouse where the launcher is. 

What OS are you on?

Do you have chrome or chromium installed? Have you installed puppeteer globally at any moment?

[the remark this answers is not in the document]
ChristopherPHolder: I am a bit confused about why it works on the cli but not with the test runner. 

If you can check the OS and the launcher location (chrome.exec) but if not it might be a question of providing a channel.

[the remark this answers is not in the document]
Tlacenka: I am on Windows. I don't think I have puppeteer installed globally. I use Google Chrome so that browser is installed.

Are there any prerequisites for this plugin? If so, they should be added to README. Currently there is only CLI mentioned as a prerequisite.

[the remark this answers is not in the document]
Tlacenka: The executable file is in `Program Files (x86)` so different from the `stubEnv`. And sure enough, after changing the hardcoded path to my path to Chrome, the tests ran successfully.

At the end of the day, integration tests should run in all environments. So I wonder if the test could first check if the file exists or try to find out where the executable is using `which` or similar commands?

packages/plugin-lighthouse/README.md
```diff
@@ -1,3 +1,118 @@
...
+`flags` is the Lighthouse [CLI flags](https://github.com/GoogleChrome/lighthouse/blob/7d80178c37a1b600ea8f092fc0b098029799a659/cli/cli-flags.js#L80) as a JS object.
+
+Within the flags object a couple of other external configuration files can be referenced. E.g. `configPath` , `preset` or `budgetPath` reference external `json` or JavaScript files.
+
+For a complete list the [official documentation of CLI flags](https://github.com/GoogleChrome/lighthouse/blob/main/readme.md#cli-options)
+
+> [!NOTE]  
+> If you are used to work with the Lighthouse CLI you would pass flags like this:
+> `lighthouse https://example.com --output=json --chromeFlags='--headless=shell'`
+>
```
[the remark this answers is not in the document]
Tlacenka: > Also applies to the config argument.

How can I make it more clear that this issue is in multiple places and I am only commenting on one of them? 🙏 
The other was left unaddressed.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
Tlacenka: From what I found online, the reason seems to be that zod currently [does not support optional function arguments](https://github.com/colinhacks/zod/issues/2990).

How to proceed - not sure. The only way around it would currently be to switch to `RunnerConfig` which does not make sense in your use case. cc @matejchalk 

Side-note: Wondering if our progress strategy involving passing `onProgress` all the way is the best solution. This is an OOS discussion though.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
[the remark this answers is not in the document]
Tlacenka: There are no changes done to this part of the file. Am I missing something? 🤔

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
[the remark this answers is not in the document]
Tlacenka: I think that testing that a specified level is correctly displayed belongs in the logger library integration tests. In the usage of the logger elsewhere, one should then assume the levels are correct and only focus on the content. So I would skip the formatting and focus on the values the test is trying to verify.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
matejchalk: Given the lack of good Zod support, I guess the best we can do is pass in `undefined` (or `() => {}`) explicitly to avoid the type errors (`runner()` :arrow_right: `runner(undefined)`).

But I also don't actually understand what the `onProgress` parameter is for. Does it even need to be there?

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -1,12 +1,48 @@
...
   it('should create valid plugin config', () => {
-    const pluginConfig = lighthousePlugin('https://code-pushup-portal.com');
+    const pluginConfig = lighthousePlugin('https://www.google.com/');
     expect(() => pluginConfigSchema.parse(pluginConfig)).not.toThrow();
     expect(pluginConfig.audits.length).toBeGreaterThan(100);
     expect(pluginConfig.groups).toHaveLength(5);
   });
 });
+
+describe('getRunner', () => {
```
Tlacenka, line 16: 🧪 Test strategy: There are two files created by these integration tests. The `tmp/plugin-lighthouse` subfolder should be deleted after these tests are run in a hook to clean up the test outputs.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: We could channel plugin specific progress logs. There are also tests for it to document it.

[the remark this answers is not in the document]
BioPhoton: added undefined

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
BioPhoton: removed them

packages/plugin-lighthouse/README.md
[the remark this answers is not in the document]
BioPhoton: Sry 💩

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: I had cleanup logic in the getMessages helper before but and later when merging changes from main I saw @matejchalk also tested the style and I broke his test. 

So I added it here https://github.com/code-pushup/cli/issues/552. If this gets implemented it will break this test and matej's test and we can decide if we implement it in a way where general styles for the log level and styles in text as bold, or color's in the respective test or if we remove it all together. In any case it is a little change.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: One thing that is open is setting env var in CI only for windows. maybe you have a hint for me.

packages/plugin-lighthouse/CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+We added consider this path in our `beforeAll` hook.
+
+```ts
+beforeEach(() => {
+  try {
+    vi.stubEnv('CHROME_PATH', getChromePath());
+  } catch (e) {
+    const customChromePath = (process.env as { CUSTOM_CHROME_PATH: string }).CUSTOM_CHROME_PATH;
```
matejchalk, line 35: No type-casting is needed, the value of `process.env` properties is `string | undefined`, which you're already safely handling. I guess you came up against the `Property 'CUSTOM_CHROME_PATH' comes from an index signature, so it must be accessed with ['CUSTOM_CHROME_PATH']` TypeScript error? The fix is to use element access.

```suggestion
    const customChromePath = process.env['CUSTOM_CHROME_PATH'];
```

testing/test-setup/src/lib/chrome-path.setup.ts
```diff
@@ -0,0 +1,16 @@
+import { getChromePath } from 'chrome-launcher';
+import * as process from 'node:process';
+import { beforeEach, vi } from 'vitest';
+
+beforeEach(() => {
+  const customChromePath =
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    (process.env as { CUSTOM_CHROME_PATH: string | undefined })
+      .CUSTOM_CHROME_PATH;
```
matejchalk, line 9:
```suggestion
  const customChromePath = process.env['CUSTOM_CHROME_PATH'];
```

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
matejchalk: > One thing that is open is setting env var in CI only for windows. maybe you have a hint for me.

I don't think it's possible to set an environment variable conditionally in the `env:` section in GitHub Actions. But you add a step for it, something like this:

```yml
steps:
  # ...
  - name: Set custom Chrome path for Windows
    if: matrix.os == 'windows-latest'
    run: echo "CUSTOM_CHROME_PATH=..." >> $GITHUB_ENV
  # ...
```

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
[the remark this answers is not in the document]
Tlacenka: I am okay with either. If this is the only thing remaining to merge the PR, I would not wait to change this.
If we go for A, please mention this refactor in the issue 🙏

.github/workflows/code-pushup.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       CP_API_KEY: ${{ secrets.CP_API_KEY }}
       CP_ORGANIZATION: code-pushup
       CP_PROJECT: cli
+      CUSTOM_CHROME_PATH: C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe
```
matejchalk, line 24: The `code-pushup` job only runs on `ubuntu-latest`, so I guess setting this environment variable is a mistake.

packages/plugin-lighthouse/CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+Make sure to install dependencies:
+
+```sh
+npm install
+```
+
+### Chrome path
+
+In this plugin we provide Lighthouse functionality exposed over the `lighthousePlugin`.
+To e2e test lighthouse properly we work with a predefined testing setup.
```
Tlacenka, line 14: 📖 Documentation: There are currently no E2E tests.
```suggestion
To test lighthouse properly we work with a predefined testing setup.
```

packages/plugin-lighthouse/CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Setup
+
+Make sure to install dependencies:
+
+```sh
+npm install
+```
+
+### Chrome path
```
Tlacenka, line 11: I am wondering whether the Chrome path belongs to the `CONTRIBUTING.md` as opposed to the plugin's `README.md`.

If I try to run the plugin locally as a part of my Code PushUp setup, won't I encounter the same issue? 🤔

Tlacenka, line 11: Thank you for taking the time to account for this and document it 🙏 
It works for me now 😁 🎉

.github/workflows/code-pushup.yml
BioPhoton: removed

Tlacenka (dismissed):
Works for me now!

Only a few minor comments left from me, otherwise it is ready to go. 🔥

packages/plugin-lighthouse/src/lib/utils.unit.test.ts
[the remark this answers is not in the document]
BioPhoton: Added it to PR under "Follow up PR"

packages/plugin-lighthouse/CONTRIBUTING.md
BioPhoton: Nice to here!!

BioPhoton: Good question!

I consider it in the PR for the Plugin shell that include docs.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -1,12 +1,57 @@
...
+  const getRunnerTestFolder = join('tmp', 'plugin-lighthouse', 'get-runner');
+
+  afterEach(async () => {
+    await rm(getRunnerTestFolder, { recursive: true, force: true });
+  });
```
Tlacenka, line 23: 🧪 Test outputs : I noticed one more thing - the whole `plugin-lighthouse` folder should be removed after the tests.

I recommend adding an `afterAll` hook for this.
```ts
 afterAll(async () => {
    await rm(join('tmp', 'plugin-lighthouse'), { recursive: true, force: true });
  });
```

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
Tlacenka: Or maybe it is necessary to put the OS in the evaluation construction, so `${{ matrix.os }}`?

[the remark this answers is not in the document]
matejchalk: > Or maybe it is necessary to put the OS in the evaluation construction, so ${{ matrix.os }}?

No, that works correctly. The `if` evaluates as an expression automatically.

image
image
image

The problem is the environment variable isn't set in the next step.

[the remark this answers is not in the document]
Tlacenka: Hmm. I noticed in the error that shell has one `\` backslash. When I run the `echo` command in my PowerShell it keeps two backslashes `\\`. Maybe the formatting of the path is the problem? 🤔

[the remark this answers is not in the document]
Tlacenka: Then again, I use two backslashes in `.env` and integration tests run successfully for me.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
BioPhoton: added

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: So I got it working.

I used:

```yml
- name: Set custom Chrome path for Windows only
  if: matrix.os == 'windows-latest'
  run: |
    echo "CUSTOM_CHROME_PATH=C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe" | Out-File -FilePath $env:GITHUB_ENV -Append -Encoding utf8
  shell: pwsh
```

and debugged with:

```yml
# ...
- name: Log all environment variables
  run: |
    printenv
# ...
```

Also left comments in the actions and `CONTRIBUTING.md` file

matejchalk (approved):

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
matejchalk: Great job debugging this :+1: :star:

matejchalk (dismissed):

matejchalk (dismissed):

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -1,12 +1,63 @@
...
+      outputPath:
+        'tmp/plugin-lighthouse/get-runner/no-latern-data/lh-report.json',
+      chromeFlags: ['--headless=shell'],
+    });
+    await expect(runner(undefined)).resolves.toBeTruthy();
+    expect(getLogMessages(ui().logger).at(0)).toMatch(
+      `Parsing precomputedLanternDataPath "${precomputedLanternDataPath}" is skipped as not implemented.`,
+    );
+  });
+}, 38_000);
```
BioPhoton, line 63: This is set to 38_000 to fix the flakyness. I documented the how I figured that out in the troubleshooting guide.

Tlacenka (changes requested):
The current 38 seconds is way above the threshold for integration tests.

If this is a minimal example (aka we cannot go lower), for now please split runner creation (if that is the fast part) and runner execution into separate tests and `it.skip` the execution ones for now (should be tested on react-todos-app instead of google and transformed into E2E tests later).

My full proposal on how to approach this can be found on [Slack](https://flowup.slack.com/archives/C05HU42QF50/p1711616992972159?thread_ts=1711536931.855229&cid=C05HU42QF50).

matejchalk (dismissed):

CONTRIBUTING.md
```diff
@@ -41,6 +41,10 @@ npx nx affected:lint
 npx nx code-pushup -- collect
 ```
 
+### Testing
+
+To skip long-running tests on your local environment set the following environment variable `SKIP_LONG_TESTS=true`.
```
Tlacenka, line 46: I think the flag should work in the opposite way - as a contributor I should not be required to know about and set the flag if I want a smooth DX - in other words, one should opt in, not opt out.

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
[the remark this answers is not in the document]
BioPhoton: To get better DX I introduced logic that skips long running integration tests locally. 

To enable it set `SKIP_LONG_TESTS=true` in your environment variables. 

I also documented this in the root `CONTRIBUTING.md` file.

CONTRIBUTING.md
BioPhoton: That way we would have to strictly configure it. I would rather prefer to have our CI run by default all tests instead of  a custom configuration that lives in the `action.yml` file.

Is there a way to detect if we are on a local setup?

matejchalk: > Is there a way to detect if we are on a local setup?

The best heuristic for that is to check if `CI` environment variable is defined.

BioPhoton: I could use that.. but then i would mess around with that var locally. 

Should I:
- A) keep as is
- B) refactor to derive the skip decision form `CI` (which would mean i have to set it to true locally if I want to run all tests.. Naming is not really intuitive.)

matejchalk: Let's first agree on what the desired outcome is :sweat_smile: 

1. Should long-running tests run in _CI_?
2. Should long-running tests run _locally by default_?

Current implementation assume the answer to both is _yes_, but @Tlacenka is suggesting something different (at least for local, not sure about CI).

As far as I'm concerned, I'm happy to go with the current implementation and tweak it later if we find it impractical. What do you guys think?

Regarding the implementation, we can create a helper function in `test-utils`, which could check multiple environment variables if needed.

BioPhoton: As a compromise I can implement the helper already and Go with `C`I || `SKIP_LONG_TESTS`, and we polish later.

BioPhoton: Updated the code with the helper function

BioPhoton: Isn't the code already skipping the long tests locally?

matejchalk (dismissed):

BioPhoton:
added a overview section with links :) 
Screenshot 2024-03-28 at 18 56 36

Tlacenka (dismissed):

matejchalk (approved):

.github/workflows/ci.yml
```diff
@@ -105,6 +105,15 @@ jobs:
...
         run: npm ci
+      - name: Set custom Chrome path for Windows only
+        if: matrix.os == 'windows-latest'
+        # This path is considered in `testing/setup/src/lib/chrome-path-setup.ts` and used in different test configurations
+        run: |
+          echo "CUSTOM_CHROME_PATH=C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe" | Out-File -FilePath $env:GITHUB_ENV -Append -Encoding utf8
+        shell: pwsh
+      #- name: Log all environment variables
+      #  run: |
+      #      printenv
```
Tlacenka, line 116: 🧹 This seems like a leftover.

CONTRIBUTING.md
```diff
@@ -12,6 +12,25 @@ Make sure to install dependencies:
...
 ```
 
+## Environment Variables
+
+This table provides a quick overview of the environmental setup, with detailed explanations in the corresponding sections.
+
+| Feature                          | Local Default | CI Default         | Description                                                                                                                   |
+| -------------------------------- | ------------- | ------------------ | ----------------------------------------------------------------------------------------------------------------------------- |
+| `env.INCLUDE_SLOW_TESTS` **❗️** | `false`       | `true`             | Controls inclusion of long-running tests. Overridden by setting. Details in the [Testing](#Testing) section.                  |
+| `env.CUSTOM_CHROM_PATH`          | N/A           | Windows **❗️❗️** | Path to Chrome executable. See [plugin-lighthouse/CONTRIBUTING.md](./packages/plugin-lighthouse/CONTRIBUTING.md#chrome-path). |
```
Tlacenka, line 22: Typo.
```suggestion
| `env.CUSTOM_CHROME_PATH`          | N/A           | Windows **❗️❗️** | Path to Chrome executable. See [plugin-lighthouse/CONTRIBUTING.md](./packages/plugin-lighthouse/CONTRIBUTING.md#chrome-path). |
```

CONTRIBUTING.md
```diff
@@ -41,6 +60,15 @@ npx nx affected:lint
 npx nx code-pushup -- collect
 ```
 
+### Testing
```
Tlacenka, line 63: 📖 Documentation hierarchy: I would not nest testing under development.

```suggestion
## Testing
```

CONTRIBUTING.md
Tlacenka, line 22: 👨‍💻 DX: It is not clear from this that the Chrome dependency is only required for using/contributing to the Lighthouse plugin, not the CLI as a whole.

packages/plugin-lighthouse/CONTRIBUTING.md
```diff
@@ -0,0 +1,126 @@
+# Contributing
+
+## Setup
+
+Make sure to install dependencies:
+
+```sh
+npm install
```
Tlacenka, line 8: 👩‍💻DX: Chrome should be added as a prerequisite here as well.

CONTRIBUTING.md
```diff
@@ -41,6 +60,15 @@ npx nx affected:lint
...
+
+Long-running tests are treated based on the environment to achieve better DX.
```
Tlacenka, line 65:
```suggestion
Some of the plugins have a longer runtime. In order to ensure better DX, longer tests are excluded by default when executing tests locally.
```

CONTRIBUTING.md
```diff
@@ -41,6 +60,15 @@ npx nx affected:lint
...
+
+You can control the execution of long-running tests over the `INCLUDE_SLOW_TESTS` environment variable.
+
+To change this setup got you your root folder and open your `.env` file.
```
Tlacenka, line 69: 🔡
```suggestion
To change this setup, open (or create) the `.env` file in the root folder.
```

packages/plugin-lighthouse/CONTRIBUTING.md
```diff
@@ -0,0 +1,126 @@
...
+- name: Set custom Chrome path for Windows only
+  if: matrix.os == 'windows-latest'
+  run: |
+    echo "CUSTOM_CHROME_PATH=C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe" | Out-File -FilePath $env:GITHUB_ENV -Append -Encoding utf8
+  shell: pwsh
+
+# Optional debug log
+- name: Log all environment variables
+  run: |
+    printenv
```
Tlacenka, line 41: ❔ Does this have to be a multi-line command?
```suggestion
  run: printenv
```

packages/plugin-lighthouse/CONTRIBUTING.md
```diff
@@ -0,0 +1,126 @@
...
+To reset the above just delete the folder and apply the settings again.
+
+_A helpful chromium setup is preconfigured with the following settings:_
+
+- A user profile is set up. This enables certain debugging related options as well as help to visually distinguish between test setups as the header bar is colored.
+  <img width="1202" alt="chrome-settings-manage-profile" src="./docs/images/chrome-settings-manage-profile.png">
+
+#### Resources
+
+- https://www.chromium.org/developers/how-tos/run-chromium-with-flags/
```
Tlacenka, line 102: 🎨 Markdown
```suggestion
- [chromium flags guide](https://www.chromium.org/developers/how-tos/run-chromium-with-flags/)
```

packages/plugin-lighthouse/CONTRIBUTING.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Troubleshooting
+
+1. Verify Chrome Installation  
+   Ensure Chrome is correctly installed and accessible to the Lighthouse process.
+   Run `npx chrome-debug` to test it. Read further under [chrome-path](#chrome-path)
+
+2. Increase Timeout  
+   In **many cases** it is just because of Lighthouse's flaky time of runs.
+   **Try increase the timeout** provide longer processing times.
```
Tlacenka, line 112:
```suggestion
2. Increase Timeout  
   Lighthouse has a longer runtime which can time out in different environments.
   **Try increasing the test timeout** in `lighthouse-plugin.integration.test.ts` for `runner creation and execution` test suite.
```

packages/plugin-lighthouse/src/lib/lighthouse-plugin.integration.test.ts
```diff
@@ -1,12 +1,76 @@
...
   it('should create valid plugin config', () => {
-    const pluginConfig = lighthousePlugin('https://code-pushup-portal.com');
+    const pluginConfig = lighthousePlugin('https://www.google.com/');
     expect(() => pluginConfigSchema.parse(pluginConfig)).not.toThrow();
     expect(pluginConfig.audits.length).toBeGreaterThan(100);
     expect(pluginConfig.groups).toHaveLength(5);
   });
 });
+
+describe('createRunnerFunction', () => {
```
Tlacenka, line 23: 🧪 Test maintainability: Both the creation and execution of the runner function is tested here.
```suggestion
describe('runner creation and execution', () => {
```

packages/plugin-lighthouse/CONTRIBUTING.md
```diff
@@ -0,0 +1,126 @@
...
+
+3. Turn on debug mode  
+   Show debug logs of Lighthouse. Set the following environment variable: `DEBUG='*'`
+
+4. Understand error messages (⏳ could also be because of timeout problems :D )
```
Tlacenka, line 117: Nice!

packages/plugin-lighthouse/CONTRIBUTING.md
BioPhoton: I did not do that on purpose. The packages inside this repo should include Chromium. 

Could you please check the stup and tell me how it could happen that the packages did not install chromium? 
If this is really the case we should put the into the `packages.json`.

CONTRIBUTING.md
BioPhoton: The path is needed for all plugin users. In general there are no "unreolved" problems in setup and usage. But as we saw there are. I would add a bit of text to the plugin readme to clarify the setup ad link to the trouble shooting section.

.github/workflows/ci.yml
BioPhoton: I kept it for the next one that needs to debug it. It is also mentioned in the docs.

packages/plugin-lighthouse/CONTRIBUTING.md
Tlacenka: Where can I find the dependency that should include chromium? In `package.json`, I do not see Chromium.

CONTRIBUTING.md
Tlacenka: If I install CLI and only use ESLint and coverage plugins, I should not need to worry about any Chrome path. This dependency is (I assume) tied to Lighthouse only at the moment. Am I missing something?

packages/plugin-lighthouse/CONTRIBUTING.md
BioPhoton: when you install lighthouse is has to run on something... what got installed on your end?
Also, can we clarify this in slack and merge it?

CONTRIBUTING.md
BioPhoton: > If I install CLI and only use ESLint and coverage plugins, I should not need to worry about any Chrome path.
Yes. therefore in a `CONTRIBUTING.md`
> This dependency is (I assume) tied to Lighthouse only at the moment.
Yes therefor it is inside the `package.json` of `plugin-lighthouse`

> Am I missing something?
I guess no. But what should I change now?
It is inside a contributor related file. I added lighthouse only. Hope that fixes it.

packages/plugin-lighthouse/CONTRIBUTING.md
Tlacenka: I have Chrome installed in a different location than the hardcoded one. That is why it did not work for me.

But my original point was something else. If we only provide Lighthouse but the browser it runs on **needs to be installed separately** by the user, it should be listed as a **prerequisite** for using this plugin in its documentation. If we install a browser and run Lighthouse on it for the user, then the documentation is okay as is.

CONTRIBUTING.md
Tlacenka: My point was to explicitly add to the root `CONTRIBUTING` that that environment variable is only applicable when using or contributing to **Lighthouse** plugin. I unfortunately cannot see any "Lighthouse only" but I can see a link to CONTRIBUTING of Lighthouse, could be enough.

Tlacenka (approved):

Tlacenka (approved):

Tlacenka (approved):

PR #550, opened 2024-03-08
feat(core): implement compare reports functions

matejchalk:
First half of #482 - all the underlying logic, unit tests and models.

Next PR will be the CLI integration and E2E tests.


packages/core/src/lib/implementation/compare-scorables.unit.test.ts
```diff
@@ -0,0 +1,320 @@
...
+                  title: 'Function coverage',
+                  score: 0.8,
+                  value: 80,
+                  displayValue: '80 %',
+                },
+              ],
+            },
+          ],
+        },
+      } as ReportsToCompare),
```
BioPhoton, line 268: `satisfies` over `as`?

vmasek (approved):

packages/core/src/lib/implementation/compare-scorables.unit.test.ts
matejchalk: It wouldn't compile with `satisfies`, I'm actually intentionally leaving out most `Report` properties because they're not relevant to the test scenario. Using `as` allows me to omit required props, while still giving me autocomplete and some basic type checking - TypeScript treats it as kind of a recursive partial.

packages/utils/src/lib/reports/flatten-plugins.ts
```diff
@@ -0,0 +1,24 @@
+import { Report } from '@code-pushup/models';
+
+// generic type params infers ScoredGroup if ScoredReport provided
+export function listGroupsFromAllPlugins<T extends Report>(
+  report: T,
+): {
+  plugin: T['plugins'][0];
```
BioPhoton, line 7:
```suggestion
  plugin: T['plugins'][number];
```

packages/utils/src/lib/reports/flatten-plugins.ts
```diff
@@ -0,0 +1,24 @@
...
+  group: NonNullable<T['plugins'][0]['groups']>[0];
```
BioPhoton, line 8:
```suggestion
  group: NonNullable<T['plugins'][number]['groups']>[number];
```

packages/utils/src/lib/reports/flatten-plugins.ts
```diff
@@ -0,0 +1,24 @@
...
+}[] {
+  return report.plugins.flatMap(
+    plugin => plugin.groups?.map(group => ({ plugin, group })) ?? [],
+  );
+}
+
+export function listAuditsFromAllPlugins<T extends Report>(
+  report: T,
+): {
+  plugin: T['plugins'][0];
```
BioPhoton, line 18:
```suggestion
  plugin: T['plugins'][number];
```

packages/utils/src/lib/reports/flatten-plugins.ts
```diff
@@ -0,0 +1,24 @@
...
+  audit: T['plugins'][0]['audits'][0];
```
BioPhoton, line 19:
```suggestion
  audit: T['plugins'][number]['audits'][number];
```

BioPhoton (changes requested):
Amazing feature! Can't wait to see it in action!

packages/core/src/lib/compare.ts
```diff
@@ -0,0 +1,65 @@
...
+    groups,
+    audits,
+    packageName,
+    version,
+    date,
+    duration,
+  };
+}
+
+export async function compareReportFiles(
```
Tlacenka, line 49: Nitpick: In order to read this file from top to bottom, this function should be above `compareReports` which is called within it.

packages/core/src/lib/compare.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        REPORT_MOCK.categories.length,
+      );
+      expect(reportsDiff.categories.removed).toHaveLength(0);
+      expect(reportsDiff.categories.changed).toHaveLength(0);
+      expect(reportsDiff.categories.unchanged).toHaveLength(0);
+    });
+
+    it('should only have added groups (minimal report has none)', () => {
+      const reportsDiff = compareReports(mockReports);
+      expect(reportsDiff.groups.added.length).toBeGreaterThan(0);
```
Tlacenka, line 93: Why not assert the actual value as calculated above for the unchanged?

packages/core/src/lib/compare.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      expect(reportsDiff.groups.removed).toHaveLength(0);
+      expect(reportsDiff.groups.changed).toHaveLength(0);
+      expect(reportsDiff.groups.unchanged).toHaveLength(0);
+    });
+
+    it('should mark multiple audits as added and 1 as removed (single audit from minimal report unmatched)', () => {
+      const reportsDiff = compareReports(mockReports);
+      expect(reportsDiff.audits.added.length).toBeGreaterThan(1);
+      expect(reportsDiff.audits.removed).toHaveLength(1);
```
Tlacenka, line 102: I would at least check its `slug`. Again, static mock, so we know which one it is.

packages/utils/src/lib/diff.ts
```diff
@@ -0,0 +1,59 @@
+export type Diff<T> = {
+  before: T;
+  after: T;
+};
+
+export function matchArrayItemsByKey<T>({
+  before,
+  after,
+  key,
+}: Diff<T[]> & { key: keyof T | ((item: T) => unknown) }) {
```
Tlacenka, line 10: By assigning these items `T[]`, what happens if they are actually undefined (groups, categories)? The `find` and other functions act as if they are always defined, or am I missing something?

packages/core/src/lib/implementation/compare-scorables.ts
```diff
@@ -0,0 +1,194 @@
...
+  return {
+    slug: audit.slug,
+    title: audit.title,
+    plugin: {
+      slug: plugin.slug,
+      title: plugin.title,
+    },
+    score: audit.score,
+    value: audit.value,
+    displayValue: audit.displayValue,
```
Tlacenka, line 164: Shouldn't this be optional? If the `displayValue` is `undefined`, the property shouldn't be there no?

packages/core/src/lib/implementation/compare-scorables.ts
```diff
@@ -0,0 +1,194 @@
...
+      before: before.audit.score,
+      after: after.audit.score,
+      diff: after.audit.score - before.audit.score,
+    },
+    values: {
+      before: before.audit.value,
+      after: after.audit.value,
+      diff: after.audit.value - before.audit.value,
+    },
+    displayValues: {
```
Tlacenka, line 189: Same here.

packages/core/src/lib/implementation/compare-scorables.unit.test.ts
```diff
@@ -0,0 +1,320 @@
...
+                { slug: 'suggestions', title: 'Suggestions', score: 0.4 }, // regressed
+                { slug: 'formatting', title: 'Formatting', score: 0.95 }, // improved
+              ],
+            },
+            {
+              slug: 'stylelint',
+              title: 'Stylelint',
+              groups: [
+                { slug: 'errors', title: 'Avoid errors', score: 1 },
+                { slug: 'conventions', title: 'Conventions', score: 0.7 },
```
Tlacenka, line 98: As a personal preference, I would minimise the size of test data when the same result in logic is expected in multiple places (e.g. here for `errors` and `conventions`). It simplifies debugging.

I understand you wanted to provide realistic scenarios, just wondering if all cases couldn't have been tested using two plugins instead of three, so that the data fits on the screen without scrolling.

Tlacenka (commented):
Overall very well designed feature.

The only question mark I have is handling optional properties (groups and potentially categories, should we decide to keep them optional throughout the CLI).

packages/models/src/lib/reports-diff.ts
```diff
@@ -0,0 +1,133 @@
...
+  scoreSchema,
+  slugSchema,
+  titleSchema,
+} from './implementation/schemas';
+import { pluginMetaSchema } from './plugin-config';
+
+function makeComparisonSchema<T extends ZodTypeAny>(schema: T) {
+  const sharedDescription = schema.description || 'Result';
+  return z.object({
+    before: schema.describe(`${sharedDescription} (previous commit)`),
```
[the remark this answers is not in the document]
matejchalk: Done, used _source commit_ and _target commit_.

packages/models/src/lib/reports-diff.ts
```diff
@@ -0,0 +1,133 @@
...
+);
+
+export type CategoryDiff = z.infer<typeof categoryDiffSchema>;
+export type GroupDiff = z.infer<typeof groupDiffSchema>;
+export type AuditDiff = z.infer<typeof auditDiffSchema>;
+export type CategoryResult = z.infer<typeof categoryResultSchema>;
+export type GroupResult = z.infer<typeof groupResultSchema>;
+export type AuditResult = z.infer<typeof auditResultSchema>;
+
+export const reportsDiffSchema = z
```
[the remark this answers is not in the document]
matejchalk: > Can this be a more generic schema?

Extracted to helper function.

> Is it ever valid for all the arrays to be empty?

Could be, e.g. when there are no categories. Keep in my mind that only we will probably create this `report-diff.json` (same as `report.json`), so I woulnd't be too concerned with making it convenient to write.

packages/core/src/lib/compare.ts
matejchalk: Re-ordered, also in unit tests.

packages/utils/src/lib/diff.ts
matejchalk: The function doesn't allow passing in an non-array value. The caller will be forced by TypeScript to do something like `?? []` when they pass in the argument.

packages/core/src/lib/implementation/compare-scorables.ts
matejchalk: It doesn't really matter. It just serialized as JSON in the end, so the end result will be the same - `undefined` isn't supported by JSON format, so `JSON.stringify` will omit such properties.

packages/core/src/lib/implementation/compare-scorables.ts
matejchalk: Same answer as here: https://github.com/code-pushup/cli/pull/550#discussion_r1519929655

packages/core/src/lib/implementation/compare-scorables.unit.test.ts
matejchalk: OK, I omitted the Stylelint plugin :cry:

packages/core/src/lib/compare.unit.test.ts
Tlacenka: Right. So if you are only concerned if there are any such entities present, why single out this instance where you expect one - then the question becomes, how do you know if it is the right one? If you only care that there is at least one in order not to disrupt the test data, I still think `.tobeGreaterThan(0)` is the correct approach here. `toHaveLength(N)` is neither here or there.

Tlacenka: I missed the commit where you addressed this so feel free to ignore, this is outdated, sorry 😅

packages/models/src/lib/reports-diff.ts
```diff
@@ -0,0 +1,139 @@
...
+  const sharedDescription = schema.description || 'Result';
+  return z.object({
+    before: schema.describe(`${sharedDescription} (source commit)`),
+    after: schema.describe(`${sharedDescription} (target commit)`),
+  });
+}
+
+function makeArraysComparisonSchema<
+  TDiff extends typeof scorableDiffSchema,
+  TResult extends ZodTypeAny,
```
Tlacenka, line 27: Question - do we lose any type safety by having this as `any` compared to how it was before?

Tlacenka (approved):
Well done 👏

packages/models/src/lib/reports-diff.ts
matejchalk: No, because it's a generic type param, the return type is inferred based on the actual parameter type. The `extends ZodTypeAny` is just an extra constraint to ensure the parameter type is some Zod schema.

Tlacenka: Thank you for clarifying.

BioPhoton (approved):

PR #554, opened 2024-03-10
refactor: adjust only plugins logic

BioPhoton:
This PR includes small changes to types of the CLI onlyPlugins internal functionality.
It replaces parts of the existing logic with helpers form `utils` and adds tests.

Changes included:
- introduce types for the plugin middleware
- use helper
- add tests


packages/cli/src/lib/implementation/only-plugins.utils.unit.test.ts
matejchalk, line 1: In the original `validateOnlyPluginsOption` tests, the input data was minimized to only what's relevant by cleverly using `as`. So the input could be simply a one-liner `[{ slug: 'plugin1' }, { slug: 'plugin2' }]` instead of ~25 lines of input data. The signal-to-noise ratio is not good now :disappointed: 

image

packages/cli/src/lib/implementation/only-plugins.middleware.unit.test.ts
matejchalk, line 1: The original unit tests managed 6 test cases in ~90 LOC. The refactored tests only manage 3 tests cases and take up ~70 LOC. So again, the signal to noise ratio got worse.

Also, let's look at how the test coverage survived the refactor:

| test case | covered by original | covered by refactor |
| :-- | :-: | :-: |
| _should return all plugins if no onlyPlugins option is provided_ (\*) | ✅ | ❌ |
| _should return only plugins with matching slugs_ | ✅ | ✅ |
| _should return all categories if no onlyPlugins option_ (\*) | ✅ | ❌ |
| _should return categories containing only defined plugins_ |  ✅ | ✅ |
| _should print ignored category and its first violating plugin_ | ✅ | ❌ |
| _should return empty array for no categories_ (\*) | ✅ | ❌ |

(\*) I don't consider these covered by new _should forward equal values if not set_ test case, because it has empty `plugins` and `categories` as inputs, so it's hardly testing something _wasn't filtered_ when there's nothing be filtered anyway.

matejchalk (changes requested):
The code refactor is fine, but the test refactor is not good - see comments.

packages/cli/src/lib/implementation/only-plugins.middleware.ts
```diff
@@ -1,23 +1,25 @@
...
+    const { plugins, categories = [], onlyPlugins } = processArgs;
 
-  return {
-    ...processArgs,
-    plugins: filterPluginsBySlug(processArgs.plugins, processArgs),
-    categories: filterCategoryByPluginSlug(processArgs.categories, processArgs),
-  };
+    validateOnlyPluginsOption(plugins, processArgs);
+
+    const onlyPluginsSet = new Set(toArray(onlyPlugins));
```
Tlacenka, line 13: Why is there `toArray`? The option type is `string[]`.

packages/cli/src/lib/implementation/only-plugins.middleware.unit.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        plugins: [],
+        categories: [],
+      }),
+    ).toStrictEqual({
+      plugins: [],
+      categories: [],
+    });
+  });
+
+  it('should filter plugins', () => {
```
Tlacenka, line 28: The test case descriptions are very vague. Is this a case where all plugins are filtered out, some, or none? These should be in dedicated test cases.

Tlacenka (changes requested):
Fully agree with the comments above.

packages/cli/src/lib/implementation/only-plugins.middleware.ts
BioPhoton: removed

packages/cli/src/lib/implementation/only-plugins.middleware.unit.test.ts
BioPhoton: added proper description

packages/cli/src/lib/implementation/only-plugins.middleware.unit.test.ts
BioPhoton: I added mock data instead of empty arrays. 
Therefore I consider:
- _should return all plugins if no onlyPlugins option is provided_
- _should return all categories if no onlyPlugins option_

I added a test for _should return empty array for no categories_:
- _should fill undefined categories with empty array_

packages/cli/src/lib/implementation/only-plugins.utils.unit.test.ts
BioPhoton: implemented it. Thank a lot for the hint! Super clean now. Really way better.

packages/cli/src/lib/implementation/only-plugins.utils.ts
```diff
@@ -1,65 +1,27 @@
...
-    ? onlyPlugins.filter(plugin => !plugins.some(({ slug }) => slug === plugin))
-    : [];
+  const onlyPluginsSet = new Set(onlyPlugins);
+  const missingPlugins = onlyPlugins.filter(
+    plugin => !plugins.some(({ slug }) => slug === plugin),
+  );
 
   if (missingPlugins.length > 0 && verbose) {
-    ui().logger.warning(
+    ui().logger.info(
```
matejchalk, line 24: Shouldn't this be a warning? It's a mistake by the user.

packages/cli/src/lib/implementation/only-plugins.utils.ts
```diff
@@ -69,4 +31,19 @@ export function validateOnlyPluginsOption(
...
+      ({ plugin }) => !onlyPluginsSet.has(plugin),
+    ).map(({ slug }) => slug);
+    ui().logger.info(
+      `${chalk.yellow(
+        '⚠',
+      )} The --onlyPlugin argument removed categories with "${removedCategorieSlugs.join(
+        '", "',
+      )}" slugs.
+    `,
+    );
```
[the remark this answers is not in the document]
BioPhoton: It always is confusing for me... So I added it.
Will remove it

```suggestion
    ui().logger.info(
      `The --onlyPlugin argument removed categories with "${removedCategorieSlugs.join(
        '", "',
      )}" slugs.
    `,
    );
```

packages/cli/src/lib/implementation/only-plugins.utils.ts
BioPhoton: This was on purpose. loggin warnings logs into `stderr` instead of `stdout`. Changed it on multiple places.

matejchalk: Warnings logged to stderr is a good thing, though.

matejchalk (approved):

packages/cli/src/lib/implementation/only-plugins.utils.ts
matejchalk: But that's out of scope of this PR.

Tlacenka:
Why was this merged with requested changes? There are still issues with this PR.
- A newly added test for "filling undefined categories with empty array" has nothing to do with the middleware functionality. It is an unwanted side-effect of incorrectly handled optional categories. So creating a test for it is not desired.
- The test descriptions are still vague. "should filter categories for slug "p1"' says nothing to someone who didn't previously implement this option. Filter how? What should happen to categories that have both plugins that are mentioned in onlyPlugins and plugins that are not? This should be the place for someone to learn about the functionality and its use cases.
- Some test descriptions look like they have not even been proof-read ("should filter plugins plugins for slug "p1"")
- I requested adding coverage to cover the following cases: Some plugins are filtered out, all (to check it returns empty array) and none (to check it returns identical array). Currently, only one out of three cases is covered for both plugins and categories.

BioPhoton:
> A newly added test for "filling undefined categories with empty array" has nothing to do with the middleware functionality. It is an unwanted side-effect of incorrectly handled optional categories. So creating a test for it is not desired.

I created [this](https://github.com/code-pushup/cli/wiki/Data-and-Type's-Strategy) for discussion.'''

> The test descriptions are still vague. "should filter categories for slug "p1"' says nothing to someone who didn't previously implement this option. Filter how? What should happen to categories that have both plugins that are mentioned in onlyPlugins and plugins that are not? This should be the place for someone to learn about the functionality and its use cases.

Getting better in testing is an incremental process, at least when it comes to non trackable things.
If you could write some suggestions that will speed up the situation currently and allows a faster & motivating learning.

> I requested adding coverage to cover the following cases: Some plugins are filtered out, all (to check it returns empty array) and none (to check it returns identical array). Currently, only one out of three cases is covered for both plugins and categories.

Please list me the 3 describe blocks and i will add them to this issue here https://github.com/code-pushup/cli/issues/568

PR #556, opened 2024-03-11
feat(cli): implement compare command

### LINKED ISSUE #482: CLI compare command (2 report.json files => reports-diff.json)

matejchalk:
Closes #482

### DIFF
--- a/e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
+++ b/e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
@@ -1,0 +1,705 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`CLI compare > should compare report.json files and create report-diff.json 1`] = `
+{
+  "audits": {
+    "added": [],
+    "changed": [
+      {
+        "displayValues": {
+          "after": "62 %",
+          "before": "62 %",
+        },
+        "plugin": {
+          "slug": "coverage",
+          "title": "Code coverage",
+        },
+        "scores": {
+          "after": 0.617,
+          "before": 0.6211,
+          "diff": -0.0040999999999999925,
+        },
+        "slug": "line-coverage",
+        "title": "Line coverage",
+        "values": {
+          "after": 62,
+          "before": 62,
+          "diff": 0,
+        },
+      },
+      {
+        "displayValues": {
+          "after": "passed",
+          "before": "1 warning",
+        },
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "scores": {
+          "after": 1,
+          "before": 0,
+          "diff": 1,
+        },
+        "slug": "arrow-body-style",
+        "title": "Require braces around arrow function bodies",
+        "values": {
+          "after": 0,
+          "before": 1,
+          "diff": -1,
+        },
+      },
+      {
+        "displayValues": {
+          "after": "passed",
+          "before": "3 warnings",
+        },
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "scores": {
+          "after": 1,
+          "before": 0,
+          "diff": 1,
+        },
+        "slug": "object-shorthand",
+        "title": "Require or disallow method and property shorthand syntax for object literals",
+        "values": {
+          "after": 0,
+          "before": 3,
+          "diff": -3,
+        },
+      },
+      {
+        "displayValues": {
+          "after": "passed",
+          "before": "1 warning",
+        },
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "scores": {
+          "after": 1,
+          "before": 0,
+          "diff": 1,
+        },
+        "slug": "prefer-const",
+        "title": "Require \`const\` declarations for variables that are never reassigned after declared",
+        "values": {
+          "after": 0,
+          "before": 1,
+          "diff": -1,
+        },
+      },
+    ],
+    "removed": [],
+    "unchanged": [
+      {
+        "displayValue": "50 %",
+        "plugin": {
+          "slug": "coverage",
+          "title": "Code coverage",
+        },
+        "score": 0.5,
+        "slug": "function-coverage",
+        "title": "Function coverage",
+        "value": 50,
+      },
+      {
+        "displayValue": "88 %",
+        "plugin": {
+          "slug": "coverage",
+          "title": "Code coverage",
+        },
+        "score": 0.875,
+        "slug": "branch-coverage",
+        "title": "Branch coverage",
+        "value": 88,
+      },
+      {
+        "displayValue": "passed",
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "score": 1,
+        "slug": "no-cond-assign",
+        "title": "Disallow assignment operators in conditional expressions",
+        "value": 0,
+      },
+      {
+        "displayValue": "passed",
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "score": 1,
+        "slug": "no-const-assign",
+        "title": "Disallow reassigning \`const\` variables",
+        "value": 0,
+      },
+      {
+        "displayValue": "passed",
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "score": 1,
+        "slug": "no-debugger",
+        "title": "Disallow the use of \`debugger\`",
+        "value": 0,
+      },
+      {
+        "displayValue": "passed",
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "score": 1,
+        "slug": "no-invalid-regexp",
+        "title": "Disallow invalid regular expression strings in \`RegExp\` constructors",
+        "value": 0,
+      },
+      {
+        "displayValue": "passed",
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "score": 1,
+        "slug": "no-undef",
+        "title": "Disallow the use of undeclared variables unless mentioned in \`/*global */\` comments",
+        "value": 0,
+      },
+      {
+        "displayValue": "passed",
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "score": 1,
+        "slug": "no-unreachable-loop",
+        "title": "Disallow loops with a body that allows only one iteration",
+        "value": 0,
+      },
+      {
+        "displayValue": "passed",
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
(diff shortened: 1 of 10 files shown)


packages/cli/src/lib/compare/compare-command.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { CompareOptions } from '../implementation/compare.model';
+import { yargsCompareOptionsDefinition } from '../implementation/compare.options';
+import { ui } from '../implementation/logging';
+
+export function yargsCompareCommandObject() {
+  const command = 'compare';
+  return {
+    command,
+    describe: 'Compare 2 report files and create a diff file',
+    builder: yargsCompareOptionsDefinition(),
```
BioPhoton, line 16: Your could add a group with name `Compare Options` to list all command options under the provided string instead of just 'Options'

packages/cli/src/lib/implementation/compare.options.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function yargsCompareOptionsDefinition(): Record<
+  keyof CompareOptions,
+  Options
+> {
+  return {
+    before: {
+      describe: 'Path to source report.json',
+      type: 'string',
+      demandOption: true,
```
BioPhoton, line 12: In the future we can use prompts and move this line into a middleware with nice user prompts

packages/cli/src/lib/yargs-cli.integration.test.ts
```diff
@@ -118,4 +120,24 @@ describe('yargsCli', () => {
       }),
     );
   });
+
+  it('should parse compare options', async () => {
+    const parsedArgv = await yargsCli<GeneralCliOptions & CompareOptions>(
+      ['--before=source-report.json', '--after', 'target-report.json'],
+      {
+        options: { ...options, ...yargsCompareOptionsDefinition() },
```
BioPhoton, line 128: Do we need all other options here?

BioPhoton (dismissed):
Left some comments. 
Nice new command! Whoop Whoop! 🎉

packages/cli/src/lib/yargs-cli.integration.test.ts
matejchalk: All the other integration tests in this file include them, so I wanted to be consistent. And the `compare` command other common options as well, so including them makes it closer to the actual scenario.

packages/cli/README.md
```diff
@@ -240,6 +240,21 @@ Run plugins, collect results and upload the report to the Code PushUp portal.
 
 Refer to the [Common Command Options](#common-command-options) for the list of available options.
 
+### `compare` command
+
+Usage:
+`code-pushup compare [options]`
```
Tlacenka, line 246: I always considered `[arg]` to be optional. Since this command has two required arguments, the command will always need the `--before` and `--after`. The usage should reflect that as accurately as possible, imo.

packages/cli/src/lib/compare/compare-command.ts
matejchalk: Sure. But I would only make that change once the other groups from #553 are merged.

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
```diff
@@ -9,6 +9,7 @@ Commands:
   code-pushup autorun       Shortcut for running collect followed by upload
   code-pushup collect       Run Plugins and collect results
   code-pushup upload        Upload report results to the portal
+  code-pushup compare       Compare 2 report files and create a diff file
```
Tlacenka, line 12: While I see the `compare` command in `help`, I cannot see the required arguments here. I only see it when I run `code-pushup compare --help`. This does not seem intuitive to me. 
If not add them directly, I would create a new group for these commands to display them explicitly for the `compare` command.

e2e/cli-e2e/tests/compare.e2e.test.ts
```diff
@@ -0,0 +1,58 @@
...
+describe('CLI compare', () => {
+  const git = simpleGit();
+
+  beforeEach(async () => {
+    if (await git.diff(['--', 'examples/react-todos-app'])) {
+      throw new Error(
+        'Unstaged changes found in examples/react-todos-app, please stage or commit them to prevent E2E tests interfering',
+      );
+    }
+    await cleanTestFolder('tmp/e2e');
```
Tlacenka, line 15: Perhaps this should be in a subfolder - what if another E2E test has a file there that runs in parallel with this one?

e2e/cli-e2e/tests/compare.e2e.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      command: 'code-pushup',
+      args: ['collect', '--persist.filename=target-report'],
+      cwd: 'examples/react-todos-app',
+    });
+  }, 20_000);
+
+  afterEach(async () => {
+    await git.checkout(['--', 'examples/react-todos-app']);
+    await cleanTestFolder('tmp/e2e');
+  });
```
Tlacenka, line 36: Question. Shouldn't this be `afterAll`? The test (tests if there are any more in the future) cleans the folder in the `before` hook already so `afterEach` is questionable for me.

e2e/cli-e2e/tests/compare.e2e.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+    await cleanTestFolder('tmp/e2e');
+    await executeProcess({
+      command: 'code-pushup',
+      args: ['collect', '--persist.filename=source-report'],
+      cwd: 'examples/react-todos-app',
+    });
+    await executeProcess({
+      command: 'npx',
+      args: ['eslint', '--fix', 'src', '--ext=js,jsx'],
```
Tlacenka, line 23: Nice 😁

e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
```diff
@@ -0,0 +1,705 @@
...
+          "before": "62 %",
+        },
+        "plugin": {
+          "slug": "coverage",
+          "title": "Code coverage",
+        },
+        "scores": {
+          "after": 0.617,
+          "before": 0.6211,
+          "diff": -0.0040999999999999925,
```
Tlacenka, line 20: I feel like the diff should also have 4 or so decimal digits - the highest level of precision based on `after` and `before` precision. 🤔

e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
```diff
@@ -0,0 +1,705 @@
...
+      },
+    ],
+    "removed": [],
+    "unchanged": [
+      {
+        "plugin": {
+          "slug": "eslint",
+          "title": "ESLint",
+        },
+        "score": 0.9166666666666666,
```
Tlacenka, line 678: This actually goes for the `score` as well then. But OOS of course. WDYT?

e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
Tlacenka, line 1: Ideally, the snapshot would be shorter - perhaps running less ESLint rules on the example repo could help. Can be done in a separate issue/PR.

Tlacenka (commented):
Well tested. 🧪 

Have some concerns about the snapshot length and `help` documentation, otherwise good to go!

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
matejchalk: Why not, but I would optimize the option groups in #553 or once it's merged, so that it's done as a whole.

e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
matejchalk: I would lean towards keeping it simple and sticking to the JavaScript precision. The JSON isn't meant for display, it's the raw values. We will definitely round these values whenever we display it (e.g. for the Markdown diff).

e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
matejchalk: Answered here: https://github.com/code-pushup/cli/pull/556#discussion_r1521227569

e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
Tlacenka: As long as different runs of E2E tests do not produce different values in snapshots and report false positives, it's fine by me.

matejchalk: It isn't random, just hard to read.

e2e/cli-e2e/tests/compare.e2e.test.ts
matejchalk: There's only test, so I didn't think about this much. But using each hooks make more sense to be if there were more tests, each test should started from the clean state.

e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
matejchalk: Agreed, would be helpful in general since most of the lint rules pass anyway. I'll create an issue for it after lunch.

packages/cli/README.md
matejchalk: Good point :+1: Updated.

e2e/cli-e2e/tests/__snapshots__/compare.e2e.test.ts.snap
matejchalk: #558

packages/cli/src/lib/compare/compare-command.ts
matejchalk: Also mentioned here, BTW :grin: https://github.com/code-pushup/cli/pull/556#discussion_r1521215217

Tlacenka (approved):
Nice! 🍻

BioPhoton (approved):

PR #563, opened 2024-03-13
fix(plugin-eslint): eslint startline can be zero

markusnissl:
In case tsconfig does not contain correct setup, e.g., "strictNullChecks": true, eslint throws warnings for line "zero". The current schema does not show the warning, it just throws an internal error. This fix allows for warning on line zero to be collected for the report.

Fixes issues such as:
  ```
{
    "code": "too_small",
    "minimum": 0,
    "type": "number",
    "inclusive": false,
    "exact": false,
    "message": "Number must be greater than 0",
    "path": [
      117,
      "details",
      "issues",
      3,
      "source",
      "position",
      "startLine"
    ]
  },

```

for issues like 
`0:1  warning  This rule requires the `strictNullChecks` compiler option to be turned on to function correctly  @typescript-eslint/prefer-nullish-coalescing`

### DIFF
--- a/packages/models/src/lib/issue.ts
+++ b/packages/models/src/lib/issue.ts
@@ -8,7 +8,7 @@ const sourceFileLocationSchema = z.object(
     position: z
       .object(
         {
-          startLine: positiveIntSchema.describe('Start line').optional(),
+          startLine: positiveIntSchema.describe('Start line'),
           startColumn: positiveIntSchema.describe('Start column').optional(),
           endLine: positiveIntSchema.describe('End line').optional(),
           endColumn: positiveIntSchema.describe('End column').optional(),
--- a/packages/models/src/lib/issue.ts
+++ b/packages/models/src/lib/issue.ts
@@ -1,6 +1,10 @@
 import { z } from 'zod';
 import { MAX_ISSUE_MESSAGE_LENGTH } from './implementation/limits';
-import { filePathSchema, positiveIntSchema } from './implementation/schemas';
+import {
+  filePathSchema,
+  nonnegativeIntSchema,
+  positiveIntSchema,
+} from './implementation/schemas';
 
 const sourceFileLocationSchema = z.object(
   {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by markusnissl** [comment]
@matejchalk I changed PR to implement first solution. 

While I like the second solution to throw an error, it is against the idea of "monitoring" the progress for transformation as one has e.g., to fix some projects in monorepos to have the rules in place.

File: packages/models/src/lib/issue.ts [hunk 1]
```diff
@@ -8,7 +8,7 @@ const sourceFileLocationSchema = z.object(
     position: z
       .object(
         {
-          startLine: positiveIntSchema.describe('Start line'),
+          startLine: positiveIntSchema.describe('Start line').optional(),
```
matejchalk, line 11: The `startLine` can stay required, instead you can take advantage of the fact that the whole `position` object is optional. If `line: 0` is encountered, the `position` should not be set, emittiing `position: { startColumn: 1 }` isn't valid.

```suggestion
          startLine: positiveIntSchema.describe('Start line'),
```

packages/plugin-eslint/src/lib/runner/transform.ts
```diff
@@ -66,7 +66,7 @@ function convertIssue(issue: LintIssue): Issue {
     source: {
       file: issue.filePath,
       position: {
-        startLine: issue.line,
+        ...(issue.line > 0 && { startLine: issue.line }),
```
matejchalk, line 69: Please change this so `position` isn't set if we don't have a valid start line. I.e. something like `...issue.line > 0 && { position: { startLine: issue.line, ... } }` :pray:

packages/plugin-eslint/src/lib/runner/transform.unit.test.ts
```diff
@@ -221,6 +238,25 @@ describe('lintResultsToAudits', () => {
...
+        details: {
+          issues: [
+            {
+              message:
+                'This rule requires the strictNullChecks compiler option to be turned on to function correctly',
+              severity: 'warning',
+              source: {
+                file: 'src/app/test/strictNullChecks.ts',
+                position: { startColumn: 1 },
+              },
```
matejchalk, line 255: This is the result I would expect:

```suggestion
              source: {
                file: 'src/app/test/strictNullChecks.ts',
              },
```

matejchalk (changes requested):

matejchalk (approved):
Thanks for the fix :slightly_smiling_face:

Issue #559, opened 2024-03-12
CLI report diff MD formatting (reports-diff.json => reports-diff.md)


matejchalk:
Part of epic #478 

### Acceptance criteria

- [x] design and implement `report-diff.md` formatting
- [x] should include categories with changed score
- [x] should include audits with changed score and/or value
- [x] should use table layout for readability

Issue #149, opened 2023-10-25
Git URLs in Markdown reports and diffs


matejchalk:
The CLI need can improve in UX when we link the report source file issues to the related file. If we click the link alter we should either open up the file locally in our IDE or jump to the github repository, depending on the usage.

This issue is a follow-up to #148 

**Comparison Report:**
- [x] commits in `report-diff.md` summary and `report.md` About section should link to web URL of given commit (using SHA)

**Report:**
- [x] #748
  - [x] #722
  - [x] #745

## Implementation details:

- **PR1** - Models and utils
- **PR2** - IDE linking #722
- **PR3** - Github linking #745

BioPhoton:
found this:
https://github.com/sshaw/git-link

matejchalk:
The file links were already implemented by @hanna-skryl in #824 and #840.

For the commit links, it turns out no logic is needed on our end, because both GitHub and GitLab autolink commit SHAs in markdown comments :slightly_smiling_face:

![Image](https://github.com/user-attachments/assets/c7628f6d-977d-423a-82b4-cdbe781bba64)
![Image](https://github.com/user-attachments/assets/7f830921-a640-4827-9ffa-e5947d39e2a3)

So this can be closed.

Issue #558, opened 2024-03-12
Reduce number of lint rules in example app


matejchalk:
The `react-todos-app` is used in E2E tests for testing ESLint plugin among others. But since a lot of recommended ESLint rules are included, the test snapshots are quite large and therefore less readable. Most of the rules pass anyway, so they don't give much data variety. So we could remove some passing rules from the ESLint config without losing any test coverage.

vmasek:
Wanted to have a look but I am not able to find the config in repo 🤔

matejchalk:
There are a couple of places:
- [examples/react-todos-app/.eslintrc.js](https://github.com/code-pushup/cli/blob/main/examples/react-todos-app/.eslintrc.js) is used in E2E tests,
- [packages/plugin-eslint/mocks/fixtures/todos-app/.eslintrc.js](https://github.com/code-pushup/cli/blob/main/packages/plugin-eslint/mocks/fixtures/todos-app/.eslintrc.js) is used in plugin's integration tests

I believe the `extends: 'plugin:react/recommended'` pulls in most of the uninteresting rules, so my plan was to remove that. But haven't gotten round to it yet, there's always something higher prio that grabs my attention :sweat_smile:

PR #569, opened 2024-03-15
feat(utils): implement markdown diff formatting

matejchalk:
Main part of #559 

#### Scope

- markdown formatting logic in `utils`
  - content: 
    - summary text includes overall outcome (positive/negative/neutral/mixed) prints compared commits 
    - Categories section list all current categories in table (changed, added and unchanged)
    - Groups section lists changed groups in collapsed table
    - Audits section lists changed audits in collapsed table (truncated to first 100)
  - notes:
    - [Shield.io static badge](https://shields.io/badges/static-badge) is used for colorizing score/value changes
      - HTML colors (`<span style="color: ...">...</span>`) aren't rendered in Markdown by GitHub or GitLab :disappointed: 
      - LaTeX colorizing (`$\color{...}{...}$`) is supported by GitHub, but not GitLab
    - for linking commits we should re-prioritize #149 
- tested on a variety of report diffs
  - **see `report-diff-*.md` files in [`packages/utils/src/lib/reports/__snapshots__`](https://github.com/code-pushup/cli/tree/ef381a66554f20fd4cf81a074369068fd845bf8d/packages/utils/src/lib/reports/__snapshots__)**
  - added more realistic mocks for this purpose 
  - replaced `toMatchSnapshot()` with `toMatchFileSnapshot('__snapshots__/<file-with-ext>')` so output is easier to read than in `.snap` files
    - `.md` snapshots are easy to render in IDE or GitHub 
    - `.txt` snapshots for stdout don't alter column layout by escaping backticks

#### Next steps
- integrate with `compare` command
- include in E2E tests
- add to CI


.prettierignore
```diff
@@ -1,4 +1,5 @@
 # Add files here to ignore them from prettier formatting
 /dist
 /coverage
-/.nx/cache
\ No newline at end of file
+/.nx/cache
+__snapshots__
```
BioPhoton, line 5: maybe all snapshots folder?
```suggestion
**/__snapshots__
```

testing/test-utils/src/lib/utils/dynamic-mocks/eslint-plugin.mock.ts
```diff
@@ -24,7 +29,7 @@ const ESLINT_PLUGIN_GROUP_MAX_LINES: Group = {
   ],
 };
 
-const eslintMeta = {
+export const ESLINT_PLUGIN_META = {
```
BioPhoton, line 32: Nice renaming! Valid also for all other similar changes.

BioPhoton (commented):
Left some comments. Like the renaming to UPPERCASE, also the feature it self is very helpful for our development so I can't wait to see it landing step by step.

.prettierignore
matejchalk: Both syntaxes match all nested `__snapshots__` folders. There's no leading `/`, so it doesn't only match root directories.

BioPhoton: Ah is this a general standard for ignore files?

matejchalk: Yes. Prettier uses the [gitignore](https://git-scm.com/docs/gitignore#_pattern_format) syntax.

packages/utils/src/lib/reports/utils.ts
```diff
@@ -16,12 +16,23 @@ import {
...
 } from '../file-system';
 import { SCORE_COLOR_RANGE } from './constants';
+import { image, style } from './md';
 import { ScoredReport, SortableAuditReport, SortableGroup } from './types';
 
 export function formatReportScore(score: number): string {
   return Math.round(score * 100).toString();
 }
 
+export function formatScoreWithColor(
```
Tlacenka, line 26: Where it makes sense, it would be nice to cover helper functions with unit tests (including those helper functions in the generate file itself.

testing/test-utils/src/lib/utils/dynamic-mocks/eslint-audits.mock.ts
```diff
@@ -1,5 +1,17 @@
 import type { AuditReport } from '@code-pushup/models';
 
+export const ESLINT_AUDITS_FIXED_SLUGS: (keyof typeof ESLINT_AUDITS_MAP)[] = [
```
Tlacenka, line 3: Can we reduce the total number of audits used in these integration tests? I am worried about the maintenance. Test data should be minimal while meeting diversity requirements - do e.g. all 9 of these audits display different behaviour?

packages/utils/src/lib/reports/__snapshots__/report-diff-added.md
```diff
@@ -0,0 +1,28 @@
+# Code PushUp
+
+😟 Code PushUp report has **regressed** – compared target commit `0123456` with source commit `abcdef0`.
+
+## 🏷️ Categories
+
+|🏷️ Category|⭐ Current score|⭐ Previous score|🗠 Score change|
+|:--|:--:|:--:|:--:|
+|Bug prevention|🟡 **63**|🟡 68|![🠋 −5](https://img.shields.io/badge/%F0%9F%A0%8B%20%E2%88%925-red)|
+|Performance|🟢 **94**|n/a|n/a|
```
Tlacenka, line 10: Wondering about the n/a... if we could add **(new)** somewhere instead or put `n/a` in code format maybe...can't quite put my finger on it. WDYT?

vmasek (dismissed):

packages/utils/src/lib/reports/__snapshots__/report-diff-improved.md
```diff
@@ -0,0 +1,48 @@
+# Code PushUp
+
+🥳 Code PushUp report has **improved** – compared target commit `0123456` with source commit `abcdef0`.
+
+## 🏷️ Categories
+
+|🏷️ Category|⭐ Current score|⭐ Previous score|🗠 Score change|
```
Tlacenka, line 7: The score change emoji is barely visible on my machine. 😢 
Could we replace it by something else, e.g. one of these? 🚀🔄🌱📈
image

packages/utils/src/lib/reports/__snapshots__/report-diff-unchanged.md
Tlacenka, line 1: Nicely handled unchanged use case.

packages/utils/src/lib/reports/generate-md-reports-diff.ts
```diff
@@ -0,0 +1,297 @@
...
+
+// to prevent exceeding Markdown comment character limit
+const MAX_ROWS = 100;
+
+export function generateMdReportsDiff(diff: ReportsDiff): string {
+  return paragraphs(
+    formatDiffHeaderSection(diff),
+    formatDiffCategoriesSection(diff),
+    formatDiffGroupsSection(diff),
+    formatDiffAuditsSection(diff),
```
Tlacenka, line 22: These could have their own unit tests.

packages/utils/src/lib/reports/generate-md-reports-diff.ts
```diff
@@ -0,0 +1,297 @@
...
+  scores,
+}: Pick<AuditDiff, 'values' | 'scores'>): string {
+  const marker = getDiffMarker(values.diff);
+  const percentage =
+    values.before === 0
+      ? values.diff > 0
+        ? Number.POSITIVE_INFINITY
+        : Number.NEGATIVE_INFINITY
+      : Math.round((100 * values.diff) / values.before);
+  const text = `${formatDiffNumber(percentage)}%`;
```
Tlacenka, line 200: Shouldn't there be a whitespace between a number and percentage?
```suggestion
  const text = `${formatDiffNumber(percentage)} %`;
```

Tlacenka (commented):
Good work!

Have a few thoughts and testing feedback below.
Not everything needs to be done in this PR of course.

testing/test-utils/src/lib/utils/dynamic-mocks/eslint-audits.mock.ts
matejchalk: I would make this part of the issue to reduce number of ESLint audits - #558

packages/utils/src/lib/reports/utils.ts
```diff
@@ -42,6 +53,35 @@ export function getSquaredScoreMarker(score: number): string {
...
+export function shieldsBadge(text: string, color: string): string {
+  return image(
+    `https://img.shields.io/badge/${encodeURIComponent(text)}-${color}`,
+    text,
+  );
+}
+
+export function formatDiffNumber(diff: number): string {
+  const number =
+    Math.abs(diff) === Number.POSITIVE_INFINITY ? '∞' : `${Math.abs(diff)}`;
```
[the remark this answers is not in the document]
matejchalk: It's the correct mathematical change percentage when the old value is 0. I definitely wouldn't omit it from the diff - if a previously passing ESLint rule failed, I'd want to know about it. And I can't think of a more transparent way of showing by how much it changed. Only arrow would be more confusing, I think. Could theoretically difference between raw values, but in general I think the percentage is a better indicator (raw value could be any unit).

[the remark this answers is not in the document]
Tlacenka: What about a different arrow, e.g. ⏫ ?

packages/utils/src/lib/reports/generate-md-reports-diff.ts
matejchalk: If you think it looks better, I can change it.

image

packages/utils/src/lib/reports/__snapshots__/report-diff-added.md
matejchalk: Are any of these better?

image

Or with an asterisk:

image

packages/utils/src/lib/reports/utils.ts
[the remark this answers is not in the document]
matejchalk: You mean like this? :raised_eyebrow: 

image

packages/utils/src/lib/reports/__snapshots__/report-diff-added.md
vmasek: `n/a` in code wrapper and * without ()

Tlacenka: After looking at this, I am a fan of the italics, be it with or without an asterisk. When it comes to the asterisk, I prefer it with the parentheses as it is done e.g. in our generated zod schema documentation.

matejchalk: Went with italics and asterisk in parentheses.

packages/utils/src/lib/reports/generate-md-reports-diff.ts
matejchalk: Added thin space character.

packages/utils/src/lib/reports/__snapshots__/report-diff-improved.md
matejchalk: Replaced with 🔄.

packages/utils/src/lib/reports/types.ts
```diff
@@ -28,3 +28,5 @@ export type SortableAuditReport = AuditReport & {
   weight: number;
   plugin: string;
 };
+
+export type DiffOutcome = 'positive' | 'negative' | 'neutral' | 'mixed';
```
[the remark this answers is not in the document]
matejchalk: Good point, renamed.

vmasek (dismissed):

packages/utils/src/lib/reports/utils.ts
matejchalk: Added some more unit tests.

vmasek (dismissed):

packages/utils/src/lib/reports/generate-md-reports-diff.ts
matejchalk: These functions are already covered extensively by the integration tests, I don't see how unit testing them again separately would improve the test coverage. A lot of data is needed in the report diff, so crafting the test cases is very time-consuming. And it's more beneficial to test it holistically because the Markdown output is inherently visual.

Is there some test case which you would like to see covered, which isn't already covered by the integration tests?

Tlacenka: Well, the general idea is to have a test written for as small a scope as possible. The test data could then be much simpler as well (only categories or only audits). So the question is more the other way around - what can we already cover with unit tests that we wouldn't need to create integration test cases for (only keep a happy path).

But I suppose this would require more effort so it is okay for now.

Tlacenka (commented):
Thank you for adding the unit tests 🙏 

Just two more minor things.

packages/utils/src/lib/reports/__snapshots__/report-diff-added.md
```diff
@@ -0,0 +1,30 @@
+# Code PushUp
+
+😟 Code PushUp report has **regressed** – compared target commit `0123456` with source commit `abcdef0`.
+
+## 🏷️ Categories
+
+|🏷️ Category|⭐ Current score|⭐ Previous score|🔄 Score change|
+|:--|:--:|:--:|:--:|
+|Bug prevention|🟡 **63**|🟡 68|![🠋 −5](https://img.shields.io/badge/%F0%9F%A0%8B%20%E2%88%925-red)|
+|Performance|🟢 **94**|_n/a (*)_|_n/a (*)_|
```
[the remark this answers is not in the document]
matejchalk: I would say yes. 

I just noticed that I don't escape the asterisk in Markdown properly though :sweat_smile: Will fix.

packages/utils/src/lib/reports/utils.unit.test.ts
```diff
@@ -37,6 +40,48 @@ describe('formatReportScore', () => {
...
+    expect(colorByScoreDiff('🠋 −8 %', 0.03)).toMatch(/shields\.io.*-green/);
+  });
+
+  it('should use red badge if score dropped', () => {
+    expect(colorByScoreDiff('🠉 +33 %', -1)).toMatch(/shields\.io.*-red/);
+  });
+
+  it("should use gray badge if score didn't change", () => {
+    expect(colorByScoreDiff('🠉 +50 %', 0)).toMatch(/shields\.io.*-gray/);
+  });
```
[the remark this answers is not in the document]
Tlacenka: Oh, I thought it was the category table. My bad then 😅

packages/utils/src/lib/reports/__snapshots__/report-diff-added.md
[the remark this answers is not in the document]
matejchalk: Fixed.

Tlacenka (approved):

PR #573, opened 2024-03-19
ci: integrate code-pushup PR comments

### LINKED ISSUE #559: CLI report diff MD formatting (reports-diff.json => reports-diff.md)

matejchalk:
Closes #559


vmasek (approved):

vmasek:
It is even better than I expected 🥳

packages/utils/src/lib/reports/__snapshots__/report-diff-mixed.md
Tlacenka, line 1: Now that I look at the mixed results, I wonder if we shouldn't sort them in the following way: regressions > improvements > no score change

Currently, regressions and improvements are mixed and without the score column, one may overlook a regression. By having them at the top, all actionable changes are nicely visible.

Tlacenka (approved):
Looking great!

Love that the comment is updated 😁

packages/utils/src/lib/reports/__snapshots__/report-diff-mixed.md
matejchalk: Not sure if that would be better :thinking:  If a major improvement is made, but there's also a minor regression, does it makes sense to list the regression first?

It's ordered by absolute score change (positive or negative), then by value change. The idea is to sort it from highest impact.

In general, regressions can be found by scanning for red badges.

Tlacenka: Let the public decide once it is used by our clients. It is not in scope of this PR anyway.

PR #574, opened 2024-03-19
feat(core): create report-diff.md file if specified by persist.format

matejchalk:
Part of #559 

I need to get these changes into `main` before I can implement the GitHub Actions workflow.

Scope:
- `compare` command now creates markdown diff
- changed default `persist.format` from `['json']` to `['json', 'md']`, so that basic `compare` flow works without custom configuration
- added markdown diff to E2E tests
- replaced multi-target `nx run-collect`, `nx run-upload` etc. with single target `nx code-pushup -- <command-and-options-here>`

### DIFF
--- a/packages/core/src/lib/compare.ts
+++ b/packages/core/src/lib/compare.ts
@@ -1,8 +1,16 @@
 import { writeFile } from 'node:fs/promises';
-import { Report, ReportsDiff, reportSchema } from '@code-pushup/models';
+import { join } from 'node:path';
+import {
+  type Format,
+  type PersistConfig,
+  Report,
+  ReportsDiff,
+  reportSchema,
+} from '@code-pushup/models';
 import {
   Diff,
   calcDuration,
+  generateMdReportsDiff,
   readJsonFile,
   scoreReport,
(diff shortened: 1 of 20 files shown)


packages/core/src/lib/compare.ts
```diff
@@ -29,7 +39,14 @@ export async function compareReportFiles(
 
   const reportsDiff = compareReports(reports);
 
-  await writeFile(outputPath, JSON.stringify(reportsDiff, null, 2));
+  return Promise.all(
+    format.map(async fmt => {
+      const outputPath = join(outputDir, `${filename}-diff.${fmt}`);
+      const content = reportsDiffToFileContent(reportsDiff, fmt);
+      await writeFile(outputPath, content);
```
Tlacenka, line 46: Shouldn't we make sure the `outputPath` exists like in `persist` in `core`?

Tlacenka (approved):

packages/core/src/lib/compare.ts
matejchalk: Good point. Updated.

PR #584, opened 2024-03-25
fix(cli): fix missing executable permissions for index.js

matejchalk:


Tlacenka (approved):

matejchalk:
The `compare` command fails because it's being run in `main` - not very practical, but I'll replace it all with the GitHub Action soon.

Issue #542, opened 2024-03-06
NPM dependencies plugin(s)


matejchalk:
### User story

As a Code PushUp user, I would like to see **security vulnerabilities** from my NPM packages listed and used to calculate a _Security_ category score. I would also like to know about **out-of-date packages**, which can be used in an _Up-to-date_ or _Future proof_ category.

It would be great to have support for all the commonly used package managers:

- [x] NPM
- [x] Yarn classic (v1) 
- [x] Yarn modern (v2+)
- [x] PNPM

### Acceptance criteria
- [x] 1 plugin for all (`@code-pushup/js-packages-plugin`)
- [x] let user configure which of the package manager should be running
- [x] How do we integrate audit levels (`info`, `low`, `moderate`, `high`, `critical`) with issue severities?
  - extend issue severity enum in our data models (current: `info`, `warning`, `error`)?
  - [x] convert them to existing levels (e.g. `info` ➡ `info`, `low`/`moderate` ➡ `warning`, `high`/`critical` ➡ `error`)?
    - [x] keep this option as default but let it configurable

### Implementation details

The commands to list vulnerabilities and outdated packages are similar for NPM, Yarn 1 and PNPM, since they're all based on the NPM CLI. However, Yarn 2+ has a an altogether different set of commands, so alternatives must be found there (more details in [migration guide](https://yarnpkg.com/migration/guide)).

| Package manager | Vulnerabilities command | Outdated command |
| :-- | :-- | :-- |
| [NPM](https://www.npmjs.com/) | [`npm audit --json`](https://docs.npmjs.com/cli/v10/commands/npm-audit) | [`npm outdated --json`](https://docs.npmjs.com/cli/v10/commands/npm-outdated) |
| [Yarn 1](https://classic.yarnpkg.com/en/) | [`yarn audit --json`](https://classic.yarnpkg.com/en/docs/cli/audit) | [`yarn outdated --json`](https://classic.yarnpkg.com/lang/en/docs/cli/outdated/) |
| [Yarn 2+](https://yarnpkg.com/) | [`yarn npm audit --json`](https://yarnpkg.com/cli/npm/audit) | no built-in command (see https://github.com/yarnpkg/berry/issues/749), but there is a [Yarn plugin](https://github.com/mskelton/yarn-plugin-outdated)  |
[...]

PR #590, opened 2024-03-27
feat(plugin-js-packages): add Yarn v2 support

Tlacenka:
Related to #542 

In this PR, I add the support for Yarn v2. Notable changes:
- I updated the documentation regarding the Yarn v2 specific things.
- Yarn v2 does not support security audit for optional dependencies. This audit is skipped.
- I improved the plugin title.

Known issues:
- I was unable to find an indirect vulnerability so the logic is not verified.

The results were tested on jest repository.

### DIFF
--- a/packages/plugin-js-packages/README.md
+++ b/packages/plugin-js-packages/README.md
@@ -10,11 +10,13 @@ This plugin checks for known vulnerabilities and outdated dependencies.
 It supports the following package managers:
 
 - [NPM](https://docs.npmjs.com/)
-- [Yarn v1](https://classic.yarnpkg.com/docs/)[Yarn v2+](https://yarnpkg.com/getting-started)
+- [Yarn v1](https://classic.yarnpkg.com/docs/)
+- [Yarn v2+](https://yarnpkg.com/getting-started)
+  - In order to check outdated dependencies for Yarn v2+, you need to install [`yarn-plugin-outdated`](https://github.com/mskelton/yarn-plugin-outdated).
 - [PNPM](https://pnpm.io/pnpm-cli)
 
 > ![NOTE]
-> As of now, in order to check outdated dependencies for Yarn v2+, you need to install [`yarn-plugin-outdated`](https://github.com/mskelton/yarn-plugin-outdated).
+> As of now, Yarn v2 does not support security audit of optional dependencies. Only production and dev dependencies audits will be included in the report.
 
 ## Getting started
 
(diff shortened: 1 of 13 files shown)


packages/plugin-js-packages/src/lib/js-packages-plugin.unit.test.ts
```diff
@@ -16,7 +16,8 @@ describe('jsPackagesPlugin', () => {
     ).resolves.toStrictEqual(
       expect.objectContaining({
         slug: 'js-packages',
-        title: 'Plugin for JS packages',
+        title:
```
BioPhoton, line 19: Could we get a shorter title and more text in  the description? This would otherwise look odd in a tooltip I think

Tlacenka: Sure! Any ideas?

BioPhoton: JS Package Versions?

matejchalk: Yes, title should be shorter and `Plugin for ` is superfluous. I would suggest that in most cases, the `title` is simply the non-slugified form of the `slug` and package name (without `plugin`). So `JS Packages` would be the obvious candidate. I wouldn't go with `JS Package Versions`, because outdated versions is only one of the features, security audit is another big part of it.

packages/plugin-js-packages/README.md
```diff
@@ -10,9 +10,13 @@ This plugin checks for known vulnerabilities and outdated dependencies.
 It supports the following package managers:
 
 - [NPM](https://docs.npmjs.com/)
-- [Yarn v1](https://classic.yarnpkg.com/docs/) & [Yarn v2+](https://yarnpkg.com/getting-started)
+- [Yarn v1](https://classic.yarnpkg.com/docs/)
+- [Yarn v2+](https://yarnpkg.com/getting-started): In order to check outdated dependencies for Yarn v2+, you need to install [`yarn-plugin-outdated`](https://github.com/mskelton/yarn-plugin-outdated)
```
matejchalk, line 14:
```suggestion
- [Yarn v2+](https://yarnpkg.com/getting-started)
  - in order to check outdated dependencies for Yarn v2+, you need to install [`yarn-plugin-outdated`](https://github.com/mskelton/yarn-plugin-outdated)
```

matejchalk (dismissed):
That was quick :rocket:

packages/plugin-js-packages/src/lib/js-packages-plugin.unit.test.ts
Tlacenka: JS Packages it is, thank you for the ideas.

matejchalk (approved):

PR #595, opened 2024-04-01
ci: tweak release configuration

### LINKED ISSUE #594: ci: fix invalid release configuration

getlarge:
Fixes #594 

---

<details>
<summary><code>migrations.json</code></summary>

```json
{
  "migrations": [
    {
      "cli": "nx",
      "version": "18.0.0-beta.2",
      "description": "Updates nx.json to disabled adding plugins when generating projects in an existing Nx workspace",
      "implementation": "./src/migrations/update-18-0-0/disable-crystal-for-existing-workspaces",
      "x-repair-skip": true,
      "package": "nx",
      "name": "18.0.0-disable-adding-plugins-for-existing-workspaces"
    },
    {
      "version": "18.1.0-beta.3",
      "description": "Moves affected.defaultBase to defaultBase in `nx.json`",
      "implementation": "./src/migrations/update-17-2-0/move-default-base",
      "package": "nx",
      "name": "move-default-base-to-nx-json-root"
    },
    {
      "cli": "nx",
      "version": "18.0.0-beta.0",
      "description": "Add NX_MF_DEV_SERVER_STATIC_REMOTES to inputs for task hashing when '@nx/webpack:webpack' is used for Module Federation.",
      "factory": "./src/migrations/update-18-0-0/add-mf-env-var-to-target-defaults",
      "package": "@nx/react",
      "name": "add-module-federation-env-var-to-target-defaults"
    },
    {
      "cli": "nx",
      "version": "18.1.1-beta.0",
      "description": "Ensure targetDefaults inputs for task hashing when '@nx/webpack:webpack' is used are correct for Module Federation.",
      "factory": "./src/migrations/update-18-1-1/fix-target-defaults-inputs",
      "package": "@nx/react",
      "name": "fix-target-defaults-for-webpack"
    }
  ]
}
```

</details>

### DIFF
--- a/packages/plugin-coverage/src/lib/nx/coverage-paths.ts
+++ b/packages/plugin-coverage/src/lib/nx/coverage-paths.ts
@@ -1,7 +1,8 @@
+/// <reference types="vitest" />
 import type { ProjectGraphProjectNode, TargetConfiguration } from '@nx/devkit';
 import chalk from 'chalk';
 import { join } from 'node:path';
-import { ui } from '@code-pushup/utils';
+import { importEsmModule, ui } from '@code-pushup/utils';
 import { CoverageResult } from '../config';
 
 /**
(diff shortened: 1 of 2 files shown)


global-setup.e2e.ts
```diff
@@ -7,9 +7,9 @@ import stopLocalRegistry from './tools/scripts/stop-local-registry';
 export async function setup() {
   await globalSetup();
   await startLocalRegistry();
-  execSync('npm install -D @code-pushup/cli@e2e');
-  execSync('npm install -D @code-pushup/eslint-plugin@e2e');
-  execSync('npm install -D @code-pushup/coverage-plugin@e2e');
+  execSync('npm install -D @code-pushup/cli@e2e --force');
+  execSync('npm install -D @code-pushup/eslint-plugin@e2e --force');
+  execSync('npm install -D @code-pushup/coverage-plugin@e2e --force');
```
getlarge, line 12: `force` option was necessary due to the move to the major version update of Nx

nx.json
```diff
@@ -39,8 +39,13 @@
...
     },
     "nx-release-publish": {
-      "dependsOn": ["build"],
-      "executor": "@nx/js:release-publish",
+      "cache": false,
+      "dependsOn": [
+        {
+          "target": "build",
+          "params": "forward"
+        }
```
getlarge, line 47: is needed to pass `--skip-nx-cache` flag down to the `nx-release-publish` task pipeline.

nx.json
```diff
@@ -115,5 +117,6 @@
       }
     },
     "releaseTagPattern": "v{version}"
-  }
+  },
+  "useInferencePlugins": false
```
getlarge, line 121: automatically added by Nx during migrations. probably to avoid surprises of Project Crystal.

global-setup.e2e.ts
matejchalk: Is it because of `@nx/devkit` peer dependency? We should probably expand `^17.0.0` to `^17.0.0 || ^18.0.0`, so that we're compatible with Nx 18.

nx.json
matejchalk: Probably for the best :grin:

migrations.json
matejchalk, line 1: The `migrations.json` shouldn't be committed. I added it to the PR description in case we need it for future reference.

nx.json
```diff
@@ -95,16 +100,13 @@
...
         "createRelease": "github"
       }
     },
     "git": {
       "commit": true,
-      "push": true,
-      "pushRemote": "origin",
       "tag": true,
       "commitMessage": "release: {version} [skip ci]"
     },
```
matejchalk, line 112: Out of interest, why did you remove these? :thinking:

package-lock.json
matejchalk, line 1: ["`npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing."](https://github.com/code-pushup/cli/actions/runs/8552772054/job/23434588593#step:3:7)

getlarge: this smells like messed up node_modules/package-lock.js combination.
I am quite sure that removing the node_modules and package-lock.json and reinstall will solve this.

PR #596, opened 2024-04-02
feat(plugin-js-packages): add PNPM support

Tlacenka:
Related to #542 
In this PR, I add the support for PNPM audit and outdated checks. Notable changes:
- I add a post-processing step for audits as multiple package managers require an extra step to filter out duplicates.

Tested manually on [SvelteKit](https://github.com/sveltejs/kit).

### DIFF
--- a/packages/plugin-js-packages/src/lib/runner/audit/constants.ts
+++ b/packages/plugin-js-packages/src/lib/runner/audit/constants.ts
@@ -4,6 +4,7 @@ import {
   PackageManager,
 } from '../../config';
 import { dependencyGroupToLong } from '../../constants';
+import { filterAuditResult } from '../utils';
 import { AuditResult } from './types';
 import {
   npmToAuditResult,
@@ -30,6 +31,38 @@ export const normalizeAuditMapper: Record<
   'yarn-classic': yarnv1ToAuditResult,
   'yarn-modern': yarnv2ToAuditResult,
   pnpm: pnpmToAuditResult,
+};
+
+const filterNpmAuditResults = (
+  results: Record<DependencyGroup, AuditResult>,
+) => ({
+  prod: results.prod,
+  dev: filterAuditResult(results.dev, 'name', results.prod),
+  optional: filterAuditResult(results.optional, 'name', results.prod),
+});
+
+const filterPnpmAuditResults = (
+  results: Record<DependencyGroup, AuditResult>,
+) => ({
+  prod: results.prod,
+  dev: results.dev,
+  optional: filterAuditResult(
+    filterAuditResult(results.optional, 'id', results.prod),
+    'id',
+    results.dev,
+  ),
+});
+
+export const postProcessingAuditMapper: Partial<
+  Record<
+    PackageManager,
+    (
+      result: Record<DependencyGroup, AuditResult>,
+    ) => Record<DependencyGroup, AuditResult>
+  >
+> = {
+  npm: filterNpmAuditResults, // prod dependencies need to be filtered out manually since v10
+  pnpm: filterPnpmAuditResults, // optional dependencies don't have an exclusive option so they need duplicates filtered out
 };
 
 const npmDependencyOptions: Record<DependencyGroup, string[]> = {
--- a/packages/plugin-js-packages/src/lib/runner/audit/unify-type.ts
+++ b/packages/plugin-js-packages/src/lib/runner/audit/unify-type.ts
@@ -206,6 +206,7 @@ export function pnpmToAuditResult(output: string): AuditResult {
   const vulnerabilities = Object.values(pnpmResult.advisories).map(
     ({
       module_name: name,
+      id,
       title,
       url,
       severity,
@@ -217,6 +218,7 @@ export function pnpmToAuditResult(output: string): AuditResult {
 
       return {
         name,
+        id,
         title,
         url,
         severity,
--- a/packages/plugin-js-packages/src/lib/runner/audit/types.ts
+++ b/packages/plugin-js-packages/src/lib/runner/audit/types.ts
@@ -104,6 +104,7 @@ export type Yarnv2AuditResultJson = {
 /* eslint-disable @typescript-eslint/naming-convention */
 export type PnpmAuditAdvisory = {
   module_name: string;
+  id: number;
   severity: PackageAuditLevel;
   vulnerable_versions: string;
   recommendation: string;
--- a/packages/plugin-js-packages/src/lib/runner/audit/unify-type.unit.test.ts
+++ b/packages/plugin-js-packages/src/lib/runner/audit/unify-type.unit.test.ts
@@ -358,6 +358,7 @@ describe('pnpmToAuditResult', () => {
           advisories: {
             '123': {
               module_name: '@cypress/request',
+              id: 123,
               severity: 'high',
               vulnerable_versions: '<2.88.12',
               recommendation: 'Upgrade to version 2.88.12 or later',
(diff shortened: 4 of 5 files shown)


packages/plugin-js-packages/src/lib/runner/audit/constants.ts
```diff
@@ -28,9 +30,41 @@ export const normalizeAuditMapper: Record<
...
+    result: Record<DependencyGroup, AuditResult>,
+  ) => Record<DependencyGroup, AuditResult>
+> = {
+  npm: filterNpmAuditResults, // prod dependencies need to be filtered out manually since v10
+  /* eslint-disable @typescript-eslint/naming-convention */
+  'yarn-classic': results => results,
+  'yarn-modern': results => results,
+  /* eslint-enable @typescript-eslint/naming-convention */
+  pnpm: filterPnpmAuditResults, // optional dependencies don't have an exclusive option so they need duplicates filtered out
 };
```
BioPhoton, line 68: I'm not fully aware of the full implementation so please ignore the question if not fitting.

Not sure if it is a good approach to group different implementations of the same thing within one thing. e.g. npm and pnpm in the same object.

The buzzword that pops up here for me is polymorphism. You basically have a shared API e.g. `filterResults` and a list that maintains the different implementations of that interface. then you call always the same API and get rid of branching and other code clutter.

BioPhoton (dismissed):
Made I architectural suggestion. Feel free to consider it.

packages/plugin-js-packages/src/lib/runner/audit/constants.ts
```diff
@@ -28,9 +30,41 @@ export const normalizeAuditMapper: Record<
...
+  PackageManager,
+  (
+    result: Record<DependencyGroup, AuditResult>,
+  ) => Record<DependencyGroup, AuditResult>
+> = {
+  npm: filterNpmAuditResults, // prod dependencies need to be filtered out manually since v10
+  /* eslint-disable @typescript-eslint/naming-convention */
+  'yarn-classic': results => results,
+  'yarn-modern': results => results,
+  /* eslint-enable @typescript-eslint/naming-convention */
```
matejchalk, line 66: These could be omitted if they don't need any post-processing.

packages/plugin-js-packages/src/lib/runner/audit/unify-type.ts
```diff
@@ -189,11 +191,66 @@ export function yarnv2ToAuditResult(output: string): AuditResult {
...
+    summary: {
+      ...pnpmResult.metadata.vulnerabilities,
+      total: getVulnerabilitiesTotal(pnpmResult.metadata.vulnerabilities),
+    },
+  };
+}
+
+export function pnpmToDirectDependency(path: string): string | true {
+  // the format is ". > <direct dependency>@<version> > ... > <current dependency>@<version>"
+  const deps = path.split(' ').filter(item => item !== '>' && item !== '.');
```
matejchalk, line 243: Since ` > ` is the separator, this would be more intuitive:

```suggestion
  const deps = path.split(' > ').filter(item => item !== '.');
```

If `.` is always at the start, could also be:

```suggestion
  const deps = path.split(' > ').slice(1);
```

packages/plugin-js-packages/src/lib/runner/audit/unify-type.ts
```diff
@@ -189,11 +191,66 @@ export function yarnv2ToAuditResult(output: string): AuditResult {
...
+
+  if (deps.length <= 1) {
```
matejchalk, line 245:
```suggestion
  if (deps.length === 0) {
```

Pretty sure the `if`-statement doesn't change the result, though.

matejchalk (dismissed):

packages/plugin-js-packages/src/lib/runner/audit/unify-type.ts
Tlacenka: Good point!

packages/plugin-js-packages/src/lib/runner/audit/unify-type.ts
Tlacenka: It should.
If the original is `. > cypress@13.7.0` then it is a direct dependency. Then the `deps.length === 1` and it should return `true`.
If the original is `. > cypress@13.7.0 > @cypress/request@3.0.0` then the `deps.length === 2` and it should return `cypress` as the name of the first dependency after `.`.

packages/plugin-js-packages/src/lib/runner/audit/constants.ts
Tlacenka: Originally, I had a ternary operator which provided only for those that have it. If I omit it here then the `Record` will break as not all package managers are listed. Should I return to the direct ternary operator? 🤔

packages/plugin-js-packages/src/lib/runner/audit/constants.ts
```diff
@@ -47,16 +81,17 @@ const yarnv2EnvironmentOptions: Record<DependencyGroup, string> = {
   optional: '',
 };
 
+const pnpmDependencyOptions: Record<DependencyGroup, string[]> = {
+  prod: ['--prod', '--no-optional'],
```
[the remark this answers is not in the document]
Tlacenka: While the object can be useful for optional arguments or arguments with values that need pre-processing, here I am listing hardcoded arguments from the package manager documentation. Their values will not change so I do not see the value of adding an extra wrapper function.

packages/plugin-js-packages/src/lib/runner/audit/constants.ts
Tlacenka: Good insight, thank you!
This will, however, require a bigger refactor so I will address this in a follow-up PR.

packages/plugin-js-packages/src/lib/runner/audit/constants.ts
Tlacenka: Partial Record added.

packages/plugin-js-packages/src/lib/runner/audit/constants.ts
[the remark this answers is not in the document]
matejchalk: I personally prefer to see `--prod --no-optional`, because it's very transparent in how it corresponds to `pnpm` command line arguments, which I can look up in their official docs. Having `objToCliArgs({ prod: true, optional: false })` makes this relationship less clear, and I don't really see what advantage it brings :man_shrugging:

packages/plugin-js-packages/src/lib/runner/audit/unify-type.ts
matejchalk: Ah, ignore my comment, I misread the code :grin: My comment would've made sense if the condition was `deps.length < 1`, but I now realize it's `deps.length <= 1` :facepalm:

matejchalk (approved):

packages/plugin-js-packages/src/lib/runner/audit/constants.ts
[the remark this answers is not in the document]
Tlacenka: As the other comments are resolved, I am going to merge the PR. That does not mean I am not open to a discussion about this but it is something that does not affect the functionality and can be changed later. Feel free to bring it up on Slack.

packages/plugin-js-packages/src/lib/runner/audit/constants.ts
matejchalk: I agree with both the proposed refactor and doing it in a different PR :+1:

Tlacenka: Included as part of #597

Issue #597, opened 2024-04-03
Packages plugin issues


Tlacenka:
- [x] Should the version not be the normalized `<major>.<minor>.<patch>` but e.g. `exotic`, the plugin throws. The occurrence should be filtered out instead. Use [semver](https://www.npmjs.com/package/semver) package to deal with these.
- [x] The Yarn v1 outdated data has a dynamic structure. The plugin currently considers it static.
- [x] The adapters for package manager support are now implemented in a declarative way. Instead, providing an object per package manager with its data in one place is more maintainable.
- [x] When `effects` is empty and `isDirect` is `false` for NPM, empty direct dependency is passed.
- [x] Outdated score is now a ratio between total outdated and total major dependencies. It should be a ratio between total dependencies (both up-to-date and outdated) and major outdated. E.g. via `npm ls --all --parseable` for `npm`.

PR #599, opened 2024-04-03
refactor(plugin-js-packages): improve package manager support

Tlacenka:
Related to  #597

In this PR, I refactor the packages plugin as suggested by @BioPhoton:
- The main flow is now only referencing the `PackageManager` object properties.
- Naming is unified.
- Documentation is updated.

The changes were tested manually.


packages/plugin-js-packages/src/lib/package-managers.ts
matejchalk, line 1: I would suggest splitting up this file so that each package manager implementation is separated:

- `package-managers/npm.ts` exports `npmPackageManager`
- `package-managers/yarn-classic.ts` exports `yarnClassicPackageManager`
- ...
- `package-managers/types.ts` exports `PackageManager` type (imported by `npm.ts` etc.)
- `package-managers/index.ts` exports `packageManagers` record and re-exports `PackageManager` type

The `package-managers/index.ts` module would be the entry point that all other parts of the code import from (the imports will actually be unchanged - `import { ... } from './package-managers'` will resolve to `./package-managers/index.ts` instead of `./package-managers.ts`).

matejchalk (commented):
This structure is much cleaner, well done :+1: Suggested some further improvements.

packages/plugin-js-packages/src/lib/package-managers.ts
Tlacenka: Good point. I took it one step further and created separate folders, so that all package manager code is isolated.

packages/plugin-js-packages/src/lib/package-managers/pnpm/audit-resullt.ts
matejchalk, line 1: Typo in file name - `audit-resullt.ts` (double "l") instead of `audit-result.ts`.

packages/plugin-js-packages/src/lib/package-managers.ts
matejchalk: Perfect :ok_hand:

packages/plugin-js-packages/src/lib/package-managers/pnpm/audit-resullt.ts
Tlacenka: Good catch, thank you!

matejchalk (approved):

PR #593, opened 2024-03-30
feat(plugin-lighthouse): add lighthouse plugin

BioPhoton:
This PR includes all remaining snippets and docs to release the Lighthouse plugin.
In the process of comparing the code with the existing plugins I renamed and moved a couple of files. To get a better overview I grouped the changes as seen below.

**Polish from #549:**
- https://github.com/code-pushup/cli/pull/549#discussion_r1547479530

**Moving existing code:**  
- introduced runner folder to encapsulate lighthouse only related logic and types

**Removed code:**  
  - removed unused `getLighthouseCliArguments` function (better types and direct derivation from lighthouse made it obsolete)

**New code in tests:**  
- e2e tests in cli-e2e
- covered edge case in `getConfig` with error
- added unit tests to `getConfig`
- changed a couple of test names to better readability and narrowed their case (mostly runner utils)

**New code in logic:**  
- added helper functions to create `CategoryRefs`
- docs
- installed dev dependency `@types/debug` for better typing
- I implemented normalisation and refactored the incoming and forwarded `flag` types 
- adjust runtime error message for lighthouse plugin as it was not clear which plugin it came from
Screenshot 2024-03-31 at 15 12 12

> [!NOTE]
> The e2e tests of compare are hard to change as they require a clean history to run. This could get discussed on how to fix it.

**Followup changes:**  
- #598 

closes #90

PR #614, opened 2024-04-10
test(plugin-lighthouse): refine tests

BioPhoton:
Addressing: https://github.com/code-pushup/cli/pull/593#discussion_r1559788572

### DIFF
--- a/packages/plugin-lighthouse/src/lib/utils.unit.test.ts
+++ b/packages/plugin-lighthouse/src/lib/utils.unit.test.ts
@@ -176,14 +176,13 @@ describe('filterAuditsAndGroupsByOnlyOptions to be used in plugin config', () =>
       );
 
     expect(filteredAudits).toHaveLength(1);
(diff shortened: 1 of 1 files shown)


packages/plugin-lighthouse/src/lib/utils.unit.test.ts
```diff
@@ -278,12 +278,14 @@ describe('filterAuditsAndGroupsByOnlyOptions to be used in plugin config', () =>
       );
 
     expect(filteredAudits).toHaveLength(1);
```
Tlacenka, line 280: 🧪 Redundancy: The `toHaveLength` assertions are no longer necessary when `toStrictEqual` is used on an array. Applicable in all affected test cases.

BioPhoton: Thx! Removed them on multiple places

Tlacenka (approved):
Looking great!

PR #625, opened 2024-04-13
chore: upgrade nx 18

### LINKED ISSUE #594: ci: fix invalid release configuration

getlarge:
Continue #595
Fixes #594


.github/workflows/release.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - name: Install dependencies
         run: npm ci
       - name: Version, release and publish packages
-        run: npx nx release --yes --dry-run
+        run: npx nx release --yes --dry-run -- --skip-nx-cache
```
BioPhoton, line 46: Why do we need to skip the cache here?

packages/nx-plugin/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.29.0",
   "license": "MIT",
   "dependencies": {
-    "@nx/devkit": "^17.1.3",
+    "@nx/devkit": "^17.0.0 || ^18.0.0",
```
BioPhoton, line 6: would a ">" operator also work to include all 2 versions?

BioPhoton (dismissed):
Thx for this PR!! 
Left some questions.

.github/workflows/release.yml
getlarge: Apparently, the package.json file changes (when bumping the version) are not detected and the `build` target (that `nx-release-publish` depends on) is using the cache output. Meaning 
1. the package.json in the `dist` folder is using the older version
2. there is nothing to publish 

So I configured the `nx-release-publish` target to forward arguments to `build`.
An alternative approach: change or create a new `namedInput` that explicitly includes the projects package.json. How about that?

nx.json
```diff
@@ -39,8 +39,13 @@
...
     "nx-release-publish": {
-      "dependsOn": ["build"],
-      "executor": "@nx/js:release-publish",
+      "cache": false,
+      "dependsOn": [
+        {
+          "target": "build",
+          "params": "forward"
+        }
+      ],
```
getlarge, line 48: @BioPhoton this is where and how arguments are forwarded to build target

nx.json
```diff
@@ -115,5 +117,6 @@
       }
     },
     "releaseTagPattern": "v{version}"
-  }
+  },
+  "useInferencePlugins": false
```
getlarge, line 121: also, just so you are aware i left the default configuration from Nx NOT to use project Crystal. You might want to opt in instead...

packages/nx-plugin/package.json
getlarge: yes, it would accept all versions above X. It might be too permissive if a breaking change occurs in the later (Nx) major version.

nx.json
BioPhoton: Let's create a followup issue for it so we can assign it to someone later please.

packages/nx-plugin/package.json
BioPhoton: ok cool! thx for the answer! the current implementation is good!

.github/workflows/release.yml
BioPhoton: The alternative approach seems more explicit and a tiny bit more self documenting.
Do you think it pays off t write something about build and release i the `CONTRIBUTING.md`?

Let me know what you think.

getlarge: Yes, i agree it's better to configure inputs correctly. I looked at the [nx.json](https://github.com/code-pushup/cli/blob/main/nx.json) and it seems like the `build` target and the related executor should catch changes from package.json... I will try this further...

getlarge: The release config and process can be a bit confusing, so yes, it's worth mentioning some bits of how it works and how it is configured. I will also add some external references to avoid repeating common knowledge...
I'll take care of it in this PR

BioPhoton: Thanks a LOT for the extra effort!

matejchalk (dismissed):
LGTM :+1:

.github/workflows/release.yml
getlarge: After testing it seems like the build cache is behaving correctly, I adapted the nx release config and arguments accordingly.

CONTRIBUTING.md
```diff
@@ -119,3 +119,16 @@ The following optional folders can be present in a project root;
...
+Versioning follows [semantic versioning](https://semver.org/) guidelines and use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) to determine the next version.
+
+Releases are managed via [Nx release CLI](https://nx.dev/features/manage-releases), which is configured in [nx.json](./nx.json) to
+
+- sync versions across projects based on Git tags (`nx release version`)
+- generate changelog based on commit messages (`nx release changelog`)
+- generate GitHub release notes (`nx release changelog`)
+- build and publish packages to NPM (`nx release publish`)
+
+Changes are automatically released to GitHub and released to NPM via [CI/CD pipeline](./.github/workflows/release.yml) when merged to the main branch.
```
getlarge, line 134: @BioPhoton and one extra section about release.

BioPhoton (approved):
Thx for the changes!

.github/workflows/release.yml
```diff
@@ -83,9 +83,9 @@ jobs:
           GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
         run: |
           echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
-          npx nx release --skipPublish
-          git push --follow-tags
-          npx nx release publish
+          npx nx release --yes
+      - name: Collect and upload Code PushUp report
+        run: npx nx code-pushup -- autorun
```
matejchalk, line 88: This is now being run by the subsequent `code-pushup/github-action` step.

```suggestion
```

matejchalk (changes requested):

.github/workflows/release.yml
getlarge: ✔️

matejchalk (approved):

package.json
matejchalk, line 1: I checked out your branch to check if the update broke anything (our CI still isn't triggered for forks :grimacing:), and ran into a problem with E2E tests (I'm on Linux):
>  Cannot find module @rollup/rollup-linux-x64-gnu. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.

Found the solution in this discussion https://github.com/vitejs/vite/discussions/15532:
```sh
npm install @rollup/rollup-linux-x64-gnu --save-optional
```

matejchalk (changes requested):

getlarge:
👋 see you in #639

PR #636, opened 2024-04-18
fix(plugin-js-packages): consider dynamic outdated fields for Yarn v1

Tlacenka:
Related to #597 

The number of fields in Yarn v1 outdated result is dynamic. I now calculate the index of each relevant field, check all required fields are present and use their information to provide the audit outputs.

Note: Was tested on `rx-angular`. Should be tested on the previously failing project cc @matejchalk

### DIFF
--- a/packages/plugin-js-packages/src/lib/package-managers/yarn-classic/outdated-result.ts
+++ b/packages/plugin-js-packages/src/lib/package-managers/yarn-classic/outdated-result.ts
@@ -1,16 +1,74 @@
-import { fromJsonLines } from '@code-pushup/utils';
-import { OutdatedResult } from '../../runner/outdated/types';
-import { Yarnv1OutdatedResultJson } from './types';
+import {
+  fromJsonLines,
+  objectFromEntries,
+  objectToEntries,
+  objectToKeys,
+} from '@code-pushup/utils';
+import {
+  OutdatedDependency,
+  OutdatedResult,
+} from '../../runner/outdated/types';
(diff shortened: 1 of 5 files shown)


packages/plugin-js-packages/src/lib/package-managers/yarn-classic/outdated-result.ts
matejchalk, line 1: > Should be tested on the previously failing project

Another problem remains, which is `exotic` versions not being handled, throws `Error: Invalid version description exotic`. The project uses `resolutions` in `package.json`.

image

Tlacenka: Thank you. That other problem is OOS for this PR. IT is mentioned within the checklist of issues of the #597 , planning to address it in another PR.

matejchalk (approved):

BioPhoton (commented):
>   throw new Error('Package is not outdated.');

This error could include more information e.g. which package it is. This makes it easier to debug the problem.

packages/plugin-js-packages/src/lib/package-managers/yarn-classic/outdated-result.ts
Tlacenka: Addressed in #656

Tlacenka:
https://github.com/code-pushup/cli/pull/636#pullrequestreview-2021705525

That code has been removed as part of #656.

matejchalk (approved):
Tested on the project Yarn 1 project and outdated packages are now correctly discovered thanks to this change.

PR #639, opened 2024-04-23
chore: upgrade to nx 18

### LINKED ISSUE #594: ci: fix invalid release configuration

getlarge:
Continue #595
Fixes #594
Supersedes #625


getlarge:
@BioPhoton @matejchalk It looks like there is a problem with the e2e tests setup:
- `cli-e2e` and `nx-plugin-e2e` declares the same `globalSetup` which is the relative path to `global-setup.e2e.ts`.
- in `global-setup.e2e.ts` the ephemeral package registry is created
- This means that when running the e2e tests in parallel, there are two registries on two different ports since Verdaccio increments the port number when one is busy.
- but the folder is the same for both...

Which leads to race conditions and conflicts with package publishing/installation.

getlarge:
Also, the e2e tests implicitly depend on Chromium being available in the OS, and believe it or not, some people do not have the Chrome browser installed.

BioPhoton:
> casualty

I have to set it as hard dependency it seems... thought lighthouse does it.

matejchalk:
@getlarge We've just fixed the [CI](https://github.com/code-pushup/cli/actions/workflows/ci.yml?query=branch%3Amain) and [Code PushUp](https://github.com/code-pushup/cli/actions/workflows/code-pushup.yml?query=branch%3Amain) workflows in `main`. So if you merge `main` into this branch, the workflows should pass and then we can merge this PR, hopefully :crossed_fingers:

getlarge:
> @getlarge We've just fixed the [CI](https://github.com/code-pushup/cli/actions/workflows/ci.yml?query=branch%3Amain) and [Code PushUp](https://github.com/code-pushup/cli/actions/workflows/code-pushup.yml?query=branch%3Amain) workflows in `main`. So if you merge `main` into this branch, the workflows should pass and then we can merge this PR, hopefully 🤞

Just did and also updated the e2e setup process. I'll add comments in a second to explain...

global-setup.e2e.ts
```diff
@@ -6,12 +6,22 @@ import stopLocalRegistry from './tools/scripts/stop-local-registry';
...
+    execSync(
+      `npm install -D @code-pushup/eslint-plugin@e2e --registry=${registry}`,
+    );
+    execSync(
+      `npm install -D @code-pushup/coverage-plugin@e2e --registry=${registry}`,
+    );
+    await setupTestFolder('tmp/e2e');
+  } catch (e) {
+    stopLocalRegistry();
+  }
```
getlarge, line 24: this way, we avoid leaving dozens of Verdaccio and Node instances in our machine :)

global-setup.e2e.ts
```diff
@@ -6,12 +6,22 @@ import stopLocalRegistry from './tools/scripts/stop-local-registry';
...
+    execSync(`npm install -D @code-pushup/cli@e2e --registry=${registry}`);
+    execSync(
+      `npm install -D @code-pushup/nx-plugin@e2e --registry=${registry}`,
+    );
+    execSync(
+      `npm install -D @code-pushup/eslint-plugin@e2e --registry=${registry}`,
+    );
+    execSync(
+      `npm install -D @code-pushup/coverage-plugin@e2e --registry=${registry}`,
+    );
```
getlarge, line 20: this is an attempt to remove the need of modifying the npm config

tools/scripts/start-local-registry.ts
```diff
@@ -2,35 +2,60 @@
...
+  if (!registry) {
+    console.error('Local registry not started');
+    process.exit(1);
+  }
+  global.stopLocalRegistry = stop;
+  global.registry = registry;
+  try {
+    const version = execSync('git describe --tags --abbrev=0')
+      .toString()
+      .trim();
```
getlarge, line 36: this enables getting the latest version of the packages (since versioning is tag-based and unique across all projects.)

tools/scripts/start-local-registry.ts
```diff
@@ -2,35 +2,60 @@
...
+      firstRelease: true,
+      generatorOptionsOverrides: {
+        skipLockFileUpdate: true,
+      },
+    });
+
+    await releasePublish({
+      tag: 'e2e',
+      registry,
+    });
```
getlarge, line 52: Adaptation of [the setup suggested by Nx](https://nx.dev/recipes/nx-release/update-local-registry-setup).
Define explicitly the registry to remove dependency on NPM config.

getlarge:
Apparently, the publishing is still flaky, I tried only once locally and the e2e tests were passing but not the second time. I got the same error as in the CI.
I will fix it now.

tools/scripts/start-local-registry.ts
```diff
@@ -48,63 +73,77 @@ function startLocalRegistry({
...
-    });
-  });
+          resolve({
+            registry,
+            stop: () => {
+              if (childProcess.pid) {
+                process.kill(-childProcess.pid);
+              }
+              // does not kill the underlying process, see https://github.com/nodejs/node/issues/46865
+              childProcess.kill();
```
getlarge, line 121: this was the second culprit why the local registry was not stopped properly; this is related to a UNIX design ~~bug~~ feature.

nx.json
```diff
@@ -88,28 +88,27 @@
     }
   },
   "release": {
-    "projects": ["packages/*", "!packages/nx-plugin"],
+    "projects": ["packages/*"],
```
getlarge, line 91: I enabled the release of the `nx-plugin` project, so it is published during the initialization of the local registry  (then it can be installed in `global-setup.e2e.ts`).
Is it actually ready to be published on NPM?

BioPhoton: We need to have docs cleaned then the MVP would already work... 

So no.

getlarge: All right, i reverted the change and removed steps for publishing and installing the `nx-plugin` in `global-setup.e2e.ts`.

getlarge:
I am stuck in the CI hole with this PR.
Do you happen to know why the [e2e tests are timing out on Windows](https://github.com/code-pushup/cli/actions/runs/8884643357/job/24394206268)? It looks like they were already slow on the previous PR but are now stuck during the publishing phase.

Also, the [Code PushUp workflow fails because a coverage file is missing](https://github.com/code-pushup/cli/actions/runs/8884643520/job/24394205838), is it an expected behavior when some tests would not run because the lib is not affected?

BioPhoton:
> Also, the [Code PushUp workflow fails because a coverage file is missing](https://github.com/code-pushup/cli/actions/runs/8884643520/job/24394205838), is it an expected behavior when some tests would not run because the lib is not affected?

It is an expected behaviour. If your tests fail the plugins fails. 

I personally dont like it and would like to have it pass with the passed tests :) But I see the reasoning behind this decision.

global-setup.e2e.ts
BioPhoton: Nice! So registry is now an URL to the current verdaccio server.

getlarge:
Apparently code coverage is not generated in my local environment. It concurs with the error in the CI

getlarge:
Thanks @BioPhoton for joining the debugging, we found out that the coverage is not generated due to [this issue](https://github.com/nrwl/nx/issues/21995).

getlarge:
@BioPhoton apparently, enabling coverage for all the tests left them hanging.
The[ CI jobs](https://github.com/code-pushup/cli/actions/runs/8930170847) kept running forever.
Yet another Vitest surprise!

getlarge:
@BioPhoton @matejchalk I've got news regarding Nx Vitest executor.
The executor now correctly handles the Vitest flags since this [PR](https://github.com/nrwl/nx/pull/22355) was merged and released.
I upgraded Nx to 18.3.4 and confirmed it works. I also noticed that the tests are left hanging when coverage is enabled and the test suite is empty! We could remove those tests' configurations, right?

One bad thing with the upgrade though, is that many unit tests fail for the ESLint plugin.

vmasek:
Amazing, these fails seem like some kind of cache access lock 🤔 
```
Error: EACCES: permission denied, mkdir '/test/.nx/cache'
   ❯ mkdirSync ../../node:fs:1380:26
   ❯ Function.Module._load ../../node_modules/nx/src/native/index.js:61:7
   ❯ Object.<anonymous> ../../node_modules/nx/src/native/native-bindings.js:184:31
   ❯ Function.Module._load ../../node_modules/nx/src/native/index.js:67:25
  
  ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
  Serialized Error: { errno: -13, code: 'EACCES', syscall: 'mkdir', path: '/test/.nx/cache' }
  ```

**Comment by getlarge** [comment]
> Amazing, these fails seem like some kind of cache access lock 🤔
> 
> ```
> Error: EACCES: permission denied, mkdir '/test/.nx/cache'
>    ❯ mkdirSync ../../node:fs:1380:26
>    ❯ Function.Module._load ../../node_modules/nx/src/native/index.js:61:7
>    ❯ Object.<anonymous> ../../node_modules/nx/src/native/native-bindings.js:184:31
>    ❯ Function.Module._load ../../node_modules/nx/src/native/index.js:67:25
>   
>   ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
>   Serialized Error: { errno: -13, code: 'EACCES', syscall: 'mkdir', path: '/test/.nx/cache' }
> ```

locally I get another FS error code:
```
ENOENT: no such file or directory, mkdir '/test/.nx/cache'  
```

In this test suite, i noticed that `process.cwd` is mocked to return the `/test` path, which could maybe interfere with some internal Nx functions... I'll keep on digging.

getlarge:
The other bad news is that the [e2e setup still hangs for the Windows environment](https://github.com/code-pushup/cli/actions/runs/9083997938/job/24963954109).
**Reminder**, I might push a test commit that restores the previous process of publishing packages to the local registry.

vmasek:
> > Amazing, these fails seem like some kind of cache access lock 🤔
> > ```
> > Error: EACCES: permission denied, mkdir '/test/.nx/cache'
> >    ❯ mkdirSync ../../node:fs:1380:26
> >    ❯ Function.Module._load ../../node_modules/nx/src/native/index.js:61:7
> >    ❯ Object.<anonymous> ../../node_modules/nx/src/native/native-bindings.js:184:31
> >    ❯ Function.Module._load ../../node_modules/nx/src/native/index.js:67:25
> >   
> >   ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
> >   Serialized Error: { errno: -13, code: 'EACCES', syscall: 'mkdir', path: '/test/.nx/cache' }
> > ```
> 
> locally I get another FS error code:
> 
> ```
> ENOENT: no such file or directory, mkdir '/test/.nx/cache'  
> ```
> 
> In this test suite, i noticed that `process.cwd` is mocked to return the `/test` path, which could maybe interfere with some internal Nx functions... I'll keep on digging.

I see mkdir

vmasek:
Sorry for closing, that was miss click 😅

I am wondering if it is allowed to make directories or add files (`mkdir`) inside of a GH action runtime inside of a `/tests/` folder

getlarge:
> The other bad news is that the [e2e setup still hangs for the Windows environment](https://github.com/code-pushup/cli/actions/runs/9083997938/job/24963954109). **Reminder**, I might push a test commit that restores the previous process of publishing packages to the local registry.

After reverting to using the previous `publish` target, it looks like the [CI workflow is going to hang during the Windows e2e job](https://github.com/code-pushup/cli/actions/runs/9187386525/job/25264968018).
I looked at many past workflow runs to compare, and I could not find a single job run for E2E-test (windows-latest) that actually runs the build. They all seem to pick the build output from the cache.

**Workflows**:
- https://github.com/code-pushup/cli/actions/runs/9175838368/job/25229713877
- https://github.com/code-pushup/cli/actions/runs/9161552409/job/25186658890
- https://github.com/code-pushup/cli/actions/runs/9158674616/job/25177512068
- https://github.com/code-pushup/cli/actions/runs/9155531208/job/25168089689

Now, I wonder if the issue is related to `cli:build` task and not the registry setup.

getlarge:
In the CI workflow, I saw the `build` job is not using a strategy matrix, maybe this should change?

BioPhoton:
Hi @getlarge imo the scope of the PR was a bit too ambitious.

Just focusing on the update and not trying to refactor testing and use new APIs etc would have made it easier to go on.
Furthermore, there is now another PR on Nx 19 open. 

What should we do here? 
Keeping it hanging around is I guess counter productive.

Is there information inside that we can add as a comment to the new PR or anything that you think helps?

Let me know how to go on. And what you think. 🙏

getlarge:
> Hi @getlarge imo the scope of the PR was a bit too ambitious.
> 
> Just focusing on the update and not trying to refactor testing and use new APIs etc would have made it easier to go on. Furthermore, there is now another PR on Nx 19 open.
> 
> What should we do here? Keeping it hanging around is I guess counter productive.
> 
> Is there information inside that we can add as a comment to the new PR or anything that you think helps?
> 
> Let me know how to go on. And what you think. 🙏

Hey, I guess it’s time to close it then!
I don’t think this was ambitious, at the time the  real blocker was the windows e2e tests which were left hanging and I never found (the time to find) the reason.
You encountered a similar situation in the Verdaccio setup repository right?

BioPhoton:
Yes let's close it!

PR #651, opened 2024-04-25
perf(plugin-eslint): lint nx projects in separate processes

### LINKED ISSUE #629: Run ESLint separately on Nx projects
### Motivation

(For more context, see https://github.com/code-pushup/cli/issues/376#issuecomment-2049696673 and subsequent discussion.)

Some Nx monorepos have hundreds of projects, but still want to run Code PushUp on all projects at once. The ESLint plugin's approach to running a single lint process from a combined config (created by `eslintConfigFromNxProjects()` helper) causes out of memory errors in such cases. Linting the projects separately similarly to `nx run-many -t lint` would likely prevent such excessive memory usage.

### Proposal

Usage of the `eslintConfigFromNxProjects` Nx helper would be the same, but the resulting arguments passed to `eslintPlugin` initializer would be different.

The usage would remain:

```ts
import eslintPlugin, {
  eslintConfigFromNxProjects,
} from '@code-pushup/eslint-plugin';

export default {
  plugins: [
     await eslintPlugin(await eslintConfigFromNxProjects()),
  ],
}
```

But instead of this being equivalent to:

```ts
import eslintPlugin from '@code-pushup/eslint-plugin';

export default {
  plugins: [
     await eslintPlugin({
       eslintrc: {
         root: true,
         overrides: [
            { files: ['apps/client/**/*'], extends: ['./apps/client/.eslintrc.json'] },
            { files: ['apps/server/**/*'], extends: ['./apps/server/.eslintrc.json'] },
            { files: ['libs/shared/**/*'], extends: ['./libs/shared/.eslintrc.json'] },
         ]
       },
       patterns: ['apps/client/**/*', 'apps/server/**/*', 'libs/shared/**/*']
     }),
  ],
}
```

it would be equivalent to:

```ts
import eslintPlugin from '@code-pushup/eslint-plugin';

export default {
  plugins: [
     await eslintPlugin([
       { eslintrc: './apps/client/.eslintrc.json', patterns: ['apps/client/**/*'] },
       { eslintrc: './apps/server/.eslintrc.json', patterns: ['apps/server/**/*'] },
       { eslintrc: './libs/shared/.eslintrc.json', patterns: ['libs/shared/**/*'] },
     ]),
  ],
}
```

This means the `eslintPlugin` function would also accept an array of `{ eslintrc: string | ESLint.ConfigData; patterns: string | string[]  }` objects instead of only a single object. Each array item (i.e. Nx project's lint configuration) would be processed separately, but the audits and results would still be combined in the end.

matejchalk:
Closes #629 

Tested by executing plugin from `dist` on [code-pushup/portal](https://github.com/code-pushup/portal). Previously, this would crash without `NODE_OPTIONS=--max-old-space-size=8192`. Now it successfully completes without any memory increase.


packages/plugin-eslint/src/lib/runner/lint.ts
```diff
@@ -1,44 +1,101 @@
...
+    return await fn(configPath);
+  } finally {
+    await rm(configPath);
+  }
+}
+
+function generateTempConfigPath(): string {
+  return join(
+    process.cwd(),
+    `.eslintrc.${Math.random().toString().slice(2)}.json`,
```
Tlacenka, line 99: ❔ Do I understand it correctly that the limit for the number of projects is 999? That should probably be documented somewhere.

Tlacenka (approved):
I like this improvement a lot!

I looked both at the plugin code and the tests.
While there are some generic tests and bigger snapshots of higher complexity, their refactor does not belong to this PR.

packages/plugin-eslint/src/lib/runner/lint.ts
Tlacenka: Ah! I misunderstood the `slice`, thought it uses only first three digits not that it uses all digits after the first 3, hence the question. My bad. 😅

vmasek (approved):

BioPhoton (approved):

PR #656, opened 2024-04-25
refactor(plugin-js-packages): use semver for version diff

Tlacenka:
Related to #597 

In this PR, I use [`semver` functions](https://github.com/npm/node-semver?tab=readme-ov-file#functions) to determine the version difference.

### DIFF
--- a/packages/plugin-js-packages/src/lib/runner/outdated/constants.ts
+++ b/packages/plugin-js-packages/src/lib/runner/outdated/constants.ts
@@ -1,8 +1,16 @@
+import type { ReleaseType } from 'semver';
 import type { IssueSeverity } from '@code-pushup/models';
-import { VersionType } from './types';
+import { objectToKeys } from '@code-pushup/utils';
 
-export const outdatedSeverity: Record<VersionType, IssueSeverity> = {
+export const outdatedSeverity: Record<ReleaseType, IssueSeverity> = {
   major: 'error',
+  premajor: 'info',
   minor: 'warning',
+  preminor: 'info',
   patch: 'info',
+  prepatch: 'info',
+  prerelease: 'info',
 };
+
+// RELEASE_TYPES directly exported from semver don't work out of the box
+export const RELEASE_TYPES = objectToKeys(outdatedSeverity);
--- a/packages/plugin-js-packages/src/lib/runner/outdated/transform.ts
+++ b/packages/plugin-js-packages/src/lib/runner/outdated/transform.ts
@@ -1,14 +1,10 @@
-import { Issue } from '@code-pushup/models';
-import { pluralize } from '@code-pushup/utils';
+import { ReleaseType, clean, diff, neq } from 'semver';
+import type { Issue } from '@code-pushup/models';
+import { objectFromEntries, pluralize } from '@code-pushup/utils';
 import { DependencyGroup, PackageManagerId } from '../../config';
 import { dependencyGroupToLong } from '../../constants';
-import { outdatedSeverity } from './constants';
-import {
-  OutdatedResult,
-  PackageVersion,
-  VersionType,
-  versionType,
-} from './types';
+import { RELEASE_TYPES, outdatedSeverity } from './constants';
+import { OutdatedResult } from './types';
 
 export function outdatedResultToAuditOutput(
   result: OutdatedResult,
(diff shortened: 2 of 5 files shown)


packages/plugin-js-packages/src/lib/runner/outdated/transform.ts
```diff
@@ -18,23 +14,27 @@ export function outdatedResultToAuditOutput(
...
-    );
 
-  const outdatedStats = outdatedDependencies.reduce(
-    (acc, dep) => {
-      const outdatedLevel = getOutdatedLevel(dep.current, dep.latest);
-      return { ...acc, [outdatedLevel]: acc[outdatedLevel] + 1 };
-    },
-    { major: 0, minor: 0, patch: 0 },
+  const validDependencies = relevantDependencies
+    .filter(dep => parse(dep.current) != null && parse(dep.latest) != null)
```
BioPhoton, line 19: Could you comment what you filter out pls

BioPhoton (dismissed):

BioPhoton:
> Can I get rid of the non-null-assertion? I use semver functions to make sure to filter out null values and up-to-date dependencies but TS does not know.

Where? Maybe a guard helps?

packages/plugin-js-packages/src/lib/runner/outdated/transform.ts
```diff
@@ -18,23 +14,27 @@ export function outdatedResultToAuditOutput(
...
+    .map(dep => ({
+      ...dep,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      current: clean(dep.current)!,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      latest: clean(dep.latest)!,
+    }));
```
matejchalk, line 26: > Can I get rid of the non-null-assertion? I use semver functions to make sure to filter out null values and up-to-date dependencies but TS does not know.

You can use a type predicate if you swap the order - first call `clean()` in `.map()`, then check both are non-null (i.e. valid) in `.filter()`. According to the docs, both `parse()` and `clean()` return `null` for invalid versions, so don't need to call both.

```suggestion
  const validDependencies = relevantDependencies
    .map(dep => ({
      ...dep,
      current: clean(dep.current),
      latest: clean(dep.latest),
    }))
    .filter(
      (dep): dep is OutdatedResult[number] =>
        dep.current != null && dep.latest != null,
    );
```

packages/plugin-js-packages/src/lib/runner/outdated/transform.ts
```diff
@@ -18,23 +14,27 @@ export function outdatedResultToAuditOutput(
...
+
+  const outdatedDependencies = validDependencies.filter(dep =>
+    neq(dep.current, dep.latest),
   );
 
+  const outdatedStats = outdatedDependencies.reduce((acc, dep) => {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const outdatedLevel = diff(dep.current, dep.latest)!;
+    return { ...acc, [outdatedLevel]: acc[outdatedLevel] + 1 };
+  }, objectFromEntries(RELEASE_TYPES.map(versionType => [versionType, 0])));
```
matejchalk, line 36: The docs say that `diff()` returns `null` if the versions are the same. That's probably not gonna happen in `npm outdated`'s output. But to be on the safe-side, such `diff()` results can easily be ignored:

```suggestion
  const outdatedStats = outdatedDependencies.reduce((acc, dep) => {
    const outdatedLevel = diff(dep.current, dep.latest);
    if (outdatedLevel == null) {
      return acc;
    }
    return { ...acc, [outdatedLevel]: acc[outdatedLevel] + 1 };
  }, objectFromEntries(RELEASE_TYPES.map(versionType => [versionType, 0])));
```

packages/plugin-js-packages/src/lib/runner/outdated/constants.ts
```diff
@@ -1,8 +1,12 @@
+import { ReleaseType } from 'semver';
 import type { IssueSeverity } from '@code-pushup/models';
-import { VersionType } from './types';
 
-export const outdatedSeverity: Record<VersionType, IssueSeverity> = {
+export const outdatedSeverity: Record<ReleaseType, IssueSeverity> = {
   major: 'error',
+  premajor: 'error', // TODO ??
   minor: 'warning',
+  preminor: 'warning',
```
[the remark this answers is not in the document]
Tlacenka: Makes sense.

packages/plugin-js-packages/src/lib/runner/outdated/transform.ts
Tlacenka: Nice, thank you

packages/plugin-js-packages/src/lib/runner/outdated/transform.ts
Tlacenka: I updated the code.

packages/plugin-js-packages/src/lib/runner/outdated/transform.ts
```diff
@@ -43,7 +43,7 @@ export function outdatedResultToAuditOutput(
   return {
     slug: `${packageManager}-outdated-${depGroup}`,
     score: calculateOutdatedScore(
-      outdatedStats.major,
+      outdatedStats.major + outdatedStats.premajor,
```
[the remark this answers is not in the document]
Tlacenka: Addressed.

matejchalk (approved):
Good work, looks very robust :+1:

Issue #647, opened 2024-04-24
Issues with coverage plugin


Tlacenka:
- [x] empty-report functions in function coverage. At least remove the name and change it to anonymous.
![image](https://github.com/code-pushup/cli/assets/6306671/352ddb5d-3a93-4f87-9d67-1ca0f01e674d)
- [x] In some cases, the line range does not work as expected (maximum consecutive range). There are then separate issues for consecutive lines.
![image](https://github.com/code-pushup/cli/assets/6306671/bcfc657c-8fd0-4517-a256-91fa38ef78c6)
- [x] Some issues are duplicated.
  <details> <summary> cli.ts </summary>

  ![image](https://github.com/code-pushup/cli/assets/6306671/44cfb82e-6485-40f3-b7e9-9e2658723ee5)
  </details

Tlacenka:
Duplicate issues have a root cause explained here https://github.com/code-pushup/cli/issues/560#issuecomment-2135630056

PR #658, opened 2024-04-29
fix(plugin-coverage): sort lines numerically to merge consecutive ranges

Tlacenka:
Related to #647 

I apply numerical sort instead of lexicographical one, so that the lines are correctly merged. Additional test added.


matejchalk (approved):
Well done :+1:

Issue #637, opened 2024-04-22
feat(models): improve table schema and parsing logic


BioPhoton:
### User story

> derived from https://github.com/code-pushup/cli/pull/630#discussion_r1574635336

As a developer I want to have good typing and types and function arguments with intuitive and types that have good DX.

The table type lacks in the following parts:
- no generic type 
- alignment separated from headings adds complexity

### Acceptance criteria

- [x] the `Table` type merges the `headings` and. `alignments`
- [x] the parse result of the `tableSchema` produces a generic typed output
  - [x] Rows are derived from the input

### Implementation details

**Schemas:**

```ts
import { z } from 'zod';

export const tableHeadingSchema = z.object(
  {
    key: z.string(),
    label: z.string().optional(),
    alignment: z.enum(['l', 'c', 'r']).optional(),
  },
  { description: 'Table heading' },
);
export type TableHeading = z.infer<typeof tableHeadingSchema>;

export const tableRowSchema = z.union([
  z.array(primitiveValueSchema),
  z.record(primitiveValueSchema),
]);
export type TableRow = z.infer<typeof tableRowSchema>;

export const tableSchema = (description = 'Table information') =>
  z.object(
    {
      headings: z.array(tableHeadingSchema).optional(),
      rows: z.array(tableRowSchema),
    },
    { description },
  );
export type Table = z.infer<ReturnType<typeof tableSchema>>;

```

**Parser:**

```ts

export function tableSchemaParse<T>() {

}
```

matejchalk:
Your example misses the merged `headings` and `alignment` into `columns`:

```ts
export const tableSchema = (description = 'Table information') =>
  z.object(
    {
      headings: z.array(tableHeadingSchema).optional(),
      alignment: z.array(tableAlignmentSchema).optional(),
      rows: z.array(tableRowSchema),
    },
    { description },
  );
```

You need to add a union of the two column types:

```ts
const tableColumnsSchema = z.union([
  z.array(tableHeadingSchema),
  z.array(tableAlignmentSchema),
]);

export const tableSchema = (description = 'Table information') =>
  z.object(
    {
      rows: z.array(tableRowSchema),
      columns: tableColumnsSchema.optional(),
    },
    { description },
  );

```

vmasek:
Drop the generic types for now

PR #667, opened 2024-05-02
refactor: improve table model

BioPhoton:


packages/models/src/lib/table.ts
```diff
@@ -1,20 +1,21 @@
 import { z } from 'zod';
 import { primitiveValueSchema } from './implementation/schemas';
 
+export const tableAlignmentSchema = z.enum(['l', 'c', 'r'], {
+  description: 'Cell alignment (l:left, r:right, c:center)',
+});
+export type TableAlignment = z.infer<typeof tableAlignmentSchema>;
```
matejchalk, line 7: Why not just make the enum values self-descriptive?

```suggestion
export const tableAlignmentSchema = z.enum(['left', 'center', 'right'], {
  description: 'Cell alignment',
});
export type TableAlignment = z.infer<typeof tableAlignmentSchema>;
```

packages/models/src/lib/table.ts
```diff
@@ -1,20 +1,21 @@
...
+
 export const tableHeadingSchema = z.object(
   {
-    key: z.string(),
+    key: z.string().optional(),
     label: z.string().optional(),
+    align: tableAlignmentSchema.optional(),
   },
   { description: 'Table heading' },
 );
```
matejchalk, line 16: Making all properties optional enables empty objects (e.g. `[{}, {}, {}]` or `[{}, { key: 'foo' }, {}]`), which I wouldn't consider valid.

My suggestion in https://github.com/code-pushup/cli/issues/637#issuecomment-2071704785 would have `key` always required, but also supports string array for setting only alignments. Configuration examples would be:
- only keys: `[{ key: 'foo' }, { key: 'bar' }]`
- with labels: `[{ key: 'foo', label: 'Foo' }, { key: 'bar', label: 'Foo' }]`
- with labels and alignment: `[{ key: 'foo', label: 'Foo', align: 'left' }, { key: 'bar', label: 'Foo', align: 'center' }]`
- only alignment: `['left', 'center']`

packages/models/src/lib/table.ts
```diff
@@ -25,7 +26,6 @@ export const tableSchema = (description = 'Table information') =>
   z.object(
     {
       headings: z.array(tableHeadingSchema).optional(),
-      alignment: z.array(tableAlignmentSchema).optional(),
       rows: z.array(tableRowSchema),
     },
```
matejchalk, line 30: I'd suggest renaming `headings` property to `columns`. It now supports specifying alignment, so it doesn't only affect how the heading is rendered. Also, I think it's more elegant to have `rows` and `columns` naming, as these terms are commonly used together.

packages/models/src/lib/table.unit.test.ts
```diff
@@ -33,37 +33,37 @@ describe('tableSchema', () => {
...
-      alignment: ['l'],
+      headings: [{ key: 'metrics', label: 'Metrics Name', align: 'l' }],
     };
     expect(() => tableSchema().parse(table)).not.toThrow();
   });
 
   it('should parse complete table', () => {
     const fullTable: Table = {
       headings: [
+        // center oftern is the default when rendering in MD or HTML
```
matejchalk, line 63: Is this what you meant?

```suggestion
        // center is often the default when rendering in MD or HTML
```

packages/utils/src/lib/reports/generate-md-report.ts
```diff
@@ -139,7 +135,21 @@ export function aboutSection(
 
 export function reportPluginMeta({ plugins }: Pick<Report, 'plugins'>) {
   return {
-    headings: pluginMetaTableHeaders,
+    headings: [
+      {
+        key: 'plugin',
+        align: 'l' as const,
```
matejchalk, line 141: You wouldn't need the `as const` here if the function used the `Table` return type. An explicit function return type will also give you better autocomplete in the `return` statement.

packages/models/src/lib/table.ts
BioPhoton: renamed

packages/models/src/lib/table.ts
BioPhoton: fixed

BioPhoton: done

packages/utils/src/lib/reports/generate-md-report.ts
BioPhoton: removed

packages/models/src/lib/table.ts
BioPhoton: renamed

packages/models/src/lib/table.ts
```diff
@@ -1,33 +1,64 @@
...
-  z.object(
-    {
-      headings: z.array(tableHeadingSchema).optional(),
-      alignment: z.array(tableAlignmentSchema).optional(),
-      rows: z.array(tableRowSchema),
-    },
+  z.union(
+    [tablePrimitiveSchema, tableObjectColPSchema, tableObjectColOSchema],
     { description },
   );
```
matejchalk, line 63: Would be simpler to merge `tableObjectColPSchema` and `tableObjectColOSchema` into one schema:

```ts
const tableObjectSchema = z.object(
  {
    columns: z.union([
      z.array(tableColumnObjectSchema),
      z.array(tableColumnPrimitiveSchema),
    ]).optional(),
    rows: z.array(tableRowObjectSchema),
  },
  {
    description:
      'Table with object rows and optional alignment or object columns',
  },
);
```

packages/models/src/lib/table.ts
```diff
@@ -1,33 +1,64 @@
...
+const tableObjectColOSchema = z.object(
+  {
+    columns: z.array(tableColumnObjectSchema).optional(),
+    rows: z.array(tableRowObjectSchema),
+  },
+  {
+    description:
+      'Table with object rows and optional alignment or object columns',
+  },
+);
```
matejchalk, line 57: Both descriptions are the same, despite each schema only supporting one kind of column exclusively. But I think merging them is actually the best solution - see comment above.

packages/models/src/lib/table.unit.test.ts
```diff
@@ -1,5 +1,84 @@
...
+} from './table';
+
+describe('tableAlignmentSchema', () => {
+  it('should accept a valid enum', () => {
+    const alignment: TableAlignment = 'center';
+    expect(() => tableAlignmentSchema.parse(alignment)).not.toThrow();
+  });
+
+  it('should throw for a invalid enum', () => {
+    const alignment = 'crooked';
```
matejchalk, line 24: Good example :laughing:

packages/models/src/lib/table.ts
BioPhoton: merged them

packages/models/src/lib/implementation/schemas.unit.test.ts
```diff
@@ -0,0 +1,14 @@
+import { describe, expect, it } from 'vitest';
+import { PrimitiveValue, primitiveValueSchema } from './schemas';
+
+describe('tableAlignmentSchema', () => {
```
matejchalk, line 4:
```suggestion
describe('primitiveValueSchema', () => {
```

packages/models/src/lib/table.ts
```diff
@@ -1,33 +1,52 @@
...
+
+const tablePrimitiveSchema = z.object(
+  {
+    columns: z.array(tableAlignmentSchema).optional(),
+    rows: z.array(tableRowPrimitiveSchema),
+  },
+  { description: 'Table with primitive rows and optional alignment columns' },
+);
+
+const tableObjectColumnSchema = z.object(
```
matejchalk, line 37:
```suggestion
const tableObjectSchema = z.object(
```

matejchalk (approved):

PR #673, opened 2024-05-20
refactor(utils):  relocate code for text styles

BioPhoton:


packages/utils/src/lib/text-formats/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const FONT_STYLES = {
+  italic: 'italic',
+  bold: 'bold',
+  'strike through': 'strike-through',
```
BioPhoton, line 4: Maybe we use "strikeThrough" or "strike-through" as key?
```suggestion
  'strikeThrough': 'strike-through',
```

packages/utils/src/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './lib/text-formats/index';
```
matejchalk, line 1:
```suggestion
export * from './lib/text-formats';
```

packages/utils/src/lib/text-formats/html/font-style.ts
matejchalk, line 1: I think it would be more intuitive to split `fontStyle` into individual style functions. The usage would be more transparent, I think, e.g. bold isn't an obvious default. And we almost always use a single style, so the styles array isn't something we take advantage of.

- `fontStyle('...')` :arrow_right: `bold('...')` - shorter and more transparent
- `fontStyle('...', ['code']) :arrow_right: `code('...')` - shorter and no more single-item arrays
- `fontStyle('...', ['bold', 'italic']) :arrow_right: `italic(bold('...'))` - rarely used anyway, but order more intuitive

packages/utils/src/lib/text-formats/constants.ts
matejchalk: Why even have separate keys and values? How about this:

```ts
export const FONT_STYLES = ['bold', 'italic', 'strike-through'] as const;
```

You can still infer the type:

```ts
export type FontStyle = (typeof FONT_STYLES)[number];
```

packages/utils/src/lib/text-formats/html/font-style.ts
BioPhoton: splitted

packages/utils/src/lib/text-formats/constants.ts
BioPhoton: removed font style all at ones

matejchalk (approved):

PR #676, opened 2024-05-21
feat(models): add optional table title

matejchalk:
Motivation: possible to specify what data the table contains (e.g. _Most frequent errors_).

### DIFF
--- a/packages/utils/src/lib/reports/__snapshots__/generate-md-report.unit.test.ts.snap
+++ b/packages/utils/src/lib/reports/__snapshots__/generate-md-report.unit.test.ts.snap
@@ -19,8 +19,6 @@ exports[`auditDetails > should render complete details section 1`] = `
 "<details>
 <summary>🟩 <b>190ms</b> (score: 99)</summary>
 
-#### Additional Information
-
 |Class Names|Element|
 |:--:|:--:|
 |.btn, .icon|button|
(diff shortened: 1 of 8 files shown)


packages/utils/src/lib/reports/__snapshots__/generate-md-report.unit.test.ts.snap
```diff
@@ -106,8 +104,6 @@ exports[`generateMdReport > should render complete md report 1`] = `
 <details>
 <summary>🟨 <b>2,7 s</b> (score: 67)</summary>
 
-#### Additional Information
```
Tlacenka, line 109: 🎨 UX consistency: For the sake of consistency of the MD hierarchy, I think we should consider a fallback default value.

matejchalk: It's at the top of the expanded details, so I don't think it needs a heading. And I don't think a non-descriptive heading is useful. The heading should tell me what's in the table, and we can't do that generically in any meaningful way.

Tlacenka: Shouldn't the title then be required if I need the title to know what's in the table?

Tlacenka (approved):
Nice improvement! 🎉

packages/utils/src/lib/reports/__snapshots__/generate-md-report.unit.test.ts.snap
matejchalk: Unfortunately that's not always possible. E.g. the tables from Lighthouse don't have titles, AFAIK.

But I would recommend plugin authors specify a title if possible.

PR #678, opened 2024-05-22
feat(plugin-eslint): eslint config from nx projects helpers

### LINKED ISSUE #675: Nx helper for standalone project config
### User story

Feedback from @markusnissl on [Slack](https://flowup.slack.com/archives/C06JCR9L1FE/p1715895409416549):

> Hi,
> we just noticed that eslintConfigFromNxProject also scans dependencies. In my opinion this does not make sense if I only want to run it for a specific project, where I maybe don’t care about the other projects (other team, etc.).
> What is the intuition to always also check the used dependencies with this function?

...

> Not sure if I get it, as we run code pushup for each project in nx individually, we just don't need to run it also for projects that are dependencies of this project as they are analyzed on its own
> 
> https://github.com/code-pushup/cli/blob/main/packages/plugin-eslint/src/lib/nx/find-project-with-deps.ts
> 
> so I think the following changes would be sufficient:
> 
> ```ts
> export async function eslintConfigFromNxProject(
>   projectName: string,
> )
> ```
> 
> to
> 
> ```ts
> export async function eslintConfigFromNxProject(
>   projectName: string,
>   withDeps: boolean = true /* (true for backward compability)*/
> )
> ```
> 
> and in line 41 `(project.name === projectName || dependencies.has(project.name))` to `(project.name === projectName || (withDeps && dependencies.has(project.name)))`, then we just have to call it with `eslintConfigFromNxProject('some-name', false)`

Background from @matejchalk:

> The reasoning behind it is that configuring a single project doesn't really need a helper function, because the `eslintrc` and `patterns` can be passed directly. The Nx helpers are intended for use cases where multiple Nx projects should be targeted - either all projects (`eslintConfigFromNxProjects()`) or all dependencies of a specific project (`eslintConfigFromNxProject('<name>')`). Manual configuration would be very inconvenient in such cases.

### Acceptance criteria

It would be best to re-purpose `eslintConfigFromNxProject` to be a standalone project config. The project + dependencies scenario would be moved to `eslintConfigFromNxProjects` - project name would be an optional parameter, default would be to list all projects. Then it should cover all use cases, and the singular vs plural function name would indicate if multiple projects are used or not.

- standalone scenario:
  - before: `{ eslintrc: 'path/to/project/.eslintrc.json', patterns: 'path/to/project/**/*' }`
  - after: `await eslintConfigFromNxProject('project')`
- project + dependencies scenarios:
  - before: `await eslintConfigFromNxProject('project')`
  - after: `await eslintConfigFromNxProjectAndDeps('project')`
- all projects scenarios:
  - before & after: `await eslintConfigFromAllNxProjects()`

vmasek:
fixes #675 

Desired helpers API after the changes:
- [x] eslintConfigFromAllNxProjects() - includes all deps of all projects
- [x] eslintConfigFromNxProjectAndDeps('project') - includes all deps of provided project target
- [x] eslintConfigFromNxProject('project') - includes just the project target

### DIFF
--- a/packages/plugin-eslint/src/lib/nx/find-project-without-deps.ts
+++ b/packages/plugin-eslint/src/lib/nx/find-project-without-deps.ts
@@ -1,17 +1,16 @@
 import type { ESLintTarget } from '../config';
 import { nxProjectsToConfig } from './projects-to-config';
-import { findAllDependencies } from './traverse-graph';
 
 /**
- * Accepts a target Nx projects, finds projects it depends on, and converts lint configurations to Code PushUp ESLint plugin parameters.
+ * Accepts a target Nx projects, converts lint configurations to Code PushUp ESLint plugin parameters.
  *
  * Use when you wish to include a targeted subset of your Nx monorepo in your Code PushUp project.
- * If you prefer to include all Nx projects, refer to {@link eslintConfigFromAllNxProjects} instead.
- * if you'd like to skip dependencies of the provided target project use {@link eslintConfigFromNxProject} instead.
+ * If you prefer to include all Nx projects, refer to {@link eslintConfigFromAllNxProjects} instead,
+ * if you'd like to auto include all dependencies of the provided target project use {@link eslintConfigFromNxProjectAndDeps} instead.
  *
  * @example
  * import eslintPlugin, {
- *   eslintConfigFromNxProjectAndDeps,
+ *   eslintConfigFromNxProject,
  * } from '@code-pushup/eslint-plugin';
  *
  * const projectName = 'backoffice'; // <-- name from project.json
@@ -19,7 +18,7 @@ import { findAllDependencies } from './traverse-graph';
  * export default {
  *   plugins: [
  *     await eslintPlugin(
- *       await eslintConfigFromNxProjectAndDeps(projectName)
+ *       await eslintConfigFromNxProject(projectName)
  *     )
  *   ]
  * }
@@ -27,18 +26,14 @@ import { findAllDependencies } from './traverse-graph';
  * @param projectName Nx project serving as main entry point
  * @returns ESLint config and patterns, intended to be passed to {@link eslintPlugin}
  */
-export async function eslintConfigFromNxProjectAndDeps(
+export async function eslintConfigFromNxProject(
   projectName: string,
 ): Promise<ESLintTarget[]> {
   const { createProjectGraphAsync } = await import('@nx/devkit');
   const projectGraph = await createProjectGraphAsync({ exitOnError: false });
 
-  const dependencies = findAllDependencies(projectName, projectGraph);
-
   return nxProjectsToConfig(
     projectGraph,
-    project =>
-      !!project.name &&
-      (project.name === projectName || dependencies.has(project.name)),
+    project => !!project.name && project.name === projectName,
   );
 }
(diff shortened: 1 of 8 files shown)


vmasek:
I've deprecated the old name for the helper and now I'm getting lint errors that I'm exporting deprecated fn 😄 
image

Do I just ignore that with comment? I do that on purpose to not break the API unnecessarily

packages/plugin-eslint/src/lib/nx/find-project-without-deps.ts
```diff
@@ -0,0 +1,39 @@
...
+  projectName: string,
+): Promise<ESLintTarget[]> {
+  const { createProjectGraphAsync } = await import('@nx/devkit');
+  const projectGraph = await createProjectGraphAsync({ exitOnError: false });
+
+  return nxProjectsToConfig(
+    projectGraph,
+    project => !!project.name && project.name === projectName,
+  );
+}
```
matejchalk, line 39: Since there can only be one matching project, this could return a single lint target instead of array. Would also be useful to inform the user if the Nx project name is wrong.

```suggestion
export async function eslintConfigFromNxProject(
  projectName: string,
): Promise<ESLintTarget> {
  const { createProjectGraphAsync } = await import('@nx/devkit');
  const projectGraph = await createProjectGraphAsync({ exitOnError: false });

  const [project] = nxProjectsToConfig(
    projectGraph,
    project => !!project.name && project.name === projectName,
  );
  
  if (!project) {
    throw new Error(`Couldn't find Nx project named "${projectName}"`);
  }
  return project;
}
```

packages/plugin-eslint/src/lib/nx/find-project-without-deps.ts
```diff
@@ -0,0 +1,39 @@
+import type { ESLintTarget } from '../config';
+import { nxProjectsToConfig } from './projects-to-config';
+
+/**
+ * Accepts a target Nx projects, converts lint configurations to Code PushUp ESLint plugin parameters.
+ *
+ * Use when you wish to include a targeted subset of your Nx monorepo in your Code PushUp project.
+ * If you prefer to include all Nx projects, refer to {@link eslintConfigFromAllNxProjects} instead,
+ * if you'd like to auto include all dependencies of the provided target project use {@link eslintConfigFromNxProjectAndDeps} instead.
```
matejchalk, line 9:
```suggestion
 * Accepts a target Nx project, converts its lint configuration to Code PushUp ESLint plugin parameters.
 *
 * Use when you wish to only have a single Nx project as your Code PushUp project, without any other dependencies.
 * If you prefer to include all Nx projects, refer to {@link eslintConfigFromAllNxProjects} instead.
 * If you'd like to auto include all dependencies of the provided target project use {@link eslintConfigFromNxProjectAndDeps} instead.
```

packages/plugin-eslint/src/lib/nx/find-project-without-deps.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * export default {
+ *   plugins: [
+ *     await eslintPlugin(
+ *       await eslintConfigFromNxProject(projectName)
+ *     )
+ *   ]
+ * }
+ *
+ * @param projectName Nx project serving as main entry point
```
matejchalk, line 26:
```suggestion
 * @param projectName Nx project name
```

matejchalk (approved):

PR #679, opened 2024-05-23
fix(plugin-coverage,plugin-js-packages): always include issues array, even if empty

matejchalk:
- To differentiate audits which may produce issues but didn't find any from audits which never produce issues (useful for display purposes), made sure Code coverage and JS Packages always output `issues` array.
- Fixed (hopefully :crossed_fingers:) NPM publish in release workflow, `nx release` uses a different environment variable for the NPM access token (according to [docs](https://nx.dev/recipes/nx-release/publish-in-ci-cd#configure-the-nodeauthtoken)).

### DIFF
--- a/packages/plugin-coverage/src/lib/runner/lcov/transform.ts
+++ b/packages/plugin-coverage/src/lib/runner/lcov/transform.ts
@@ -112,6 +112,8 @@ export function lcovCoverageToAuditOutput(
     score: toNumberPrecision(coverage, MAX_DECIMAL_PLACES),
     value: roundedIntValue,
     displayValue: `${roundedIntValue} %`,
-    ...(stat.issues.length > 0 && { details: { issues: stat.issues } }),
+    details: {
+      issues: stat.issues,
+    },
   };
 }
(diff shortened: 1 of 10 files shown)


vmasek (approved):

packages/plugin-coverage/src/lib/runner/lcov/transform.ts
Tlacenka, line 1: > To differentiate audits which may produce issues but didn't find any from audits which never produce issues (useful for display purposes), made sure Code coverage and JS Packages always output issues array.

This is a new piece of information to keep in mind when developing a new plugin. Therefore, it should be documented somewhere. I suggest updating [audit details section](https://github.com/code-pushup/cli/blob/main/packages/cli/docs/custom-plugins.md?plain=1#L485) in `custom-plugins.md`.

matejchalk: Updated the example :heavy_check_mark:

vmasek (approved):

PR #686, opened 2024-05-29
refactor(plugin-js-packages): make optional deps opt-in, adjust weights

Tlacenka:
In this PR, I make the following changes:
- dependency groups are configurable
- weights are adjusted to make optional dependencies have less influence on the score

As a side-effect, optional audits will no longer appear on portal for our current CLI config (can be changed).

### DIFF
--- a/packages/plugin-js-packages/src/lib/runner/index.ts
+++ b/packages/plugin-js-packages/src/lib/runner/index.ts
@@ -37,15 +37,19 @@ export async function createRunnerConfig(
 }
 
 export async function executeRunner(): Promise<void> {
-  const { packageManager, checks, auditLevelMapping } =
-    await readJsonFile<FinalJSPackagesPluginConfig>(PLUGIN_CONFIG_PATH);
+  const {
+    packageManager,
+    checks,
+    auditLevelMapping,
+    dependencyGroups: depGroups,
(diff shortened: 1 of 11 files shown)


packages/plugin-js-packages/src/lib/runner/index.ts
```diff
@@ -104,9 +116,10 @@ async function processAudit(
 
   const uniqueResults = pm.audit.postProcessResult?.(fulfilled) ?? fulfilled;
 
-  return supportedDepGroups.map(depGroup =>
+  return compatibleAuditDepGroups.map(depGroup =>
     auditResultToAuditOutput(
-      uniqueResults[depGroup],
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
Tlacenka, line 121: Here only the results of dependency groups that were configured will appear. But TS is not smart enough to know which those are. Not sure if there is a way around this.

matejchalk (approved):
LGTM! :+1:

packages/plugin-js-packages/src/lib/runner/index.ts
Tlacenka: I'll take that as a no 😅 okay.

matejchalk: Sorry, yeah, can't think of a better way :sweat_smile:

PR #687, opened 2024-05-29
fix(plugin-eslint): revert to explicit config file in nx helper

matejchalk:
Partially reverts #683. Omitting the config file produces [warnings](https://github.com/code-pushup/cli/actions/runs/9284755601/job/25547803390?pr=686#step:5:14438). I'm keeping the config file optional, but reverting back to explicit config in Nx helpers since it appears to be introducing more problems if otherwise :confused:


Tlacenka (approved):
😢 🙏

PR #690, opened 2024-05-30
fix(plugin-js-packages): calculate total dependencies from package.json

### LINKED ISSUE #597: Packages plugin issues

Tlacenka:
Closes #597 

In this PR I fix the way of calculating outdated score. Now the number of all dependencies of a given type is used to create the ratio of outdated major dependencies.

At the moment, only one `package.json` can be referenced to calculate the total.


packages/plugin-js-packages/src/lib/runner/index.ts
```diff
@@ -69,16 +72,29 @@ async function processOutdated(
     ignoreExitCode: true, // outdated returns exit code 1 when outdated dependencies are found
   });
 
+  const totalDeps = objectFromEntries(
+    await Promise.all(
+      depGroups.map(async depGroup => [
+        depGroup,
+        await getTotalDependencies(depGroup, packageJsonPath),
+      ]),
+    ),
```
matejchalk, line 81: :snail: This would reads the same `package.json` file multiple times. Reading from disk is in general not a cheap operation, so it would be more optimal to read the file once, since it contains all the dependency groups.

packages/plugin-js-packages/src/lib/runner/utils.ts
```diff
@@ -40,3 +44,12 @@ export function filterAuditResult(
...
   };
 }
+
+export async function getTotalDependencies(
+  depGroup: DependencyGroup,
+  packageJsonPath: string,
+) {
+  const packageJson = await readJsonFile<PackageJson>(packageJsonPath);
+  const totalDeps = packageJson[dependencyGroupToLong[depGroup]];
+  return totalDeps == null ? 0 : objectToKeys(totalDeps).length;
```
matejchalk, line 54: The `totalDeps` variable contains an object, not a number, so the name is misleading:

```suggestion
  const deps = packageJson[dependencyGroupToLong[depGroup]];
  return deps == null ? 0 : objectToKeys(totalDeps).length;
```

packages/plugin-js-packages/src/lib/runner/utils.unit.test.ts
```diff
@@ -1,6 +1,42 @@
...
+      MEMFS_VOLUME,
+    );
+  });
+
+  it.each([
+    ['prod', 1],
+    ['dev', 3],
+    ['optional', 0],
+  ] satisfies [DependencyGroup, number][])(
+    'should return correct number of dependencies',
```
matejchalk, line 32: Please use the `it.each` variables to avoid duplicate test names:

```suggestion
    'should return correct number of %p dependencies - %p',
```

Tlacenka: Oh damn, forgot, my bad 😅

packages/plugin-js-packages/src/lib/runner/index.ts
Tlacenka: Good point, improved.

packages/plugin-js-packages/src/lib/runner/utils.unit.test.ts
Tlacenka: Dynamic testing removed as the function scope changed.

matejchalk (approved):

PR #692, opened 2024-06-03
ci: disable nx cloud

matejchalk:
Lost my patience - convinced it does more harm than good at this point. The speed-up is negligable, and nowhere near worth the amount of mindf***s related to overzealous caching.


Tlacenka (approved):
I was praying for this moment ever since we adopted Nx Cloud.
Purge it! 🔥

vmasek (approved):

Issue #394, opened 2024-01-04
package.json configuration for @code-pushup/eslint-plugin is blocking it's use


beaussan:
### What happened?

While I was trying to use the project, I followed the readme.

First of all, I tried with the ts config file name, however it wasn't picked automatically (fixed by https://github.com/code-pushup/cli/issues/307 ), but then I was hit with the following error :

```
code-pushup  --config code-pushup.config.ts 
file:///Users/beaussan/perso/git/beaussan-mono/code-pushup.config.bundled_tqbo9am8nv.mjs:2
import eslintPlugin, { eslintConfigFromNxProjects } from "@code-pushup/eslint-plugin";
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: Named export 'eslintConfigFromNxProjects' not found. The requested module '@code-pushup/eslint-plugin' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@code-pushup/eslint-plugin';
const { eslintConfigFromNxProjects } = pkg;
```

So I circle around the issue by changing the import to the following :
```ts
import defaultEslintPlugin from '@code-pushup/eslint-plugin';
import type { CoreConfig } from '@code-pushup/models';

const eslintPlugin = defaultEslintPlugin.eslintPlugin;

const eslintConfigFromNxProjects =
  defaultEslintPlugin.eslintConfigFromNxProjects;
```
And that gave another error :
```
code-pushup  --config code-pushup.config.ts 
(node:57828) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/Users/nicolasbeaussarthatchuel/perso/git/beaussan-mono/node_modules/@code-pushup/eslint-plugin/index.js:2
import { mkdir as mkdir2, writeFile } from "fs/promises";
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

### What would you expect to happen?

I expect typescript to be happy, node to resolve correctly and to be able to run the eslint plugin properly.

After following my hunch, I can confirm that the types of the packages are wrong ([report here](https://arethetypeswrong.github.io/?p=%40code-pushup%2Feslint-plugin%400.6.5)) :

<img width="781" alt="image" src="https://github.com/code-pushup/cli/assets/7281023/2362c03d-6b66-4eeb-a1c8-e30c23c08878">

[...]

PR #700, opened 2024-06-11
fix: round down report scores to avoid misleading perfect results

hanna-skryl:
## Description

This PR updates the `formatReportScore` function to refine the way scores are rounded and displayed. 

The function now scales the score by 100 and then checks if the rounded score is 100. If the score is close to, but not actually, a perfect score, it uses `Math.floor` to avoid rounding up to 100.

### Context:
In the portal repository (Issue [#394](https://github.com/code-pushup/portal/issues/394)), it was identified that the rounding method used was causing confusion by displaying near-perfect scores as 100. To maintain consistency and accuracy across projects, the CLI's score formatting function has been updated to address the same issue.


matejchalk (approved):

BioPhoton (approved):
Thx for keeping the CLI aligned with the portal!

PR #703, opened 2024-06-12
chore: synchronise prod dependencies with root

Tlacenka:
When implementing a merge of `package.json` dependencies, I found out that some of plugins' production dependencies are incorrectly categories or missing in our root `package.json`. This PR aims to amend that.

## Overview
There are 9 dependencies which are production dependencies somewhere but not mentioned in root.
Out of these 9, 3 are internal package dependencies (`@code-pushup/models`, `@code-pushup/utils`, `@code-pushup/core`)

Rest should be synchronised:

- `nx` (dev dependency in root, prod v `nx-plugin` )
- `tslib` (missing in root, prod in `nx-plugin`)
- `@nx/devkit` (dev dependency in root, prod in `nx-plugin`)
- `eslint` (dev in root, prod in `plugin-eslint`)
- `lighthouse` (dev in root, prod in `plugin-lighthouse`)
- `lighthouse-logger` (missing in root, prod in `plugin-lighthouse`)

🆘 Please advise on how to resolve different versions in plugin vs root `package.json` I took max at the moment but not sure if that is the best strategy.


packages/nx-plugin/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "0.29.0",
   "license": "MIT",
   "dependencies": {
-    "@nx/devkit": "^17.1.3",
+    "@nx/devkit": "^17.3.2",
     "tslib": "2.6.2",
-    "nx": "^17.1.3"
+    "nx": "^17.3.2"
   },
```
matejchalk, line 9: Please use the minimum values here, so that we support older versions as well.

packages/plugin-eslint/package.json
```diff
@@ -5,7 +5,7 @@
   "dependencies": {
     "@code-pushup/utils": "0.45.0",
     "@code-pushup/models": "0.45.0",
-    "eslint": "^8.46.0",
+    "eslint": "~8.56.0",
```
matejchalk, line 8: Again, keep the older version range, so that someone using ESLint <8.56.0 can still use Code PushUp.

matejchalk (changes requested):

packages/nx-plugin/package.json
Tlacenka: Ok, makes sense!

matejchalk (approved):

PR #706, opened 2024-06-13
ci: setup a config per package

BioPhoton:
This PR adds a `code-pushup.config.ts` to a couple of packages.

**Depends on:**
#735 
#737 

**Included changes:**
- [ ] ✏️ edit `code-pushup.config.ts` at root with only plugins `lighthouse`, `js-packages`
- [ ] ➕ add `code-pushup.config.ts` to 📦 `models` with plugins `eslint`, `js-packages`
- [ ] ➕ add `code-pushup.config.ts` to 📦 `utils` with plugins `eslint`, `js-packages`
- [ ] ➕ add `code-pushup.config.ts` to 📦 `core` with plugins `eslint`, `js-packages`
- [ ] ➕ add `code-pushup.config.ts` to 📦 `cli` with plugins `eslint`, `js-packages`
- [ ] ➕ add `code-pushup.config.ts` to 📦 `plugin-eslint` with plgins `eslint`, `js-packages`
- [ ] ➕ add `code-pushup.config.ts` to 📦 `plugin-lighthouse` with plugins `eslint`, `js-packages` 
- [ ] ➕ add `code-pushup.config.ts` to 📦 `plugin-coverage` with plgins `eslint`, `js-packages`


vmasek:
We could merge it like this and I'll simplify it later with `mergeConfig` helper

vmasek (approved):
LGTM, I'd wait for @matejchalk for final approval tho 🤔

packages/cli/code-pushup.config.ts
```diff
@@ -0,0 +1,87 @@
+import 'dotenv/config';
+import { join } from 'path';
+import { z } from 'zod';
+import type { CoreConfig } from '@code-pushup/models';
```
matejchalk, line 4: You realize this doesn't give any type-safety, right? 

image

The package obviously isn't (and shouldn't be) installed, and for the TS path aliases to work the `code-pushup.config.ts` would have to part of some `tsconfig` (would require a special `tsconfig.code-pushup.json`, which is probably overkill). Simple solution is to import type from source directly:

```suggestion
import type { CoreConfig } from '../models/src';
```

packages/cli/code-pushup.config.ts
```diff
@@ -0,0 +1,87 @@
...
+        organization: env.CP_ORGANIZATION,
+        project: env.CP_PROJECT,
+      },
+    }),
+
+  plugins: [
+    await eslintPlugin(await eslintConfigFromNxProject('cli')),
+    await jsPackagesPlugin({
+      packageManager: 'npm',
+      packageJsonPath: join('packages', 'cli', 'package.json'),
```
matejchalk, line 38: Invalid property name:

image


```suggestion
      packageJsonPaths: [join('packages', 'cli', 'package.json')],
```

---

However, I don't think it makes sense to run JS packages plugin per project. It still runs `npm audit`, etc. on the root `package-lock.json`, so the results are the same for each project. E.g. outdated `lighthouse` is reported for `cli` project, even though `cli` doesn't use it.

matejchalk (changes requested):

e2e/cli-e2e/code-pushup.config.ts
```diff
@@ -0,0 +1,55 @@
...
+    env.CP_API_KEY &&
+    env.CP_ORGANIZATION &&
+    env.CP_PROJECT && {
+      upload: {
+        server: env.CP_SERVER,
+        apiKey: env.CP_API_KEY,
+        organization: env.CP_ORGANIZATION,
+        project: env.CP_PROJECT,
+      },
+    }),
```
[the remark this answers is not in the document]
BioPhoton: It will end up in the executor #737

packages/cli/code-pushup.config.ts
BioPhoton: Wouldnt it help if we know per project what we release?

PR #707, opened 2024-06-13
ci: adopt lighthouse URL

BioPhoton:


code-pushup.config.ts
```diff
@@ -78,9 +78,12 @@ const config: CoreConfig = {
       type: 'module',
     }),
 
-    await lighthousePlugin('https://codepushup.dev/', {
-      chromeFlags: DEFAULT_FLAGS.concat(['--headless']),
-    }),
+    await lighthousePlugin(
+      'https://github.com/code-pushup/cli?tab=readme-ov-file#code-pushup-cli/',
```
vmasek, line 82: I'm wondering why to change measurement target from our landing page to GitHub readme?

BioPhoton: The portal should be measured in portal repo as the code lives there. 

Optional we can add a project here and maintain the website in the open repo

vmasek: I'm for setting up website & maybe even portal landing page tracking in portal repo but the LH on CLI produces very random numbers we do not really influence. Tracking github landing page is another case where we would just get numbers on how much the GH infra is under load that moment 😄 .

Is there any way how we can achieve something more stable for demonstration purposes?

BioPhoton: > get numbers on how much the GH infra is under load that moment 😄 .

We have images etc in the readme and a certain amount of other elements there...

BioPhoton: > Is there any way how we can achieve something more stable for demonstration purposes?

We can focus on more stable audits in the portal repo

vmasek (approved):
let's change it to readme, see how stable it gets and if it's worse, get rid of it.

vmasek (approved):

code-pushup.config.ts
matejchalk: IMO as soon we have Lighthouse integrated in some other repo (like the portal), we should remove it from the CLI config. It makes no sense to have it here, this isn't a web app. Doesn't matter if it's measuring the portal login page or the GitHub website, it's still random unrelated data which makes the CLI report less meaningful overall (especially the PR comments). Only reason is to keep it here at all is because at the moment we don't have another place to demo the Lighthouse plugin.

PR #715, opened 2024-06-17
fix: wrap paths in quotes in case of spaces

### LINKED ISSUE #699: Handle spaces in folder names correctly
Code PushUp plugins don't handle folders which contain spaces correctly. On Unix, this is solved by wrapping paths in quotes. Not sure if this works correctly on Windows or not, though.

Used [patch-package](https://github.com/ds300/patch-package) as workaround for now:

----

Patch for `@code-pushup/eslint-plugin@0.44.4`:

```diff
diff --git a/node_modules/@code-pushup/eslint-plugin/bin.js b/node_modules/@code-pushup/eslint-plugin/bin.js
index de62c68..d60c371 100644
--- a/node_modules/@code-pushup/eslint-plugin/bin.js
+++ b/node_modules/@code-pushup/eslint-plugin/bin.js
@@ -1212,7 +1212,7 @@ function executeLint({
       command: "npx",
       args: [
         "eslint",
-        ...configPath ? [`--config=${configPath}`] : [],
+        ...configPath ? [`--config='${configPath}'`] : [],
         ...typeof eslintrc === "object" ? ["--no-eslintrc"] : [],
         "--no-error-on-unmatched-pattern",
         "--format=json",
diff --git a/node_modules/@code-pushup/eslint-plugin/index.js b/node_modules/@code-pushup/eslint-plugin/index.js
index 4d3b19d..697757b 100644
--- a/node_modules/@code-pushup/eslint-plugin/index.js
+++ b/node_modules/@code-pushup/eslint-plugin/index.js
@@ -1357,7 +1357,7 @@ async function createRunnerConfig(scriptPath, audits, targets) {
   await writeFile(PLUGIN_CONFIG_PATH, JSON.stringify(config));
   return {
     command: "node",
-    args: [scriptPath],
+    args: [`'${scriptPath}'`],
     outputFile: RUNNER_OUTPUT_PATH
   };
 }
```

Patch for `@code-pushup/js-packages-plugin@0.44.4`:

```diff
diff --git a/node_modules/@code-pushup/js-packages-plugin/index.js b/node_modules/@code-pushup/js-packages-plugin/index.js
index 2f9fe17..d01effe 100644
--- a/node_modules/@code-pushup/js-packages-plugin/index.js
+++ b/node_modules/@code-pushup/js-packages-plugin/index.js
@@ -1667,7 +1667,7 @@ async function createRunnerConfig(scriptPath, config) {
   await writeFile(PLUGIN_CONFIG_PATH, JSON.stringify(config));
   return {
     command: "node",
-    args: [scriptPath],
+    args: [`'${scriptPath}'`],
     outputFile: RUNNER_OUTPUT_PATH
   };
 }
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

matejchalk:
Fixes #699

### DIFF
--- a/packages/plugin-eslint/src/lib/__snapshots__/eslint-plugin.integration.test.ts.snap
+++ b/packages/plugin-eslint/src/lib/__snapshots__/eslint-plugin.integration.test.ts.snap
@@ -508,7 +508,7 @@ exports[`eslintPlugin > should initialize ESLint plugin for React application 1`
   "packageName": "@code-pushup/eslint-plugin",
   "runner": {
     "args": [
-      "<dirname>/bin.js",
+      ""<dirname>/bin.js"",
     ],
     "command": "node",
     "outputFile": "node_modules/.code-pushup/eslint/runner-output.json",
--- a/packages/utils/src/lib/file-system.unit.test.ts
+++ b/packages/utils/src/lib/file-system.unit.test.ts
@@ -7,6 +7,7 @@ import {
   FileResult,
   crawlFileSystem,
   ensureDirectoryExists,
+  filePathToCliArg,
   findLineNumberInText,
   logMultipleFileResults,
 } from './file-system';
@@ -137,5 +138,13 @@ describe('findLineNumberInText', () => {
 
   it('should return null if pattern not in content', () => {
     expect(findLineNumberInText(``, 'x')).toBeNull();
+  });
+});
+
+describe('filePathToCliArg', () => {
+  it('should wrap path in quotes', () => {
+    expect(filePathToCliArg('My Project/index.js')).toBe(
+      '"My Project/index.js"',
+    );
   });
 });
--- a/packages/plugin-coverage/src/lib/runner/index.ts
+++ b/packages/plugin-coverage/src/lib/runner/index.ts
@@ -6,6 +6,7 @@ import {
   ProcessError,
   ensureDirectoryExists,
   executeProcess,
+  filePathToCliArg,
   readJsonFile,
   ui,
 } from '@code-pushup/utils';
@@ -60,7 +61,7 @@ export async function createRunnerConfig(
 
   return {
     command: 'node',
-    args: [scriptPath],
+    args: [filePathToCliArg(scriptPath)],
     outputFile: RUNNER_OUTPUT_PATH,
     ...(threshold != null && {
       outputTransform: outputs =>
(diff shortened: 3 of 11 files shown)


packages/utils/src/lib/file-system.unit.test.ts
```diff
@@ -139,3 +140,11 @@ describe('findLineNumberInText', () => {
     expect(findLineNumberInText(``, 'x')).toBeNull();
   });
 });
+
+describe('filePathToCliArg', () => {
```
Tlacenka, line 144: ❔ Does this make the path "invalid" in face of `node:path`? If so, do I understand it correctly that this has to be the last operation done on the path (i.e. after path separator unification)?

packages/plugin-eslint/src/lib/__snapshots__/eslint-plugin.integration.test.ts.snap
```diff
@@ -508,7 +508,7 @@ exports[`eslintPlugin > should initialize ESLint plugin for React application 1`
   "packageName": "@code-pushup/eslint-plugin",
   "runner": {
     "args": [
-      "<dirname>/bin.js",
+      ""<dirname>/bin.js"",
```
Tlacenka, line 511: Umm why doesn't this break? Double double-quotes next to each other would be invalid JSON, right? 😅

Tlacenka (approved):
Thanks for fixing this for both Unix and Windows paths! 💯

packages/plugin-eslint/src/lib/__snapshots__/eslint-plugin.integration.test.ts.snap
matejchalk: The format used by `.snap` files isn't actually JSON. E.g. they apparently escape quotes by doubling them instead of with backslash.

packages/utils/src/lib/file-system.unit.test.ts
matejchalk: It's up to whoever created the path to apply proper path separator. This function is only concerned with what needs to be handled when it's used as a command-line argument.

packages/plugin-eslint/src/lib/__snapshots__/eslint-plugin.integration.test.ts.snap
Tlacenka: Ok, make sense!

PR #724, opened 2024-06-24
refactor(utils): switch score columns in comparison

### LINKED ISSUE #696: Switch positions of current and previous values/scores in reports
Currently we display the comparison as  `current` ➡  `previous`,
It seems to be common standard (diff, GitHub, ...) to use `previous` ⬅  `current`.

It especially makes sense in case where we display the calculated score change as it would be better to have it alongside `current` value.

Image

<details><summary>GitHub compare</summary>

Image

</details>

BioPhoton:
closes #696

related:
https://github.com/code-pushup/portal/issues/376

### DIFF
--- a/packages/utils/src/lib/reports/__snapshots__/report-diff-unchanged.md
+++ b/packages/utils/src/lib/reports/__snapshots__/report-diff-unchanged.md
@@ -4,11 +4,11 @@
 
 ## 🏷️ Categories
 
-|🏷️ Category|⭐ Score|
+|🏷️ Category|⭐ Previous score|
 |:--|:--:|
-|Performance|🟢 **92**|
-|Bug prevention|🟡 **68**|
-|Code style|🟡 **54**|
+|Performance||
+|Bug prevention||
+|Code style||
 
 ## 🗃️ Groups
 
--- a/e2e/cli-e2e/tests/__snapshots__/compare.report-diff.md
+++ b/e2e/cli-e2e/tests/__snapshots__/compare.report-diff.md
@@ -4,19 +4,19 @@
 
 ## 🏷️ Categories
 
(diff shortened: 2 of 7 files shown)


vmasek (approved):

packages/utils/src/lib/reports/__snapshots__/report-diff-unchanged.md
```diff
@@ -4,11 +4,11 @@
 
 ## 🏷️ Categories
 
-|🏷️ Category|⭐ Score|
+|🏷️ Category|⭐ Previous score|
 |:--|:--:|
-|Performance|🟢 **92**|
-|Bug prevention|🟡 **68**|
-|Code style|🟡 **54**|
```
matejchalk, line 11: @BioPhoton @vmasek Seriously, this looks OK to you? :sweat_smile: 

**Please don't blindly replace snapshots and blindly approve PRs.** :see_no_evil: 

The whole point of snapshots is to catch unexpected regressions. But it relies on developers and reviewers _looking at them_ and judging that either:
- _"Yes, the output changing this way is expected and intentional."_
- _"No, this is unexpected, the changes must've broken something that previously worked. Thank god we have tests to catch these regressions."_ 

I can't imagine that you actually thought this is what an unchanged Categories table should look like, so my conclusion is that **neither of you really looked**. :disappointed:


image

vmasek: @BioPhoton you broke this 😢 and I approve it 😓
Great example of my latest retro AP, I'll be much more careful with my future approvals

matejchalk: Fixed in #734

PR #685, opened 2024-05-28
refactor(plugin-lighthouse): update lighthouse and adopt tests 

BioPhoton:
Lighthouse v12 received significant changes see [release notes](https://github.com/GoogleChrome/lighthouse/blob/main/changelog.md#1200-2024-04-22).

PR #725, opened 2024-06-25
feat(plugin-lighthouse,plugin-coverage): use audit float values

matejchalk:
Related to https://github.com/code-pushup/portal/issues/390#issuecomment-2186910105


hanna-skryl (approved):
I was looking for a way to test that `numericValue` was not being rounded down to zero and accidentally discovered that the first test case in `lighthouse-plugin.unit.test.ts` was failing.

```ts
expect(groups).toStrictEqual([
  …
]);
```

should be

```ts
expect(groups).toStrictEqual(expect.arrayContaining([
  …
])
```

Seems OOS but I thought you should know. Other than that, looks good 👍

vmasek (approved):

matejchalk:
> I was looking for a way to test that `numericValue` was not being rounded down to zero and accidentally discovered that the first test case in `lighthouse-plugin.unit.test.ts` was failing.

The unit test passes in [CI job](https://github.com/code-pushup/cli/actions/runs/9665667686/job/26663335003?pr=725#step:6:460) and also locally when I run `npx nx unit-test plugin-lighthouse`. Did you re-run `npm install`? The `lighthouse` package was updated recently in #685, one consequence is that there are now 4 groups instead of 5.

hanna-skryl:
@matejchalk After re-running `npm install` tests pass. Sorry for the confusion.

Issue #728, opened 2024-06-28
Coverage plugin doesn't handle absolute coverage directory paths correctly


matejchalk:
The `coverageDirectory` path is currently assumed to be relative to the project root. However, absolute paths are also valid (e.g. created with `path.resolve(__dirname,'../../coverage/...')`), but the plugin doesn't handle them correctly.

The `return join(pathToRoot, coverageDirectory, 'lcov.info')` expression creates an invalid path when `coverageDirectory` is absolute.

```diff
1215c1228
< import { join as join4 } from "node:path";
---
> import { isAbsolute, join as join4 } from "node:path";
1301a1315,1317
>   }
>   if (isAbsolute(coverageDirectory)) {
>     return join4(coverageDirectory, "lcov.info");
```

Issue #727, opened 2024-06-28
Coverage plugin doesn't allow using Jest presets for reporters


matejchalk:
A validation check is in place which expects there to be a `coverageReporters` array including `'lcov'` in the jest config export. However, this doesn't take into account setups which set reporters in a shared file extending with `preset: '../../jest.preset.js'` (common practice in Nx), to reduce duplication. Since the preset file values aren't directly exported by a given project's config in this case, validation fails with `Jest coverage configuration at ... does not include LCOV report format for target .... Add 'lcov' format under coverageReporters.`.

```diff
1298c1311
<   if (!testConfig.coverageReporters?.includes("lcov")) {
---
>   if (!testConfig.coverageReporters?.includes("lcov") && !("preset" in testConfig)) {
```

Issue #726, opened 2024-06-28
Coverage plugin can't import `jest.config.ts`


matejchalk:
The `cjs` format is assumed, which doesn't work with `jest.config.ts` files.

```diff
1288,1290c1302,1303
<   const testConfig = await importEsmModule({
<     filepath: pathToConfig,
<     format: "cjs"
---
>   const { mod } = await bundleRequire2({
>     filepath: pathToConfig
1291a1305
>   const testConfig = "default" in mod ? mod.default : mod;
```

Issue #716, opened 2024-06-20
Fix expected Jest config location in coverage plugin's Nx helper


matejchalk:
The coverage plugin's `getNxCoveragePaths` helper expects to find jest config path in `config`, but the `@nx/jest` executor actually uses `jestConfig`.

---

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@code-pushup/coverage-plugin@0.46.0` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@code-pushup/coverage-plugin/index.js b/node_modules/@code-pushup/coverage-plugin/index.js
index 7c80e86..0af80e1 100644
--- a/node_modules/@code-pushup/coverage-plugin/index.js
+++ b/node_modules/@code-pushup/coverage-plugin/index.js
@@ -1251,12 +1251,12 @@ function hasNxTarget(project, target) {
   return project.data.targets != null && target in project.data.targets;
 }
 async function getCoveragePathsForTarget(targetConfig, pathToRoot, targetContext) {
-  const { config } = targetConfig.options;
+  const { options } = targetConfig;
   if (targetConfig.executor?.includes("@nx/vite")) {
-    return await getCoveragePathForVitest(config, pathToRoot, targetContext);
+    return await getCoveragePathForVitest(options.config, pathToRoot, targetContext);
   }
   if (targetConfig.executor?.includes("@nx/jest")) {
-    return await getCoveragePathForJest(config, pathToRoot, targetContext);
+    return await getCoveragePathForJest(options.jestConfig, pathToRoot, targetContext);
   }
   throw new Error(
     `Unsupported executor ${targetConfig.executor}. @nx/vite and @nx/jest are currently supported.`
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

PR #729, opened 2024-06-28
fix(plugin-coverage): handle different Jest/Vitest config scenarios in Nx helper

matejchalk:
Fixes:
- #716
- #728 
- #727 
- #726 

All of these fixes are needed for integrating coverage plugin in portal.

### DIFF
--- a/packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
+++ b/packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
@@ -51,6 +51,11 @@ vi.mock('bundle-require', () => ({
       coverageDirectory: 'coverage',
     };
 
+    const JEST_PRESET: JestCoverageConfig & { preset?: string } = {
+      preset: '../../jest.preset.ts',
+      coverageDirectory: 'coverage',
+    };
+
     const wrapReturnValue = (
       value: VitestCoverageConfig | JestCoverageConfig,
     ) => ({ mod: { default: value } });
@@ -69,6 +74,8 @@ vi.mock('bundle-require', () => ({
         return wrapReturnValue(JEST_NO_LCOV);
       case 'jest-no-dir':
         return wrapReturnValue(JEST_NO_DIR);
+      case 'jest-preset':
+        return wrapReturnValue(JEST_PRESET);
       default:
         return wrapReturnValue({});
     }
@@ -325,6 +332,29 @@ describe('getCoveragePathForJest', () => {
         'integration-test',
       ),
     ).rejects.toThrow(/configuration .* does not include LCOV report format/);
+  });
+
+  it('should not throw if lcov reporter in both project.json and jest config', async () => {
+    await expect(
+      getCoveragePathForJest(
+        {
+          jestConfig: 'jest-valid.config.integration.ts',
+          coverageReporters: ['lcov'],
+        },
+        { name: 'core', root: join('packages', 'core') },
+        'integration-test',
+      ),
+    ).resolves.toBeTypeOf('string');
+  });
+
+  it('should not throw regarding missing lcov reporter if jest config uses preset', async () => {
+    await expect(
+      getCoveragePathForJest(
+        { jestConfig: 'jest-preset.config.ts' },
+        { name: 'core', root: join('packages', 'core') },
+        'test',
+      ),
+    ).resolves.toBeTypeOf('string');
   });
 
   it('should handle absolute path in coverageDirectory', async () => {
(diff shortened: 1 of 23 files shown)


vmasek (approved):

packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
```diff
@@ -118,11 +141,16 @@ describe('getCoveragePathForTarget', () => {
...
       getCoveragePathsForTarget(
         {
-          executor: '@nx/cypress',
-          options: { config: 'cypress.config.ts' },
+          name: 'ui',
+          root: join('apps', 'ui'),
+          targets: {
+            'component-test': {
+              executor: '@nx/cypress',
+              options: { config: 'cypress.config.ts' },
```
Tlacenka, line 149: Nitpick: Cypress uses `cypressConfig` according to [nx documentation](https://nx.dev/nx-api/cypress/executors/cypress).
```suggestion
              options: { cypressConfig: 'cypress.config.ts' },
```

packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
```diff
@@ -193,32 +259,90 @@ describe('getCoveragePathForJest', () => {
...
+        { jestConfig: 'jest-no-dir.config.integration.ts' },
+        { name: 'core', root: join('packages', 'core') },
+        'integration-test',
       ),
     ).rejects.toThrow(
       /configuration .* does not include coverage path .* Add the path under coverageDirectory/,
     );
   });
 
+  it('should override jest config coverageDirectory from project.json', async () => {
```
Tlacenka, line 281: Good one! 👍

packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
```diff
@@ -193,32 +259,90 @@ describe('getCoveragePathForJest', () => {
...
-        'packages',
-        'integration-test in core',
+        { jestConfig: 'jest-no-lcov.config.integration.ts' },
+        { name: 'core', root: join('packages', 'core') },
+        'integration-test',
       ),
     ).rejects.toThrow(/configuration .* does not include LCOV report format/);
   });
+
+  it('should not throw if lcov reporter in project.json instead of jest config', async () => {
```
Tlacenka, line 311: 📖 Functionality documentation: It would be valuable to include here which value is prioritised in the case of it being provided both in `project.json` and the config.

packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
```diff
@@ -49,6 +51,11 @@ vi.mock('bundle-require', () => ({
       coverageDirectory: 'coverage',
     };
 
+    const JEST_PRESET: JestCoverageConfig & { preset?: string } = {
+      preset: '../../jest.preset.ts',
```
Tlacenka, line 55: ❓ What if the LCOV format is not set in preset configuration either?

packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
```diff
@@ -193,32 +259,90 @@ describe('getCoveragePathForJest', () => {
...
+    await expect(
+      getCoveragePathForJest(
+        { jestConfig: 'jest-preset.config.ts' },
+        { name: 'core', root: join('packages', 'core') },
+        'test',
+      ),
+    ).resolves.toBeTypeOf('string');
+  });
+
+  it('should handle absolute path in coverageDirectory', async () => {
```
Tlacenka, line 334: 💯

Tlacenka (commented):
Thank you for testing and fixing the flow on a project using CommonJS and Jest. 🙏

vmasek (approved):

packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
matejchalk: Added a test case, although it seems a bit pointless to me. If lcov is set in both `jest.config.ts` and `project.json`, then obviously it doesn't throw either. How could the answer be anything else?

packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
matejchalk: Then the user will get an error regarding non-existant `lcov.info` file. Recursively resolving Jest presets is not trivial and would be overkill for this simple sanity check.

packages/plugin-coverage/src/lib/nx/coverage-paths.unit.test.ts
matejchalk: Thought of a more interesting test case just now :thinking: - `jest.config.ts` sets lcov, but `project.json` overrides it with something else. Added a test for it.

Tlacenka (approved):

Issue #603, opened 2024-04-07
add CLI flag `--skipPlugins`


BioPhoton:
### User story

As a developer sometimes I want to skip the execution of certain plugin. E.g. to save time.

ATM there is no way to do this but comment them out in the `code-pushup.config.ts`

### Acceptance criteria

- [ ] the CLI supports a terminal argument `skipPlugins` that is. used as string array that is `undefined` by default
- [ ] if `skipPlugins` is given exclude the provided plugins form the core config
- [ ] notify the user similar to `onlyPlugins`

NachoVazquez:
@BioPhoton, I'll work on this if everyone is good with it!

PR #730, opened 2024-06-29
feat: implement skip-plugins option

NachoVazquez:
## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

The CLI does not have an option to skip plugins from the commands

Closes: #603 

## What is the new behavior?

The CLI has an option `skipPlugins` that will ignore the specified plugins from the command. 

## Does this PR introduce a breaking change?

```
[ ] Yes
[x] No
```


packages/cli/src/lib/implementation/only-plugins.middleware.unit.test.ts
```diff
@@ -48,7 +48,7 @@ describe('onlyPluginsMiddleware', () => {
     expect(plugins).toStrictEqual([expect.objectContaining({ slug: 'p1' })]);
   });
 
-  it('should filter categories for slug "p1"', () => {
+  it('should forward plugins and categories for a slug not present in plugins', () => {
```
BioPhoton, line 51: Nice improvement of the test message!

packages/cli/src/lib/implementation/only-plugins.middleware.unit.test.ts
```diff
@@ -69,7 +69,7 @@ describe('onlyPluginsMiddleware', () => {
     expect(plugins).toBe(originalPlugins);
   });
 
-  it('should forward plugins and categories for a slug not present in plugins', () => {
+  it('should filter categories for slug "p1"', () => {
```
BioPhoton, line 72: Nice improvement of the test message!

packages/cli/src/lib/implementation/skip-plugins.utils.ts
```diff
@@ -0,0 +1,46 @@
...
+  }: { skipPlugins?: string[]; verbose?: boolean } = {},
+): void {
+  const skipPluginsSet = new Set(skipPlugins);
+  const missingPlugins = skipPlugins.filter(
+    plugin => !plugins.some(({ slug }) => slug === plugin),
+  );
+
+  if (missingPlugins.length > 0 && verbose) {
+    ui().logger.info(
+      `${chalk.yellow(
```
BioPhoton, line 25: Not a change request just thoughts.

This message is quite similar to the one for the `onlyPlugins` arg. Would it make sense to reuse logic here?

packages/cli/src/lib/yargs-cli.integration.test.ts
```diff
@@ -30,6 +32,14 @@ describe('yargsCli', () => {
     expect(parsedArgv.onlyPlugins).toEqual([]);
   });
 
+  it('should parse an empty array as a default skipPlugins option', async () => {
```
BioPhoton, line 35: As the `skip` and `only` arguments for plugins are applied after another, I think it would help to document this behaviour in a test.

e.g. 'should use first onlyPlugins and then skipPlugins options on a given config object'

BioPhoton (changes requested):
Very nice PR!! The option was long awaited. 🙌
Also nice you took care about tests! 

I left some thoughts and comments, and a request to move a test into another place.

NachoVazquez:
Awesome @BioPhoton! Thank you, I will handle those.

e2e/cli-e2e/tests/collect.e2e.test.ts
```diff
@@ -89,7 +89,7 @@ describe('CLI collect', () => {
         '--no-progress',
         `--config=${configPath}`,
         '--persist.outputDir=tmp/e2e',
-        '--onlyPlugins=coverage',
+        '--skipPlugins=eslint, lighthouse',
```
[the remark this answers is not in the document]
NachoVazquez: Makes sense, thanks!

packages/cli/src/lib/implementation/only-plugins.middleware.unit.test.ts
NachoVazquez: I think they were inverted.

packages/cli/src/lib/implementation/skip-plugins.utils.ts
NachoVazquez: That's a great idea. I was trying to be as less invasive as possible, but this is a good approach, especially if we start adding more similar flags.

packages/cli/src/lib/yargs-cli.integration.test.ts
NachoVazquez: great idea!

NachoVazquez:
We are ready for a new review! Thanks for the great suggestions.

e2e/cli-e2e/tests/collect.e2e.test.ts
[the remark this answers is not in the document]
BioPhoton: Could you please revert the changes in `e2e/cli-e2e/tests/collect.e2e.test.ts` and move the tests into `e2e/cli-e2e/tests/print-config.e2e.test.ts`

e2e/cli-e2e/tests/collect.e2e.test.ts
```diff
@@ -121,7 +121,7 @@ describe('CLI collect', () => {
   it('should run Lighthouse plugin that runs lighthouse CLI and creates report.json', async () => {
     const { code, stderr } = await executeProcess({
       command: 'code-pushup',
-      args: ['collect', '--no-progress', '--onlyPlugins=lighthouse'],
+      args: ['collect', '--no-progress', '--skipPlugins=eslint,coverage'],
```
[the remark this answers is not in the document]
BioPhoton: see https://github.com/code-pushup/cli/pull/730/files/5581ce9722dd88a509cf8f98e42d14cce29be320#r1667766818

BioPhoton (changes requested):
Thx for the updates! The `collect.e2e.test.ts` should get reverted and we are good to go imho.

e2e/cli-e2e/tests/collect.e2e.test.ts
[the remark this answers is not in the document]
NachoVazquez: You are right; I thought I had done this. My bad.

packages/cli/src/lib/implementation/skip-plugins.utils.ts
BioPhoton: I ment creating an custom error.

e.g.:

```ts
export class PluginFilterError extends Error {
  constructor(filterName: 'onlyPlugins' | 'skipPlugins') {
      super(`missingPlugins: ... for ${filterName}`); 
  }
}
```

e2e/cli-e2e/tests/collect.e2e.test.ts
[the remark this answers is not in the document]
NachoVazquez: Uhm, I did revert the changes, didn't I?

packages/cli/src/lib/implementation/skip-plugins.utils.ts
NachoVazquez: Oh, silly me. The logic was so similar I thought you referred to the whole file.

NachoVazquez:
> Thx for the updates! The `collect.e2e.test.ts` should get reverted and we are good to go imho.

Awesome! The collect.e2e.test.ts did get reverted, but I will move the changes to the print.e2e as you suggested.

packages/cli/src/lib/implementation/skip-plugins.utils.ts
NachoVazquez: Do you want me to rollback these changes?

packages/cli/src/lib/implementation/only-plugins.middleware.unit.test.ts
matejchalk: > I think they were inverted.

Indeed, good catch :+1:

matejchalk (approved):
LGTM :heavy_check_mark:

BioPhoton (approved):
Thx for the updates!

packages/cli/src/lib/implementation/skip-plugins.utils.ts
BioPhoton: it ok. lets get it in

NachoVazquez:
Thank you both! This was fun, let's do it again :)

PR #734, opened 2024-07-08
fix(utils): unchanged categories table markdown content

matejchalk:
Fixes bug from https://github.com/code-pushup/cli/pull/724#discussion_r1668249084

Related issue: #696 

--- 

Example of bug: https://github.com/code-pushup/github-action/pull/36#issuecomment-2213549584

image


vmasek (approved):
thanks for fixing this

PR #735, opened 2024-07-08
feat(utils): merge configs function

vmasek:
Introducing a helper function to deep merge config objects.

Implemented:
- [x] Deep merge arrays with respect to the `slug` field
- [x] Merging of empty (partial) configs supported
- [x] Comprehensive tests

Could be used in https://github.com/code-pushup/cli/pull/706 to support extraction & simplify configs

### DIFF
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -58,6 +58,7 @@ export {
 } from './lib/guards';
 export { logMultipleResults } from './lib/log-results';
 export { CliUi, Column, link, ui } from './lib/logging';
+export { mergeConfigs } from './lib/merge-configs';
 export { ProgressBar, getProgressBar } from './lib/progress';
 export {
   CODE_PUSHUP_DOMAIN,
--- a/packages/utils/src/lib/merge-configs.ts
+++ b/packages/utils/src/lib/merge-configs.ts
@@ -1,0 +1,229 @@
+import {
+  CategoryConfig,
+  CoreConfig,
+  Group,
+  GroupRef,
+  PersistConfig,
+  PluginConfig,
+  UploadConfig,
+} from '@code-pushup/models';
+
+export function mergeConfigs(
+  config: CoreConfig,
+  ...configs: Partial<CoreConfig>[]
+) {
+  return configs.reduce(
+    (acc, obj) => ({
+      ...acc,
+      ...mergeCategories(acc.categories, obj.categories),
+      ...mergePlugins(acc.plugins, obj.plugins),
+      ...mergePersist(acc.persist, obj.persist),
+      ...mergeUpload(acc.upload, obj.upload),
+    }),
+    config,
+  );
+}
+
+function mergeCategories(
+  a: CategoryConfig[] | undefined,
+  b: CategoryConfig[] | undefined,
+) {
+  if (!a && !b) {
+    return {};
+  }
+
+  const mergedMap = new Map<string, CategoryConfig>();
+
+  const addToMap = (categories: CategoryConfig[]) => {
+    categories.forEach(newObject => {
+      if (mergedMap.has(newObject.slug)) {
+        const existingObject: CategoryConfig | undefined = mergedMap.get(
+          newObject.slug,
+        );
+
+        mergedMap.set(newObject.slug, {
+          ...existingObject,
+          ...newObject,
+
+          refs: mergeBySlugs(existingObject?.refs, newObject.refs),
+        });
+      } else {
+        mergedMap.set(newObject.slug, newObject);
+      }
+    });
+  };
+
+  if (a) {
+    addToMap(a);
+  }
+  if (b) {
+    addToMap(b);
+  }
+
+  // Convert the map back to an array
+  return { categories: [...mergedMap.values()] };
+}
+
+function mergePlugins(
+  a: PluginConfig[] | undefined,
+  b: PluginConfig[] | undefined,
+) {
+  if (!a && !b) {
+    return {};
+  }
+
+  const mergedMap = new Map<string, PluginConfig>();
+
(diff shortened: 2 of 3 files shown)


packages/utils/src/lib/merge-configs.ts
```diff
@@ -0,0 +1,229 @@
...
+  GroupRef,
+  PersistConfig,
+  PluginConfig,
+  UploadConfig,
+} from '@code-pushup/models';
+
+export function mergeConfigs(
+  config: CoreConfig,
+  ...configs: Partial<CoreConfig>[]
+) {
```
BioPhoton, line 14: We could type the return type.. e.g. 
```suggestion
): Record<'categories', CategoryConfig[]> {
```

packages/utils/src/lib/merge-configs.ts
```diff
@@ -0,0 +1,229 @@
...
+      ...mergeUpload(acc.upload, obj.upload),
+    }),
+    config,
+  );
+}
+
+function mergeCategories(
+  a: CategoryConfig[] | undefined,
+  b: CategoryConfig[] | undefined,
+) {
```
BioPhoton, line 30: see https://github.com/code-pushup/cli/pull/735/files#r1668451622

packages/utils/src/lib/merge-configs.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  // Convert the map back to an array
+  return { categories: [...mergedMap.values()] };
+}
+
+function mergePlugins(
+  a: PluginConfig[] | undefined,
+  b: PluginConfig[] | undefined,
+) {
```
BioPhoton, line 70: see https://github.com/code-pushup/cli/pull/735/files#r1668451622

packages/utils/src/lib/merge-configs.ts
```diff
@@ -0,0 +1,229 @@
...
+  return [...map.values()];
+}
+
+function mergeGroupsBySlugs<T extends Group>(
+  a: T[] | undefined,
+  b: T[] | undefined,
+) {
+  const map = new Map<string, T>();
+
+  const addToMap = (groups: T[]) => {
```
BioPhoton, line 147: If possible move the function out. Types, and execution will be faster and the code is more pure.

packages/utils/src/lib/merge-configs.unit.test.ts
```diff
@@ -0,0 +1,404 @@
...
+  categories: [
+    {
+      slug: 'nx-validators',
+      title: 'Workspace Validation',
+      refs: [],
+    },
+  ],
+};
+
+const MOCK_CONFIG_BASIC_PLUGIN: CoreConfig = {
```
BioPhoton, line 71: We could move these objects into a fixture folder to reduce the loc of the file

packages/utils/src/lib/merge-configs.ts
```diff
@@ -0,0 +1,229 @@
...
+  return [...map.values()];
+}
+
+function mergeUpload(a: UploadConfig | undefined, b: UploadConfig | undefined) {
+  if (!a && !b) {
+    return {};
+  }
+
+  if (a) {
+    return b ? { upload: { ...a, ...b } } : {};
```
BioPhoton, line 225: This line seems untested.

BioPhoton (changes requested):
I mentioned minor typing improvements. Also the export is missing to reuse the script.
Other than this the additional helper is nice and can reduce setup code a lot!

vmasek:
> I mentioned minor typing improvements. Also the export is missing to reuse the script. Other than this the additional helper is nice and can reduce setup code a lot!

The main function is exported, is there a reason to export any of the internals? I wouldn't expose them as public API

packages/utils/src/lib/merge-configs.ts
matejchalk: Agreed, although the accurate type would be:

```suggestion
): Pick<CoreConfig, 'categories'> {
```

packages/utils/src/lib/merge-configs.unit.test.ts
matejchalk: I personally disagree, the individual test assertions rely on knowing exactly what's in these mocks, so colocating them actually makes it easier to understand the tests for me.

packages/utils/src/lib/merge-configs.ts
matejchalk: I don't think this function is easily extracted, the reason it's inlined is precisely because it isn't pure and mutates the `map` variable from outer scope via closure. Would require refactoring to `[a, b].reduce(...)` and use POJO instead of `Map` to make it truly immutable. But I don't think that's necessary, mutations are OK as long as they're restricted to local function implementation and function is pure from the outside.

matejchalk (changes requested):
:lady_beetle: Found a couple of problems with the merge logic.

matejchalk (commented):
Missing one last thing - export the function from `packages/utils/src/index.ts` so its accessible to import from `@code-pushup/utils` package :grin: 

Otherwise everything looks good :+1:

packages/utils/src/index.ts
```diff
@@ -58,6 +58,7 @@ export {
 } from './lib/guards';
 export { logMultipleResults } from './lib/log-results';
 export { CliUi, Column, link, ui } from './lib/logging';
+export { mergeConfigs } from './lib/merge-configs';
```
BioPhoton, line 61: Thanks!

BioPhoton (approved):
Thanks for adding the export!

vmasek:
> Thanks for adding the export!

Initially I didn't realize you mean this one 😅

BioPhoton (approved):

matejchalk (approved):

Issue #619, opened 2024-04-11
nx-plugin - export `createNodes` to have dynamic tasks 


BioPhoton:
### User story

As a user of the `nx-plugin` I want to have dynamic targets to avoid setup and configuration cost.
In Nx v17 they released the crystal API and this should be leveraged to generate the dynamic targets.

### Acceptance criteria

- [x] export `createNodes` to be abel to register the plugins like:
```json
// nx.json
{
  "plugin": ["@code-pushup/nx-plugin"]
}
```
Add the following targets:

- [x] target `code-pushup` if `code-pushup.config.ts` is present in root - equal to execute `npx code-pushup` with nx defaults.  #773 
  - [x] consider `projectPrefix`
- [x] (nice to have) target `code-pushup--configuration` if `code-pushup.config.ts` is NOT present in root - equal to execute `npx g @code-pushup/nx-plugin:configuration` with nx defaults

### Implementation details

- init target - https://github.com/code-pushup/cli/tree/add-createNodes-init-target
- autorun taget - https://github.com/code-pushup/cli/tree/add-createNodes-api-to-nx-plugin

BioPhoton:
- init target - https://github.com/code-pushup/cli/tree/add-createNodes-init-target
- autorun taget - https://github.com/code-pushup/cli/tree/add-createNodes-api-to-nx-plugin

PR #754, opened 2024-07-18
feat(nx-plugin): add crystal utils

BioPhoton:
Preparation for #619 

This PR includes:
- basic `createNodes` function
   - add configuration target to projects without a code-pushup config file
- relevant testing helpers for the `createNodes` function
- types, constants and utils for future logic
- test helper refactoring and documentation over tests

**Command example:**
- `nx run-many --target=code-pushup--configuration --exclude=test-utils`


testing/test-utils/src/lib/utils/nx-plugin.ts
```diff
@@ -0,0 +1,47 @@
...
+  createNodes: CreateNodes,
+  context: CreateNodesContext,
+  createNodeOptions: T,
+  mockData: {
+    matchingFilesData: Record<string, string>;
+  },
+): Promise<CreateNodesResult> {
+  const { matchingFilesData } = mockData;
+  vol.reset();
+  vol.fromJSON(matchingFilesData, MEMFS_VOLUME);
```
BioPhoton, line 38: I was not sure if this place is the best as it is not 100% clear to the user of this function that memfs is used. Also a cleanup in `beforeEach` for `vol.reset()` is a thing i avoided as it splits the code more.. 

Any feedback welcome

testing/test-utils/src/lib/utils/os-agnostic-paths.unit.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from 'vitest';
+import { osAgnosticPath } from './os-agnostic-paths';
+
+describe('osAgnosticPath', () => {
```
BioPhoton, line 12: I documented the helper with tests

packages/nx-plugin/src/plugin/plugin.ts
```diff
@@ -0,0 +1,31 @@
+import type { CreateNodes, CreateNodesContext } from '@nx/devkit';
+import type { CreateNodesResult } from 'nx/src/utils/nx-plugin';
+import { PROJECT_JSON_FILE_NAME } from '../internal/constants';
+import { createTargets } from './target/targets';
+import { CreateNodesOptions } from './types';
+import { normalizedCreateNodesContext } from './utils';
+
+// name has to be "createNodes" to get picked up by nx
```
BioPhoton, line 8:
```suggestion
// name has to be "createNodes" to get picked up by Nx
```

testing/test-utils/src/lib/utils/os-agnostic-paths.ts
```diff
@@ -4,8 +4,94 @@ import type {
   AuditReport,
 } from '@code-pushup/models';
 
-export function osAgnosticPath(path: string): string {
-  return path.replace(process.cwd(), '<CWD>').replace(/\\/g, '/');
```
BioPhoton, line 8: I refactored the logic and added documentation to clarify the behaviour

testing/test-utils/src/lib/utils/nx-plugin.ts
```diff
@@ -0,0 +1,47 @@
...
+ *  // target should be the init target
+ *  expect(targets[`${CP_TARGET_NAME}--init`]).toBeDefined();
+ *  ```
+ *
+ * @param createNodes
+ * @param context
+ * @param createNodeOptions
+ * @param mockData
+ */
+export async function createFilesAndInvokeCreateNodesOnThem<
```
BioPhoton, line 26: This helper is inspired by https://github.com/nrwl/nx/blob/94101643139680f5a9c1072879ca292686db8e3f/packages/js/src/plugins/typescript/plugin.spec.ts#L2178

I did not go with pulling in minimatch as if the implementation changes our tests still pass :D 
WDYT? @Tlacenka

packages/nx-plugin/src/plugin/target/targets.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { NormalizedCreateNodesContext } from '../types';
+import { createConfigurationTarget } from './configuration-target';
+
+export async function createTargets(
+  normalizedContext: NormalizedCreateNodesContext,
+) {
+  const { targetName = CP_TARGET_NAME } = normalizedContext.createOptions;
+  const rootFiles = await readdir(normalizedContext.projectRoot);
+  return rootFiles.some(filename =>
+    filename.match(/code-pushup\.config.(\w*\.)*(ts|js|mjs)$/),
```
matejchalk, line 12: To ensure file name starts with `code-pushup...`.

```suggestion
    filename.match(/^code-pushup\.config.(\w*\.)*(ts|js|mjs)$/),
```

packages/nx-plugin/src/plugin/types.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type CreateNodesOptions = DynamicTargetOptions;
+
+export type NormalizedCreateNodesContext = CreateNodesContext & {
+  projectJson: ProjectConfiguration;
+} & {
+  projectRoot: string;
+} & {
+  createOptions: CreateNodesOptions;
+};
```
matejchalk, line 12:
```suggestion
export type NormalizedCreateNodesContext = CreateNodesContext & {
  projectJson: ProjectConfiguration;
  projectRoot: string;
  createOptions: CreateNodesOptions;
};
```

packages/nx-plugin/src/plugin/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+): Promise<NormalizedCreateNodesContext> {
+  const projectRoot = dirname(projectConfigurationFile);
+
+  const projectJson: ProjectConfiguration = JSON.parse(
+    (await readFile(projectConfigurationFile)).toString(),
+  ) as ProjectConfiguration;
+  return {
+    ...context,
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    projectJson: projectJson ?? {},
```
matejchalk, line 19: If you think it can be `undefined`, then say so in the type:

```suggestion
  const projectJson: ProjectConfiguration = JSON.parse(
    (await readFile(projectConfigurationFile)).toString(),
  ) as ProjectConfiguration | undefined;
  return {
    ...context,
    projectJson: projectJson ?? {},
```

testing/test-utils/src/lib/utils/os-agnostic-paths.ts
```diff
@@ -4,8 +4,97 @@ import type {
...
+export const AGNOSTIC_PATH_SEP_REGEX = /[/\\]/g;
+export const OS_AGNOSTIC_PATH_SEP = '/';
+export const OS_AGNOSTIC_CWD = `<CWD>`;
```
matejchalk, line 9: These constants seem like they should be internal.

```suggestion
const AGNOSTIC_PATH_SEP_REGEX = /[/\\]/g;
const OS_AGNOSTIC_PATH_SEP = '/';
const OS_AGNOSTIC_CWD = `<CWD>`;
```

testing/test-utils/src/lib/utils/os-agnostic-paths.unit.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  const cwdSpy: MockInstance<[], string> = vi.spyOn(process, 'cwd');
+
+  it('should forward nullish paths on Linux/macOS and Windows', () => {
+    expect(osAgnosticPath(undefined)).toBeUndefined();
+  });
+
+  describe('Unix-based systems (Linux/macOS)', () => {
+    const linuxmMacOSCWD = '/Users/jerry';
```
matejchalk, line 20: This variable name is very hard to read, I would suggest renaming it.

```suggestion
    const unixCwd = '/Users/jerry';
```

testing/test-utils/src/lib/utils/os-agnostic-paths.unit.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      expect(osAgnosticPath('./.code-pushup/.code-pushup.config.ts')).toBe(
+        './.code-pushup/.code-pushup.config.ts',
+      );
+    });
+
+    it('should handle relative paths correctly on Windows', () => {
+      expect(osAgnosticPath('../../.code-pushup/.code-pushup.config.ts')).toBe(
+        '../../.code-pushup/.code-pushup.config.ts',
+      );
+    });
```
matejchalk, line 114: Shouldn't you be using backslashes here in the input?

BioPhoton: fixed! nice catche!

packages/nx-plugin/src/plugin/plugin.unit.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      })}`,
+    };
+    const projects = await createFilesAndInvokeCreateNodesOnThem(
+      createNodes,
+      context,
+      {},
+      { matchingFilesData },
+    );
+
+    // project should have one target created
```
Tlacenka, line 59: I recommend having one direct assertion regarding the content of the targets instead of 3. This would also get rid of the nullish operator causing branching logic.

Applicable in multiple places.

testing/test-utils/src/lib/utils/nx-plugin.ts
Tlacenka: I would not call vol.reset before (what if the caller does preprocessing) or after (that would remove all the created files in this function). I would have the caller take care of it and make sure all is set up for the function, e.g. mocking setup.

I would also call this function something to indicate it works on virtual system not a regular fs.

testing/test-utils/src/lib/utils/nx-plugin.ts
Tlacenka: I am not an expert on nx plugins so cannot offer advice here.

BioPhoton: Ok. Thanks anyway for looking.

packages/nx-plugin/src/plugin/plugin.unit.test.ts
BioPhoton: i don't get it

testing/test-utils/src/lib/utils/nx-plugin.ts
BioPhoton: removed the reset call.

testing/test-utils/src/lib/utils/nx-plugin.ts
```diff
@@ -0,0 +1,57 @@
...
+ *  // target should be the init target
+ *  expect(targets[`${CP_TARGET_NAME}--init`]).toBeDefined();
+ *  ```
+ *
+ * @param createNodes
+ * @param context
+ * @param createNodeOptions
+ * @param mockData
+ */
+export async function createVirtualFilesAndInvokeCreateNodesOnThem<
```
matejchalk, line 26: Although I appreciate the descriptiveness, this function name is a real mouthful :sweat_smile: Since there's also an extensive description in JSDocs, I think it's fine to shorten the name a little, e.g.:

```suggestion
export async function invokeCreateNodesOnVirtualFiles<
```

BioPhoton: Happily change it back!

packages/nx-plugin/src/plugin/plugin.unit.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const targets = projects[projectRoot]?.targets ?? {};
```
matejchalk, line 38: There's a type error because the test util incorrectly claims to resolve to `CreateNodesResult`:

image

packages/nx-plugin/src/plugin/plugin.unit.test.ts
matejchalk: I think she means something like this?

image

packages/nx-plugin/src/plugin/plugin.unit.test.ts
BioPhoton: I adjusted the types. Have a look.

matejchalk: No, that's still not accurate. The code un-nests `projects`, but doesn't match the `CreatesNodeResult` type. 

If you want to resolve to `CreatesNodeResult`, then the merged projects should be under `projects`:

image

In this case, usage would be `const { projects } = await invokeCreateNodesOnVirtualFiles(...)`.

Also, not sure if you want to ignore the `externalNodes` this way instead of merging them from the results (same as for `projects`).

BioPhoton: Sry for the comment before. I deleted it and wanted to come up with a better solution. 
Thanks for the information I get back to it later.

packages/nx-plugin/src/plugin/plugin.unit.test.ts
Tlacenka: @matejchalk Yes, exactly! Thank you, I was on my tablet when I gave you the review, not the best place to write code 😅

BioPhoton: Yesterday night it popped up again. And I remembered previose similar feedback.. Will do! THX!

BioPhoton: Updated now!  Thx for the hint, WAY better now.

packages/nx-plugin/src/plugin/plugin.unit.test.ts
BioPhoton: fixed!! thx for the second detailed answer! Should have understood it earlier :)

testing/test-utils/src/lib/utils/nx-plugin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Unit Testing helper for the createNodes function of a Nx plugin.
+ * This function will create files over `memfs` from testCfg.matchingFilesData
+ * and invoke the createNodes function on each of the files provided including potential createNodeOptions.
+ * It will aggregate the results of each invocation and return the CreateNodesResult.
+ *
+ * @example
+ * ```ts
+ * const {projects = {}} = await createFilesAndInvokeCreateNodesOnThem(createNodes, context, undefined, { matchingFilesData});
```
matejchalk, line 13:
```suggestion
 * const projects = await createFilesAndInvokeCreateNodesOnThem(createNodes, context, undefined, { matchingFilesData});
```

testing/test-utils/src/lib/utils/nx-plugin.ts
```diff
@@ -0,0 +1,61 @@
+import { CreateNodes, CreateNodesContext, CreateNodesResult } from '@nx/devkit';
+import { vol } from 'memfs';
+import { MEMFS_VOLUME } from '../constants';
+
+/**
+ * Unit Testing helper for the createNodes function of a Nx plugin.
+ * This function will create files over `memfs` from testCfg.matchingFilesData
+ * and invoke the createNodes function on each of the files provided including potential createNodeOptions.
+ * It will aggregate the results of each invocation and return the CreateNodesResult.
```
matejchalk, line 9:
```suggestion
 * It will aggregate the results of each invocation and return the projects from CreateNodesResult.
```

matejchalk (approved):

PR #737, opened 2024-07-08
feat(nx-plugin): add executor to nx-plugin

BioPhoton:
**Precondition:**
#740 
closes #289

**This PR includes:**
  - add executor `autorun`
  - add executor utils
  - add format normalization

**Not included in this PR:**
- e2e tests for the autorun executor
- [ ] #761

 
Related to:
#63

PR #756, opened 2024-07-21
test(nx-plugin): fix format normalization

BioPhoton:
**🔥 Hot fix for #737** 

This PR fixes a missing feature for passing formats as comma separated string.

When executing `npx @code-pushup/cli collect --persist.format=md,json` (side effect of forwarding Nx terminal args)

Before it was: `["md,json"]`

Now it is: `["md", "json"]`

**Additional information:**
I'm sure we had an issue + PR for this but I can't find it.

### DIFF
--- a/packages/cli/src/lib/implementation/core-config.middleware.ts
+++ b/packages/cli/src/lib/implementation/core-config.middleware.ts
@@ -4,6 +4,7 @@ import {
   DEFAULT_PERSIST_FILENAME,
   DEFAULT_PERSIST_FORMAT,
   DEFAULT_PERSIST_OUTPUT_DIR,
+  Format,
   uploadConfigSchema,
 } from '@code-pushup/models';
 import { CoreConfigCliOptions } from './core-config.model';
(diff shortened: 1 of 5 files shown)


packages/cli/src/lib/yargs-cli.ts
```diff
@@ -115,7 +115,9 @@ export function yargsCli<T = unknown>(
 function validatePersistFormat(persist: PersistConfig) {
   try {
     if (persist.format != null) {
-      persist.format.forEach(format => formatSchema.parse(format));
+      persist.format
+        .flatMap(format => format.split(','))
```
BioPhoton, line 119: I couldn't get it in over coerce in the command options and the check function in `yargs-cli` is called before.

packages/cli/src/lib/implementation/core-config.middleware.ts
```diff
@@ -57,11 +57,16 @@ export async function coreConfigMiddleware<
         DEFAULT_PERSIST_OUTPUT_DIR,
       filename:
         cliPersist?.filename ?? rcPersist?.filename ?? DEFAULT_PERSIST_FILENAME,
-      format: cliPersist?.format ?? rcPersist?.format ?? DEFAULT_PERSIST_FORMAT,
+      format: normalizeFormats(
```
BioPhoton, line 60: placing this in a coerce fn from yargs was not working. :(

matejchalk (approved):

Issue #646, opened 2024-04-24
Possible improvements to the JS packages plugin


Tlacenka:
Here are features that could be supported by this plugin:
- Peer dependencies
- Deprecated packages

PR #762, opened 2024-07-23
feat: add defaults to js-packages plugin options

### LINKED ISSUE #601: feat(plugin-js-packages): add default packageManager
### User story

As a developer I want to have the minimum setup possible and best smart defaults.
In the plugin options the `packageManager` options is required.

### Acceptance criteria

- [x] in the plugin options the `packageManager` options is optional
- [x] by default 'npm' is used for the `packageManager` options
- [x] in the plugin options the `packageManager` options defaults to a smart fallback (e.g. derived from the filesystem)

### Implementation details

I want to use the plugin like this:

```ts
export default {
// ...
plugins: [
  // no options required
  jsPackages()
]
}
```

BioPhoton:
Blocked by #754 as it includes the `osAgnosticPath` helper

closes #601


packages/plugin-js-packages/README.md
```diff
@@ -59,7 +59,7 @@ It supports the following package managers:
      // ...
      plugins: [
        // ...
-       await jsPackagesPlugin({ packageManager: ['yarn-classic'], checks: ['audit'], dependencyGroups: ['prod'] }),
+       await jsPackagesPlugin({ packageManager: 'yarn-classic', checks: ['audit'], dependencyGroups: ['prod'] }),
```
BioPhoton, line 62: I removed the array as it might confuse. Later we anyway go deeper on configuration.

packages/plugin-js-packages/src/lib/js-packages-plugin.ts
```diff
@@ -32,13 +33,20 @@ import { createRunnerConfig } from './runner';
...
+  );
+  const {
+    packageManager = await derivePackageManager(),
+    dependencyGroups: dependencyGroupsCfg = [],
+    checks: checksCfg = [],
+    ...jsPackagesPluginConfigRest
+  } = jsPackagesPluginConfig;
+  const checks = [...new Set(checksCfg)];
+  const depGroups = [...new Set(dependencyGroupsCfg)];
+  const pm = packageManagers[packageManager];
```
matejchalk, line 49: It would be cleaner to extract this to some `normalizeConfig` function.

packages/plugin-js-packages/src/lib/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  // Check for lock files
+  if (await fileExists(join(currentDir, 'package-lock.json'))) {
+    return 'npm';
+  } else if (await fileExists(join(currentDir, 'pnpm-lock.yaml'))) {
+    return 'pnpm';
+  } else if (await fileExists(join(currentDir, 'yarn.lock'))) {
+    if (await fileExists(join(currentDir, '.yarnrc.yml'))) {
+      return 'yarn-modern';
+    }
+    return 'yarn-classic';
```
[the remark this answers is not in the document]
Tlacenka: There isn't currently a fool-proof way to do so, no. It is optional. So if they don't have `.yarnrc.yml`, they will need to provide the argument explicitly.

📖 This should be documented in the plugin README.

packages/plugin-js-packages/src/lib/constants.ts
```diff
@@ -1,7 +1,13 @@
 import { IssueSeverity } from '@code-pushup/models';
-import type { DependencyGroup, PackageAuditLevel } from './config';
+import type {
+  DependencyGroup,
+  PackageAuditLevel,
+  PackageManagerId,
+} from './config';
 import { DependencyGroupLong } from './runner/outdated/types';
 
+export const DEFAULT_PACKAGE_MANAGER: PackageManagerId = 'npm';
```
Tlacenka, line 9: 👨‍💻 Naming: Not sure default is a good name - it is only used if attempts to find out the package manager fail. `FALLBACK_PACKAGE_MANAGER` might be better.

Tlacenka (commented):
Great improvement, thank you for taking care of the plugin 🙏 

Left a few comments.

packages/plugin-js-packages/src/lib/utils.unit.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('should return yarn-modern if yarn.lock and .yarnrc.yml is present', async () => {
+    fileExistsSpy.mockImplementation(path =>
+      Promise.resolve(
+        path.endsWith('yarn.lock') || path.endsWith('.yarnrc.yml'),
+      ),
+    );
+    await expect(derivePackageManager()).resolves.toBe('yarn-modern');
+  });
+});
```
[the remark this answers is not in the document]
Tlacenka: +1, the tests would simulate the situation in a more realistic way.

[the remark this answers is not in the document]
BioPhoton: Done!

packages/plugin-js-packages/src/lib/utils.ts
```diff
@@ -0,0 +1,21 @@
+import { join } from 'node:path';
+import { fileExists } from '@code-pushup/utils';
+import { PackageManagerId } from './config';
+import { DEFAULT_PACKAGE_MANAGER } from './constants';
+
+export async function derivePackageManager(
+  currentDir = process.cwd(),
+): Promise<PackageManagerId> {
+  // Check for lock files
```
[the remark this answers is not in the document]
BioPhoton: done!

packages/plugin-js-packages/src/lib/constants.ts
BioPhoton: done!

packages/plugin-js-packages/src/lib/utils.unit.test.ts
```diff
@@ -0,0 +1,301 @@
...
+  });
+
+  it('should return yarn-classic if yarn.lock is present', async () => {
+    vol.fromJSON(
+      {
+        'yarn.lock': '',
+      },
+      MEMFS_VOLUME,
+    );
+    (exec as MockInstance<[], unknown>).mockImplementation((_, fn) =>
```
BioPhoton, line 253: Could need some help to type exec mock :)

packages/plugin-js-packages/src/lib/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+        },
+      );
+    });
+
+    if (yarnVersion === '2' || yarnVersion === '3') {
+      return 'yarn-modern';
+    }
+    return 'yarn-classic';
+  }
+  return FALLBACK_PACKAGE_MANAGER;
```
vmasek, line 98: maybe it should end with error message hinting user to configure the package manager manually as it could not be determined

BioPhoton: well if nothing is there then they never installed the packages. or?

packages/plugin-js-packages/src/lib/utils.unit.test.ts
matejchalk: I'd suggest the following approach:

```ts
import cp from 'node:child_process';

// ...

describe('...', () => {
  // 👇 casting narrows mockImplementation types
  const execSpy = vi.spyOn(cp, 'exec') as MockInstance<
    [string, (error: Error | null, stdout: string, stderr?: string) => void]
  >;

  afterAll(() => {
    execSpy.mockRestore();
  });

  it('...', () => {
    // ...
    execSpy.mockImplementation((_, callback) => {
      callback(null, '1.22.19');
    });
    // ...
    expect(execSpy).toHaveBeenCalledWith('yarn -v', expect.any(Function));
  });
});
```

packages/plugin-js-packages/src/lib/utils.ts
vmasek: there could be other things that install deps, they could also just copy them over from cache and never have package manager binary installed.

Also JS ecosystem moves fast, `bun` can install packages as far as I know too 😅

vmasek: what I currently know of:
- npm
- pnpm
- Yarn
- Yarn 2
- Bower
- jspm
- volta
- Entropic
- Bun

BioPhoton: Thx for the more information!

Just started to introduce an error.

packages/plugin-js-packages/src/lib/utils.unit.test.ts
BioPhoton: I can't define the spy at the top of the describe with spyOn. It throws: 
`TypeError: Cannot redefine property: exec`. I don't know where but we had that problem already somewhere with `child_process`.

BioPhoton: I refactored it and pushed so someone could reproduce the case.

packages/plugin-js-packages/src/lib/utils.ts
BioPhoton: done

vmasek:
as merged it should close this issue too #646

packages/plugin-js-packages/src/lib/utils.unit.test.ts
matejchalk: Ah, it's possible this spying doesn't work for built-in `node:` modules.

On seconds thoughts, why not just use our `executeProcess` function from `utils`? It's a lot easier to mock as it's already promisifed.

image

BioPhoton: what I learned:
- All node internal seems to work only with `vi.mock('import', () => { })`
- `vi-spyOn(obj, prop)` only works on the shallow level of a function. Meaning if I spy on A and test C which uses B which intern uses A (C->B->A). Means B is using the original function

e2e/cli-e2e/vite.config.e2e.ts
```diff
@@ -6,7 +6,7 @@ export default defineConfig({
   cacheDir: '../../node_modules/.vite/cli-e2e',
   test: {
     reporters: ['basic'],
-    testTimeout: 140_000,
+    testTimeout: 150_000,
```
BioPhoton, line 9: I bump it by 10 and see if it makes a difference in terns of flakyness

vmasek (dismissed):

packages/plugin-js-packages/src/lib/package-managers/derive-package-manager.ts
```diff
@@ -0,0 +1,57 @@
...
+    return 'pnpm';
+  } else if (await fileExists(join(currentDir, 'yarn.lock'))) {
+    const yarnVersion = await deriveYarnVersion();
+    if (yarnVersion) {
+      return yarnVersion;
+    }
+  }
+
+  throw new Error(
+    'Could not detect package manager. Please provide in in the js-packages plugin config.',
```
matejchalk, line 55:
```suggestion
    'Could not detect package manager. Please provide it in the js-packages plugin config.',
```

packages/plugin-js-packages/src/lib/package-managers/derive-package-manager.unit.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    await expect(derivePackageManager()).resolves.toBe('yarn-classic');
+    expect(fileExistsSpy).toHaveBeenCalledWith(
+      expect.stringContaining('yarn.lock'),
+    );
+    expect(deriveYarnVersionSpy).toHaveBeenCalledTimes(1);
+    expect(deriveYarnVersionSpy).toHaveBeenCalledWith();
+  });
+
+  it('should fall back to npm if neither filesystem nor env shows hints', async () => {
```
matejchalk, line 176: Inaccurate test title:

```suggestion
  it('should report error if neither filesystem nor env shows hints', async () => {
```

packages/plugin-js-packages/src/lib/package-managers/derive-yarn.ts
```diff
@@ -0,0 +1,16 @@
...
+    command: 'yarn',
+    args: ['-v'],
+  });
+
+  const yarnVersion = stdout.toString().trim().at(0);
+  if (yarnVersion === '2' || yarnVersion === '3') {
+    return 'yarn-modern';
+  } else if (yarnVersion === '1') {
+    return 'yarn-classic';
+  }
```
matejchalk, line 14: The latest Yarn is version 4, which this code doesn't consider. For future-proofing, I would simply consider any version larger than 1 as Yarn Modern.

```suggestion
  const yarnVersion = Number.parseInt(stdout.toString().trim().at(0), 10);
  if (yarnVersion >= 2) {
    return 'yarn-modern';
  } else if (yarnVersion === 1) {
    return 'yarn-classic';
  }
```

BioPhoton: Thanks!! Way better. I have no clue why I did not plan future in :D

packages/plugin-js-packages/src/lib/package-managers/derive-yarn.unit.test.ts
```diff
@@ -24,7 +24,7 @@ describe('deriveYarnVersion', () => {
     });
   });
 
-  it('should return yarn-modern  if and yarn v2 is installed', async () => {
+  it('should return yarn-modern if and yarn greater than v1 is installed', async () => {
```
matejchalk, line 27:
```suggestion
  it('should return yarn-modern if yarn greater than v1 is installed', async () => {
```

matejchalk (approved):
Good work :+1:

matejchalk (approved):

vmasek (approved):

PR #755, opened 2024-07-21
refactor(utils): use build-md for markdown formatting

matejchalk:
I've published a library called [`build-md`](https://matejchalk.github.io/build-md/) for dynamically rendering Markdown documents :rocket: 

I refactored our Markdown rendering logic to use this library. The advantage is **our logic now focuses on document semantics**, and is **no longer concerned with Markdown formatting rules** (like where to add separator lines and what to indent). The library's **Markdown output is better formatted** (fewer Markdown lint violations) - HTML is only used when necessary (context-sensitive rendering), there are no extra or missing separator lines, and the table columns are automatically aligned (as if by Prettier).

image

---

The reason I want this to be a standalone library is because it's applicable to other projects as well like `@code-pushup/eslint-config` (migrated in https://github.com/code-pushup/eslint-config/commit/8ac87936b8d004a66d3e88166a5ebd19c69c9240) or `zod2md` (migration planned).

PR #772, opened 2024-07-27
refactor(plugin-js-packages): use build-md to format issue messages

matejchalk:
Follow-up from #755.

### DIFF
--- a/packages/plugin-js-packages/src/lib/runner/audit/transform.unit.test.ts
+++ b/packages/plugin-js-packages/src/lib/runner/audit/transform.unit.test.ts
@@ -221,6 +221,14 @@ describe('vulnerabilitiesToDisplayValue', () => {
 });
 
 describe('vulnerabilitiesToIssues', () => {
+  const vulnerabilityDefaults: Vulnerability = {
+    name: 'verdaccio',
+    severity: 'high',
+    versionRange: '<=5.28.0',
+    fixInformation: false,
+    directDependency: true,
+  };
+
   it('should provide a vulnerability summary', () => {
     expect(
       vulnerabilitiesToIssues(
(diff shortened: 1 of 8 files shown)


BioPhoton (dismissed):
Nice reduction of the code and also better formatted! 👍

We could move the issue parsing into a separate helper. (Just a suggestion, no requirement)

packages/plugin-js-packages/src/lib/runner/audit/transform.unit.test.ts
```diff
@@ -271,16 +280,17 @@ describe('vulnerabilitiesToIssues', () => {
...
           },
-        ] as Vulnerability[],
+        ],
         defaultAuditLevelMapping,
       ),
     ).toEqual<Issue[]>([
       expect.objectContaining({
         message: expect.stringContaining(
-          "`cypress`' dependency `@cypress/request` has",
+          "`cypress`'s dependency `@cypress/request` has",
```
Tlacenka, line 293: 🐞 This was the main added value of `apostrophize`- it could tell when the word ended with an "s" so that it would not add another "s" after the apostrophe. The current result is grammatically incorrect.

Tlacenka (commented):
Much more readable, thank you 🙏 

I found one grammatical issue, see below.

packages/plugin-js-packages/src/lib/runner/audit/transform.unit.test.ts
matejchalk: I know, but I deliberately removed this logic because it was both over-complicated and unnecessary. 

Grammatically, "`cypress`'s" is perfectly acceptable. In fact, both `s's` and `s'` forms are in common use. There is no hard rule about when to use which, just [stylistic preferences](https://www.grammarly.com/blog/possessive-case/#:~:text=This%20general%20rule,be%20used%20consistently). The only differentiation (loosely upheld) I've come across is [`s's` for singular nouns](https://www.scribbr.com/language-rules/apostrophe-s/#:~:text=Names%20ending%20in,is%20always%20wrong.) and `s'` for plural nouns. And NPM package name's are predominantly singular.

Tlacenka: I see. Not a fan of this change but thank you for the explanation, seems legit.

matejchalk: Removed it in 9d0947518c5d848dd003bba411cceac78308e8d7

vmasek (approved):

Issue #643, opened 2024-04-24
test: consolidate e2e tests setup


getlarge:
@BioPhoton @matejchalk It looks like there is a problem with the e2e tests setup:
- `cli-e2e` and `nx-plugin-e2e` declares the same `globalSetup` which is the relative path to `global-setup.e2e.ts`.
- in `global-setup.e2e.ts` the ephemeral package registry is created
- This means that when running the e2e tests in parallel, there are two registries on two different ports since Verdaccio increments the port number when one is busy.
- but the folder is the same for both...

Which leads to race conditions and conflicts with package publishing/installation.

_Originally posted by @getlarge in https://github.com/code-pushup/cli/issues/639#issuecomment-2072213132_

getlarge:
We started to dig on this problem with @BioPhoton and found out that:
- each local registry instantiated should have a unique folder so we appended a random number to the folder name (e.g. `/tmp/local-registry/storage-${randomInt}`
- the registry is not always correctly teardown in case of failure (processes remain open)

There are still other factors to consider that might be related to environment variables and/or NPM configuration conflicts, 
since the e2e tests still fail when run in parallel when using a random Verdaccio storage location.

I looked for how Nx handles the local registry for their end-to-end tests and they use similar logic in a global setup hook to create unique storage path for [each e2e project by joining](https://github.com/nrwl/nx/blob/e8ca02a6121eadbf498711f66334e07b563221b8/e2e/utils/global-setup.ts#L16-L25) `process.env.NX_TASK_TARGET_PROJECT` to the path.

Their setup differs in the way they run the e2e tests in the CI, instead of running tests in parallel in the same parent process, they [run each test in a different runner using a matrix](https://github.com/nrwl/nx/blob/master/.github/workflows/e2e-matrix.yml#L387-L390). This way they isolate environment variables, NPM and Git configuration concurrent access.

getlarge:
On the same topic, the generic `global-setup.e2e.ts` and the `cli-e2e` test suite are currently coupled.
The functions `setupTestFolder` and `teardownTestFolder` are called with the argument `tmp/e2e`, which seems specific to  the `cli-e2e` moreover, if multiple projects would use the same `global-setup.e2e.ts` the folder could be deleted while it is still being used by another test suite which is pending.
Consider moving these functions in `beforeAll` and `afterAll` hooks?

[...]

PR #773, opened 2024-07-31
feat(nx-plugin): implement `bin` property in options

BioPhoton:
**Precondition to reduce the scope:**
- #777
- #778
- #785

This PR implements a `bin` option in for different places in the `nx-plugin`. It also. includes better e2e tests.
- implementation of a `bin` property for the following areas:
  - `nx-plugin:configuration` generator
  - `nx-plugin:autorun` executor
  - `nx-plugin` Derived Targets
- Lot of tests
  - add e2e tests for nx-plugin
  
> [!note]
> The new way of dynamically generate test setups is the highlight of this PR. It already discovered that our generators did not execute but fail. 🔥


testing/test-nx-utils/src/lib/utils/tree.ts
```diff
@@ -0,0 +1,30 @@
...
+    return;
+  } catch (error) {
+    console.error((error as { code: string; message: string }).message);
+    if ((error as { code: string }).code !== 'EEXIST') {
+      throw error;
+    }
+  }
+}
+
+export async function materializeTree(tree: Tree, targetFolder: string) {
```
BioPhoton, line 17: This is a first a new helper to unlock a new way of generating a e2e setup. See it un use here:

e2e/nx-plugin-e2e/tests/init.e2e.test.ts
```diff
@@ -1,25 +1,80 @@
...
+
+  afterEach(async () => {
+    await rm(baseDir, { recursive: true, force: true });
+  });
+
+  it('should inform about dry run', async () => {
+    const cwd = join(baseDir, 'configure');
+    await materializeTree(tree, cwd);
+
+    const { stderr } = await executeInitGenerator([project, '--dryRun'], cwd);
```
BioPhoton, line 37: Here an example of the new test helper un use. We can reuse Nx helpers to generate and logic from our own generators to create dynamic test scenarios in the file system.

e2e/nx-plugin-e2e/mocks/utils.ts
```diff
@@ -0,0 +1,27 @@
+import {join, relative} from "node:path";
+import {executeProcess} from "@code-pushup/utils";
+
+export function distPluginPackage(cwd: string): string {
+  return relative(join(process.cwd(), cwd), join(process.cwd(), "dist/packages/nx-plugin"));
+}
+export function pluginFilePath(cwd: string): string {
+  return relative(join(process.cwd(), cwd), join(process.cwd(), "packages/nx-plugin/src/index.ts"));
+}
```
BioPhoton, line 9: To e2e test the build output we have to use a relative path to the dist folder. 
If we use `@code-pushup/nx-plugin` as `bin`, verdaccio throws: `// npm ERR! 404 '@code-pushup/nx-plugin@0.29.0' is not in this registry`. 
We can also not use the published package as our [e2e tests need a refactoring](https://github.com/code-pushup/cli/issues/643)

packages/nx-plugin/src/generators/configuration/code-pushup-config.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export type GenerateCodePushupConfigOptions = {
+  fileImports?: ItemOrArray<string>;
+  persist?: Partial<PersistConfig>;
+  upload?: Partial<UploadConfig>;
+  plugins?: ExecutableCode[];
+  categories?: ExecutableCode[];
+};
+
+export function generateCodePushupConfig(
```
BioPhoton, line 19: This is a more flexible generator function that will enable dynamic plugin registration for a test setup. For better understanding check out their tests as we do not use the full functionality ATM.

packages/nx-plugin/src/generators/configuration/files/code-pushup.config.ts.template
```diff
@@ -1,10 +1,9 @@
-import type { CoreConfig } from '@code-pushup/models';
+<%- fileImports %>
```
BioPhoton, line 1: More flexible options for the cp config file are visible best here.

packages/nx-plugin/src/generators/configuration/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+  return `${lines.join(separator).replace('"', '')}`;
+}
+
+export function formatArrayToJSArray(lines?: string[]) {
+  if (lines == null) {
+    return;
+  }
+
+  return `[${formatArrayToLinesOfJsString(lines, ',\n') ?? ''}]`;
+}
```
BioPhoton, line 62: Im not really satisfied. What this functions do is the turn strings and objects ints JS as string. 
This string later on in the template is executable.

I could need hints for:
- better naming
- better technique than replace string patterns
- more intuitive API

e2e/nx-plugin-e2e/project.json
```diff
@@ -19,5 +19,5 @@
     }
   },
   "implicitDependencies": ["nx-plugin"],
-  "tags": ["scope:tooling", "type:e2e"]
+  "tags": ["scope:core", "type:e2e"]
```
BioPhoton, line 22: The current scope configuration is not made for this case :D

packages/nx-plugin/src/generators/configuration/code-pushup-config.ts
BioPhoton: Cc @Tlacenka

e2e/nx-plugin-e2e/project.json
BioPhoton, line 22:
```suggestion
  "tags": ["scope:tooling", "type:e2e"]
```

e2e/nx-plugin-e2e/tests/nx-plugin.e2e.test.ts
```diff
@@ -0,0 +1,256 @@
...
+  materializeTree,
+  nxShowProjectJson,
+  registerPluginInWorkspace,
+} from '@code-pushup/test-nx-utils';
+import { removeColorCodes } from '@code-pushup/test-utils';
+import { executeProcess, readTextFile } from '@code-pushup/utils';
+
+// @TODO replace with default bin after https://github.com/code-pushup/cli/issues/643
+export function relativePathToCwd(testDir: string): string {
+  return relative(join(process.cwd(), testDir), join(process.cwd()));
```
matejchalk, line 18:
```suggestion
  return relative(join(process.cwd(), testDir), process.cwd());
```

e2e/nx-plugin-e2e/tests/nx-plugin.e2e.test.ts
```diff
@@ -0,0 +1,256 @@
...
+    await materializeTree(tree, cwd);
+
+    const { stdout, stderr } = await executeProcess({
+      command: 'npx',
+      args: ['nx', 'run', `${project}:code-pushup -- --dryRun`],
+      cwd,
+    });
+
+    const cleanStderr = removeColorCodes(stderr);
+    // @TODO create test environment for working plugin. This here misses package.jock to execute correctly
```
matejchalk, line 211: I assume you mean one of these? :grin: 

```suggestion
    // @TODO create test environment for working plugin. This here misses package.json to execute correctly
```

```suggestion
    // @TODO create test environment for working plugin. This here misses package-lock.json to execute correctly
```

packages/nx-plugin/src/generators/configuration/__snapshots__/code-pushup-config.integration.test.ts.snap
```diff
@@ -0,0 +1,20 @@
...
+  filename: "report-123"
+},
+    update: {
+  apiKey: "123"
+},
+    plugins: [myPlugin({ timeout: 42})],
+    categories: [myPluginCategory()]
+} satisfies CoreConfig;
+"
+`;
```
matejchalk, line 20: Would be easier to read a formatted `.ts` file as the snapshot (same for unit test snapshot) :pray:

packages/nx-plugin/src/plugin/target/configuration-target.ts
```diff
@@ -1,12 +1,22 @@
...
+}): TargetConfiguration<RunCommandsOptions> {
+  const {
+    projectName,
+    bin = PACKAGE_NAME,
+    targetName = CP_TARGET_NAME,
+  } = options ?? {};
   const projectFlag = projectName && ` --project=${projectName}`;
   return {
-    command: `nx g ${PACKAGE_NAME}:configuration${projectFlag ?? ''}`,
+    command: `nx g ${bin}:configuration --skipTarget --targetName=${targetName}${
```
matejchalk, line 18: Curious what the reason is for adding `--skipTarget`? :thinking:

packages/nx-plugin/src/plugin/target/configuration-target.ts
```diff
@@ -1,12 +1,22 @@
...
+      projectFlag ?? ''
+    }`,
   };
```
matejchalk, line 21: Why not use the `objectToCliArgs` util here? Seems like a good fit because you have some logic for conditional arguments.

Also, if the target name isn't set, then maybe we don't have to explicitly set the default value?

packages/nx-plugin/src/plugin/target/targets.ts
```diff
@@ -2,21 +2,26 @@ import { readdir } from 'node:fs/promises';
...
+  const { targetName = CP_TARGET_NAME, bin } = normalizedContext.createOptions;
   const rootFiles = await readdir(normalizedContext.projectRoot);
   return rootFiles.some(filename =>
     filename.match(/^code-pushup\.config.(\w*\.)*(ts|js|mjs)$/),
   )
     ? // @TODO return code-pushup cli target https://github.com/code-pushup/cli/issues/619
-      {}
+      {
+        [targetName]: createExecutorTarget({ bin }),
+      }
```
matejchalk, line 18: Is the TODO comment now resolved?

packages/nx-plugin/src/plugin/target/targets.unit.test.ts
```diff
@@ -75,6 +79,86 @@ describe('createTargets', () => {
...
+    ).resolves.toStrictEqual(
+      expect.objectContaining({
+        [targetName]: {
+          executor: `${PACKAGE_NAME}:autorun`,
+        },
+      }),
+    );
+  });
+
+  it('should return executor targets for project is configured', async () => {
```
matejchalk, line 117:
```suggestion
  it('should return executor targets for project if configured', async () => {
```

packages/nx-plugin/src/plugin/types.ts
```diff
@@ -3,8 +3,10 @@ import { DynamicTargetOptions } from '../internal/types';
 
 export type CreateNodesOptions = DynamicTargetOptions;
 
+export type ProjectConfigWithName = Omit<ProjectConfiguration, 'name'> &
```
matejchalk, line 6: I think is something we could create a utility type for. Can be implemented [like this](https://www.typescriptlang.org/play/?ts=5.3.3#code/PTAEDMCcHsFtQHIA8BQAXAngBwKagAowBWOAxmgMLQB24AlgOYCukAhmnTaALygDeKUKGqtYOAPwAuUAGc0kOtQYBuQaBjQ00uQqWqAvihQhQAExwKAbjlOhMudNjyFoJclVqMW7TtQDqdGgAFgiieLwBwQBKOACOTHSQNgA8Lm6UNPTMbBw0ADSgAOQiYoUAfEYmiqBMHAA2Mig4SFjQkGh2TqCRQTHxiSkAKgUA0qDNaDjUpjKgANY4GNDgoINlPARJaBzgGMkA8rCBycOgI+sAZKB9CUmmqXSkcyejZW9NLW0d9s5bO3trDZ8UAAbTG1QWSxWgwAutJBmCYaBDEA).

image

testing/test-utils/src/lib/utils/os-agnostic-paths.ts
```diff
@@ -43,6 +43,7 @@ const OS_AGNOSTIC_CWD = `<CWD>`;
  * Output: ../my-folder/my-file.ts
  *
  */
+export function osAgnosticPath(filePath: string): string;
```
matejchalk, line 46: Why add this function declaration?

packages/nx-plugin/src/plugin/target/configuration-target.ts
BioPhoton: To give the user the option to only create the config file. 

Internally it is used for the derived targets of the plugin. If plugin is used no need to create a target.

packages/nx-plugin/src/plugin/target/configuration-target.ts
BioPhoton: Refactored. Good hint!

BioPhoton: for the target name, we need a string as target name to set it in the project. In this case it is a bit useless as we also use `skipTarget`. I just did not want to wrestle with the scheme of the configuration generator.

testing/test-utils/src/lib/utils/os-agnostic-paths.ts
BioPhoton: To get better return type narrowing. (thx btw for looking at it, i forgot the one for undefined :) )

**Before:** 
Screenshot 2024-08-08 at 17 17 57

**After:** 

Screenshot 2024-08-08 at 17 18 06

packages/nx-plugin/src/plugin/types.ts
BioPhoton: 🔥

packages/nx-plugin/src/plugin/target/targets.ts
BioPhoton: consider `projectPrefix` is missing. I added this PR in the issue to show what it resolves

testing/test-utils/src/lib/utils/os-agnostic-paths.ts
matejchalk: I see. Yes, with both overloads it makes sense :+1:

packages/nx-plugin/src/generators/configuration/__snapshots__/code-pushup-config.integration.test.ts.snap
BioPhoton: the integration are updated

matejchalk: I meant `code-pushup-config.integration.test.ts.snap` and `code-pushup-config.unit.test.ts.snap`. But I see you've delete both in caaacc2c973d19a28df0ebab276f8ae7c49c5b00, so I guess they were outdated anyway.

matejchalk (approved):

packages/nx-plugin/src/generators/configuration/__snapshots__/code-pushup-config.integration.test.ts.snap
BioPhoton: yes! tests got refactored and files splitted

PR #777, opened 2024-08-02
feat(test-nx-utils): add testing lib for nx specific logic

BioPhoton:
This PR adds a new testing package the encapsulate nx dependencies.

It includes:
- new package `test-nx-utils` to group Nx related dependencies
  - new helper to generate files


testing/test-nx-utils/src/lib/utils/tree.unit.test.ts
```diff
@@ -0,0 +1,31 @@
+import { join } from 'node:path';
+import { createTreeWithEmptyWorkspace } from 'nx/src/generators/testing-utils/create-tree-with-empty-workspace';
+import { describe, expect } from 'vitest';
+import { readJsonFile } from '@code-pushup/utils';
+import { materializeTree } from './tree';
+
+describe('materializeTree', () => {
+  const baseDir = join('tmp', 'materialize-tree');
```
matejchalk, line 8: Since this is interacting with the actual file system, I would make it an integration test.

Alternatively, you could use `memfs` and keep this as a unit test.

testing/test-nx-utils/src/lib/utils/tree.unit.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  it('should create files from tree', async () => {
```
matejchalk, line 9:
```suggestion

  it('should create files from tree', async () => {
```

testing/test-nx-utils/project.json
```diff
@@ -0,0 +1,33 @@
...
+      }
+    },
+    "unit-test": {
+      "executor": "@nx/vite:test",
+      "options": {
+        "configFile": "testing/test-nx-utils/vite.config.unit.ts"
+      }
+    }
+  },
+  "tags": ["scope:shared", "type:testing"]
```
matejchalk, line 32: If this intended only for the `nx-plugin`, then `scope:tooling` would be more appropriate. The `.eslintrc.json` would have to tweaked, see https://github.com/code-pushup/cli/pull/779#discussion_r1703857058

testing/test-nx-utils/src/lib/utils/tree.unit.test.ts
BioPhoton: Very true. I on purpose did not use memfs. so will rename to integration test

testing/test-nx-utils/project.json
BioPhoton: tweaked!

testing/test-nx-utils/src/lib/utils/tree.unit.test.ts
BioPhoton: done

matejchalk (approved):

testing/test-nx-utils/src/lib/utils/tree.integration.test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { readJsonFile } from '@code-pushup/utils';
+import { materializeTree } from './tree';
+
+describe('materializeTree', () => {
+  const baseDir = join('tmp', 'materialize-tree');
+
+  it('should create files from tree', async () => {
+    const root = join(baseDir, 'materialize');
+
+    const tree = createTreeWithEmptyWorkspace({ layout: 'apps-libs' });
```
Tlacenka, line 13: 🧪 Output cleanup: While the `tmp` folder is ignored by default, each test should clean up after itself (the strategy is on our [Wiki](https://github.com/code-pushup/cli/wiki/Testing-Strategy#test-outputs)). I recommend adding an `after` hook that removes the `materialize` subfolder to ensure clean state.

testing/test-nx-utils/.eslintrc.json
Tlacenka, line 1: 🧪 Utils structure: Wondering if it would not be better to keep this in a subfolder of `test-utils` (the idea was only to separate utilities and mock setup). Then there would not need to be separate config files and targets.

testing/test-nx-utils/src/lib/utils/nx.unit.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+});
+
+describe('registerPluginInWorkspace', () => {
+  it('should register plugin name in workspace', () => {
+    const tree = createTreeWithEmptyWorkspace({ layout: 'apps-libs' });
+
+    registerPluginInWorkspace(tree, 'nx-plugin');
+
+    const nxJson = JSON.parse(tree.read('nx.json')?.toString() ?? '{}');
```
Tlacenka, line 52: 🧪 Linear flow: There should be no need for conditional statements in tests. The test should know what is expected and not insert any default values such as `{}`. I would suggest using `!` here. Curious to hear what @matejchalk thinks.

This is applicable to multiple places.

testing/test-nx-utils/src/lib/utils/nx.unit.test.ts
```diff
@@ -0,0 +1,84 @@
+import * as process from 'node:process';
+import { createTreeWithEmptyWorkspace } from 'nx/src/generators/testing-utils/create-tree-with-empty-workspace';
+import { describe, expect } from 'vitest';
+import { executorContext, registerPluginInWorkspace } from './nx';
+
+describe('executorContext', () => {
+  it('should create context for given project name', () => {
+    expect(executorContext('my-lib')).toStrictEqual({
```
Tlacenka, line 8: ⚙️ TypeScript: To get auto-complete if there is a non-trivial type to match, you can use `satisfies` or a type argument. Applicable in multiple places.

Tlacenka (commented):
Nice prep for testing `nx`. 👏 Did not get to this earlier so at least added a few comments now.

testing/test-nx-utils/src/lib/utils/tree.integration.test.ts
matejchalk: Good catch, I missed that :see_no_evil: Yes, the test should clean up the file system changes after it runs :+1:

PR #778, opened 2024-08-02
feat(nx-plugin): extend config generator

BioPhoton:
Better generator logic to create config files


packages/nx-plugin/src/generators/configuration/generator.ts
```diff
@@ -29,14 +34,10 @@ export function addTargetToProject(
...
-  const { targetName, project, skipTarget } = options;
-
-  if (skipTarget) {
-    return;
-  }
+  const { targetName, project } = options;
 
   const codePushupTargetConfig = {
-    executor: '@code-pushup/nx-plugin:autorun',
+    executor: `@code-pushup/nx-plugin:autorun`,
```
BioPhoton, line 40:
```suggestion
    executor: '@code-pushup/nx-plugin:autorun',
```

packages/nx-plugin/src/generators/configuration/code-pushup-config.ts
```diff
@@ -0,0 +1,71 @@
...
+import {
+  formatArrayToJSArray,
+  formatArrayToLinesOfJsString,
+  formatObjectToFormattedJsString,
+  normalizeExecutableCode,
+  normalizeItemOrArray,
+} from './utils';
+
+export const DEFAULT_IMPORTS = [
+  "import type { CoreConfig } from '@code-pushup/models;'",
```
matejchalk, line 15: Invalid syntax here:

```suggestion
  "import type { CoreConfig } from '@code-pushup/models';",
```

packages/nx-plugin/src/generators/configuration/__snapshots__/code-pushup-config.integration.test.ts.snap
```diff
@@ -0,0 +1,20 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`generateCodePushupConfig options > should create code-pushup.config.ts with options in tree 1`] = `
+"import type { CoreConfig } from '../dist/packages/models;
+import * as myPlugin from my-plugin";
+import {myPluginCategory} from "my-plugin";
```
matejchalk, line 6: This doesn't look like valid syntax, there are missing opening and closing quotes in the imports.

I would suggest replacing `.toMatchSnapshot()` with [`.toMatchFileSnapshot(...)`](https://vitest.dev/api/expect.html#tomatchfilesnapshot), that way the output file can have a `.ts` extension and it's easier to see if it's correct.

packages/nx-plugin/src/generators/configuration/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export function formatArrayToLinesOfJsString(
+  lines?: string[],
+  separator = '\n',
+) {
+  if (lines == null || lines.length === 0) {
+    return;
+  }
+  return `${lines.join(separator).replace('"', '')}`;
```
matejchalk, line 54: This only replace the 1st occurrence. If you want to remove all quotes, use `.replaceAll('"', '')` or `.replace(/"/g, '')`. Also, the template string is unnecessary.

```suggestion
  return lines.join(separator).replace(/"/g, '');
```

matejchalk (changes requested):

packages/nx-plugin/src/generators/configuration/__snapshots__/code-pushup-config.integration.test.ts.snap
BioPhoton: Done

packages/nx-plugin/src/generators/configuration/__snapshots__/root-code-pushup.config.ts
matejchalk, line 1: To prevent these files from interfering with our coverage reports, could you adjust the Vitest configs using [`coverage.exclude`](https://vitest.dev/config/#coverage-exclude)? Adding something like `**/__snapshots__/**` to the default patterns should do the trick.

image

BioPhoton: Done!

matejchalk (approved):

Issue #739, opened 2024-07-10
Npx command to setup default config


grizlizli:
### User story

As a developer, I would like to setup the CLI as easy as possible.

Running the command such as `npx code-pushup init` should create code-pushup.config.[ `ts` ~| `js` | `mjs`~ ] file.
init command should have arguments for file extension, default ts.

### Acceptance criteria

Running the above mentioned command should generate the config file in the project.

**Add package `create-cli`:**
- [x] create `create-cli` package to enable the `npx init @code-pushup/cli` command.
   The init command translates to `npm exec @code-pushup/create-cli`.  
   #779 
- [x] create init script that is able to reuse the `nx-plugin` logic

**Extend `nx-plugin`**
- ~extend the `ext` to accept a `ext` options to specify the file extension.~
- ~auto detect typescript and take `ts` if given. otherwise take `mjs` as extension.~

### Implementation details

Generate a config file in the root of the project.

vmasek:
Please finalize the issue for refinement

BioPhoton:
As this has quite a decent overlap with the nx generators we could do the following:

0. `npm init @code-pushup/cli` -> `npm exec @code-pushup/create-cli`
1. create a publishable package `create-cli`
2. inside use `nx g nx-plugin:config` 
2.1. I have to figure out how to execute the generator without a nx.json prelsent.

  @matejchalk WDYT?

matejchalk:
> As this has quite a decent overlap with the nx generators we could do the following:
> 
> 0. `npm init @code-pushup/cli` -> `npm exec @code-pushup/create-cli`
> 1. create a publishable package `create-cli`
> 2. inside use `nx g  @code-pushup/nx-plugin:configuration`
>    2.1. I have to figure out how to execute the generator without a nx.json prelsent.

Yes! Having a `create-` package would give us compatibility with non-Nx codebases, since this package initializer format is supported by [NPM](https://docs.npmjs.com/cli/v10/commands/npm-init), [PNPM](https://pnpm.io/cli/create) and [Yarn 1](https://classic.yarnpkg.com/lang/en/docs/cli/create/).

BioPhoton:
Updated issue and started work.

PR #779, opened 2024-08-02
feat(create-cli): package to align with package managers init command

BioPhoton:
> [!NOTE]
> I get a resource blocked error in in the skipped tests which I can not resolve.

This PR introduces a "`create`" package for the `cli` package. 
Many package manager support this format.

`npx init @code-pushup/cli` => `npm exec @code-pushup/create-cli`

It includes:
- adding the new package
- adding logic to execute the `nx-plugin` logic
- extend the `configuration` generator with a `skipNxJson` option

Related to:
#739 cc @grizlizli


packages/create-cli/src/lib/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+    teardown: false,
+  };
+
+  try {
+    const stats = await stat(filename);
+    if (!stats.isFile()) {
+      await writeFile(filename, content);
+    }
+  } catch (error) {
+    if ((error as Error).message.includes('no such file or directory')) {
```
matejchalk, line 30: Safer handling:

```suggestion
    if (error instanceof Error && error.message.includes('no such file or directory')) {
```

packages/create-cli/src/lib/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+      await writeFile(filename, content);
+      return {
+        ...setupResult,
+        teardown: true,
+      };
+    } else {
+      console.error((error as Error).toString());
```
matejchalk, line 37:
```suggestion
      console.error(error);
```

packages/create-cli/src/lib/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+    NX_JSON_CONTENT,
+  );
+  const { teardown: projectJsonTeardown } = await setupFile(
+    PROJECT_JSON_FILENAME,
+    PROJECT_JSON_CONTENT,
+  );
+
+  const projectJsonContent = (
+    await readFile(PROJECT_JSON_FILENAME, 'utf8')
+  ).toString();
```
matejchalk, line 64: It's already a string because you specified `utf8` encoding.

```suggestion
  const projectJsonContent = await readFile(PROJECT_JSON_FILENAME, 'utf8');
```

packages/create-cli/src/lib/utils.ts
```diff
@@ -0,0 +1,93 @@
...
+  const toDelete: Promise<unknown>[] = [];
+  if (nxJsonTeardown) {
+    // eslint-disable-next-line functional/immutable-data
+    toDelete.push(rm(NX_JSON_FILENAME));
+  }
+  if (projectJsonTeardown) {
+    // eslint-disable-next-line functional/immutable-data
+    toDelete.push(rm(PROJECT_JSON_FILENAME));
+  }
+  await Promise.all(toDelete);
```
matejchalk, line 92: Can be refactored without mutations:

```suggestion
  const filesToDelete = [
    ...nxJsonTeardown ? [NX_JSON_FILENAME] : [],
    ...projectJsonTeardown ? [PROJECT_JSON_FILENAME] : [],
  ];
  await Promise.all(filesToDelete.map(rm));
```

packages/create-cli/src/lib/constants.ts
```diff
@@ -0,0 +1,12 @@
+export const NX_JSON_FILENAME = 'nx.json';
+export const NX_JSON_CONTENT = JSON.stringify({
+  $schema: './node_modules/nx/schemas/nx-schema.json',
+  releaseTagPattern: 'v{version}',
```
matejchalk, line 4: Why include this? Version tags don't seem relevant.

packages/create-cli/project.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "create-cli",
+  "$schema": "../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "packages/create-cli/src",
+  "projectType": "application",
+  "implicitDependencies": ["nx-plugin"],
```
matejchalk, line 6: The dependency on `nx-plugin` is already clear from imports, I don't see any need to add it as an implicit dependency too.

```suggestion
```

packages/nx-plugin/src/generators/init/generator.integration.test.ts
```diff
@@ -60,4 +60,11 @@ describe('init generator', () => {
       ),
     ).toHaveLength(0);
   });
+
+  it('should skip nx.Json', () => {
```
matejchalk, line 64:
```suggestion
  it('should skip nx.json', () => {
```

packages/nx-plugin/src/index.ts
```diff
@@ -1,4 +1,7 @@
-export { InitGeneratorSchema } from './generators/init/schema';
-export { initGenerator } from './generators/init/generator';
+export * from './internal/versions';
+export { type InitGeneratorSchema } from './generators/init/schema';
+export { initGenerator, initSchematic } from './generators/init/generator';
 export { ConfigurationGeneratorOptions } from './generators/configuration/schema';
 export { configurationGenerator } from './generators/configuration/generator';
+export { executeProcess, ProcessConfig } from './internal/execute-process';
+export { objectToCliArgs } from './executors/internal/cli';
```
matejchalk, line 7: Exporting files from `internal` folder seems like an anti-pattern :raised_eyebrow: 

Some of these exports are just copies from `utils` package (e.g. `executeProcess`, `objectToCliArgs`), why not have `create-cli` import them from `utils`. 

I see you have `create-cli` configured as CommonJS, I'd suggest changing it to ESM. Then it can import both CJS and ESM.

packages/nx-plugin/src/internal/execute-process.ts
```diff
@@ -0,0 +1,177 @@
...
+ *      error: handleError,
+ *      complete: cleanLogs,
+ *    }
+ * });
+ *
+ * console.info(result);
+ *
+ * @param cfg - see {@link ProcessConfig}
+ */
+export function executeProcess(cfg: ProcessConfig): Promise<ProcessResult> {
```
matejchalk, line 138: Since the `executeProcess` function is already asynchronous, there's no disadvantage to having a dynamic import. This gets around the CJS vs ESM problem, so you can import the original implementation from `utils` without duplicating it.

```ts
export async function executeProcess(cfg: ProcessConfig): Promise<ProcessResult> {
  const { executeProcess } = await import('@code-pushup/utils');
  return executeProcess(cfg);
}
```

packages/create-cli/src/lib/utils.ts
BioPhoton: Thanks for this improvement! Way better!

packages/nx-plugin/src/internal/execute-process.ts
BioPhoton: Don't we have the problem that @poppins/cliui is not ESM ready?

packages/nx-plugin/src/index.ts
BioPhoton: Thought it has to be CJS... Will change it.

packages/nx-plugin/src/internal/execute-process.ts
matejchalk: The problem is that the Nx Plugin has to CommonJS, which is the less compatible format. CJS can only import other CJS modules statically, while ESM can import both ESM and CJS. To import ESM (`utils`) into CJS (`nx-plugin`), you have to use a dynamic import. Because dynamic imports make the code async, it's incovienent for most of our utils. But for asynchronous functions like `executeProcess` the dynamic import doesn't add an extra inconvenience, since the usage is already async anyway.

As for `@poppins/cliui`, it's the other way around. The package is built only for the modern ESM format. That's a problem for CJS-only modules like Nx plugins.

packages/create-cli/src/lib/constants.ts
BioPhoton: it was the output of the generateWorkspace command. I copied it over.
removed the `releaseTagPattern` now

packages/nx-plugin/src/index.ts
BioPhoton: Update:
I changed to CJS and it seems the imports don't work anymore. In the current state you should be able to adjust tsconfig and reproduce it.

matejchalk: If you change `create-cli` to ESM and import directly from `utils`, then it works just fine. Got it working in 263e2e7bdcdea6adb0cf684ec7a23ff63ada94d7, along with some E2E test fixes (see [CI statuses](https://github.com/code-pushup/cli/actions/runs/10384678810/job/28752220308)).

BioPhoton: Thanks a lot!!

e2e/create-cli-e2e/mocks/create-npm-workshpace.ts
```diff
@@ -0,0 +1,14 @@
+import {mkdir, writeFile} from "node:fs/promises";
+import {join} from "node:path";
+
+export async function createNpmWorkspace(cwd: string) {
+  await mkdir(cwd, {recursive: true});
```
matejchalk, line 5: Not formatted:

```suggestion
import { mkdir, writeFile } from "node:fs/promises";
import { join } from "node:path";

export async function createNpmWorkspace(cwd: string) {
  await mkdir(cwd, { recursive: true });
```

packages/create-cli/src/index.ts
```diff
@@ -0,0 +1,5 @@
+#! /usr/bin/env node
+import { initCodePushup } from './lib/init';
+
+// eslint-disable-next-line unicorn/prefer-top-level-await
+initCodePushup().catch(console.error);
```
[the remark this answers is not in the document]
matejchalk: Now that it's ESM, top-level `await` should work.

matejchalk (dismissed):

BioPhoton:
@matejchalk did you ever saw such an issue? 
https://github.com/code-pushup/cli/actions/runs/10407760960/job/28823712012

I am not really able to get it...

matejchalk:
> @matejchalk did you ever saw such an issue? https://github.com/code-pushup/cli/actions/runs/10407760960/job/28823712012
> 
> I am not really able to get it...

I hadn't, but tried a few things to replicate and at a certain point I reached a similar error: `ENOTEMPTY: directory not empty, rmdir 'tmp/create-cli-e2e/npm-exec'`. I got there by removing `tmp` folder before running `npx nx run-many -t e2e --skipNxCache --parallel 1`. But I'm not able to reproduce it reliably, most of the time the E2E tests pass for me.

image

Fix could be to add `force: true` flag to `rm` call in `afterEach`.

e2e/create-cli-e2e/tests/init.e2e.test.ts
```diff
@@ -48,8 +49,9 @@ describe('create-cli-node', () => {
     );
   });
 
-  it('should execute package correctly over npm init', async () => {
-    const cwd = join(baseDir, 'init');
+  // eslint-disable-next-line vitest/no-disabled-tests
+  it.skip('should execute package correctly over npm init', async () => {
```
matejchalk, line 53: Could you disable the `node ...` test in favour of leaving this `npm init` test enabled? This test is closest to how a real user would execute it.

global-setup.e2e.ts
```diff
@@ -8,22 +8,29 @@ const localRegistryNxTarget = '@code-pushup/cli-source:local-registry';
...
+    console.info('Installing packages');
+    execFileSync(
+      'npx',
+      ['nx', 'run-many', '--targets=npm-install', '--parallel=1'],
+      { env: process.env, stdio: 'inherit', shell: true },
+    );
+    await setupTestFolder('tmp/e2e');
+  } catch (error) {
+    console.info('setup error: ' + error.message);
+  }
```
matejchalk, line 23: Is this `try`/`catch` appropriate? It will proceed with running tests if the setup fails, that's probably not what we want, right?

tools/scripts/publish.mjs
```diff
@@ -58,6 +58,7 @@ invariant(
   `Could not find "build.options.outputPath" of project "${name}". Is project.json configured  correctly?`,
```
matejchalk, line 58:
```suggestion
  `Could not find "build.options.outputPath" of project "${name}". Is project.json configured correctly?`,
```

global-setup.e2e.ts
```diff
@@ -8,18 +8,28 @@ const localRegistryNxTarget = '@code-pushup/cli-source:local-registry';
...
   await globalSetup();
+  await setupTestFolder('tmp/local-registry');
+  await setupTestFolder('tmp/e2e');
+
   try {
-    await setupTestFolder('tmp/local-registry');
     await startLocalRegistry({ localRegistryTarget: localRegistryNxTarget });
+  } catch (error) {
+    console.error(
+      'Error starting local verdaccio registry: \nr: ' + error.message,
```
matejchalk, line 18:
```suggestion
      'Error starting local verdaccio registry:\n' + error.message,
```

matejchalk (approved):

e2e/create-cli-e2e/tests/init.e2e.test.ts
BioPhoton: It was flay in past runs, but sems stable now.

matejchalk (approved):

PR #794, opened 2024-08-14
refactor(plugin-lighthouse): format audit value

### LINKED ISSUE #744: Custom matcher for paths on different OS
### User story

As a developer I want to have a good signal to noise ration and want to write my tests with as less boilerplate as possible. 
Paths are  environment specific and cause boilerplate to get properly tested.
Therefor I need a [custom matcher](https://vitest.dev/guide/extending-matchers)  for paths.

Bad example:
```ts

expect(toNormalizedPath(process.cwd())).toEqual(
      expect.stringContaining(
        toNormalizedPath('project-root/other-folder'),
      ),
    );
```
Suggested matcher:
- `expect(myPath).toContainNormalizedPath('...')` 
- `expect(myPath).toEndWithNormalizedPath('...')` 

Good example:
```ts

expect(process.cwd()).toEndsWithNormalizedPath('project-root/other-folder');
```

### Acceptance criteria

I want to be able to use the custom matcher in the following libs:
- [ ] vitest
- jest (optional)

Internal logic:
- [ ] it handles path differences of all OS
I want to be able to have different matcher present:
- [ ] `toContainNormalizedPath('...')` 
- [ ] `toEndWithNormalizedPath('...')`

### LINKED ISSUE #774: Show passed/failed value instead of 0 for scored Lighthouse audits
### User story

As a user, I expect that when an audit's score changes there is also some change to the value to give it more context. This isn't true for some Lighthouse audits (e.g. _Displays images with correct aspect ratio_ or _Background and foreground colors have a sufficient contrast ratio_), where the value is always `0` even if the score changes.

Image

### Acceptance criteria

- [x] if an audit from `lighthouse` provides a score without a raw/display value, then use the score to determine the `value` and `displayValue` for the resulting CP audit
  - [x] set raw `value` equal to `score`
  - [x] set `displayValue` to `'passed'` if `score` equals `1`, otherwise set it to `failed`

### Implementation details

_No response_

hanna-skryl:
Closes #774 

- [x] Update `value` and `displayValue` formatting and test the change
- [x] Refactor `toAuditOutputs` by extracting base audit output formatting and details processing into functions

### DIFF
--- a/packages/plugin-lighthouse/src/lib/runner/utils.ts
+++ b/packages/plugin-lighthouse/src/lib/runner/utils.ts
@@ -1,9 +1,10 @@
 import { bold } from 'ansis';
-import type { Config } from 'lighthouse';
+import type { Config, FormattedIcu } from 'lighthouse';
 import log from 'lighthouse-logger';
 import desktopConfig from 'lighthouse/core/config/desktop-config.js';
 import experimentalConfig from 'lighthouse/core/config/experimental-config.js';
 import perfConfig from 'lighthouse/core/config/perf-config.js';
+import Details from 'lighthouse/types/lhr/audit-details';
 import { Result } from 'lighthouse/types/lhr/audit-result';
 import { AuditOutput, AuditOutputs } from '@code-pushup/models';
 import { importModule, readJsonFile, ui } from '@code-pushup/utils';
@@ -38,6 +39,32 @@ export class LighthouseAuditParsingError extends Error {
   }
 }
 
+function formatBaseAuditOutput(lhrAudit: Result): AuditOutput {
+  const { id: slug, score, numericValue, displayValue } = lhrAudit;
+  return {
+    slug,
+    score: score ?? 1, // score can be null
+    value: numericValue ?? score ?? 0, // not every audit has a numericValue
+    displayValue:
+      displayValue ??
+      (score === 1 ? 'passed' : score === 0 ? 'failed' : undefined),
+  };
+}
+
+function processAuditDetails(
+  auditOutput: AuditOutput,
+  details: FormattedIcu<Details>,
+): AuditOutput {
+  try {
+    const parsedDetails = toAuditDetails(details);
+    return Object.keys(parsedDetails).length > 0
+      ? { ...auditOutput, details: parsedDetails }
+      : auditOutput;
+  } catch (error) {
+    throw new LighthouseAuditParsingError(auditOutput.slug, error as Error);
+  }
+}
+
 export function toAuditOutputs(
   lhrAudits: Result[],
   { verbose = false }: { verbose?: boolean } = {},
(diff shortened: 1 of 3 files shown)


packages/plugin-lighthouse/src/lib/runner/utils.ts
```diff
@@ -38,43 +39,46 @@ export class LighthouseAuditParsingError extends Error {
   }
 }
 
+function formatBaseAuditOutput(lhrAudit: Result): AuditOutput {
+  const { id: slug, score, numericValue, displayValue } = lhrAudit;
+  return {
+    slug,
+    score: score ?? 1, // score can be null
+    value: numericValue ?? score ?? 0, // not every audit has a numericValue
```
hanna-skryl, line 47: ❓ I copied the comments from the existing implementation and now wonder if they are superfluous. One can easily see the `score` and `numericValue` typing, and the comments don't add more details to that.

packages/plugin-lighthouse/src/lib/runner/utils.ts
```diff
@@ -38,43 +39,46 @@ export class LighthouseAuditParsingError extends Error {
...
+    const parsedDetails = toAuditDetails(details);
+    return Object.keys(parsedDetails).length > 0
+      ? { ...auditOutput, details: parsedDetails }
+      : auditOutput;
+  } catch (error) {
+    throw new LighthouseAuditParsingError(auditOutput.slug, error as Error);
+  }
+}
+
 export function toAuditOutputs(
```
hanna-skryl, line 68: ❓ There is a `lhrToAuditOutputs` function in examples that pretty much mirrors the `toAuditOutputs` logic. Should it also be updated?

packages/plugin-lighthouse/src/lib/runner/utils.ts
BioPhoton: Agree! you can remove them

packages/plugin-lighthouse/src/lib/runner/utils.ts
BioPhoton: No. The implementation is the examples is just to showcase. I could use any other less complex CLI there... 
Maybe I can think about a better example that does not overlap with our official plugins or move it into the example-plugins repo.

packages/plugin-lighthouse/src/lib/runner/utils.ts
```diff
@@ -38,43 +39,46 @@ export class LighthouseAuditParsingError extends Error {
...
 
+function formatBaseAuditOutput(lhrAudit: Result): AuditOutput {
+  const { id: slug, score, numericValue, displayValue } = lhrAudit;
+  return {
+    slug,
+    score: score ?? 1, // score can be null
+    value: numericValue ?? score ?? 0, // not every audit has a numericValue
+    displayValue:
+      displayValue ??
+      (score === 1 ? 'passed' : score === 0 ? 'failed' : undefined),
```
matejchalk, line 50: I just realized (from your test example) that Lighthouse uses `scoreDisplayMode: 'binary'` :open_mouth: 

That seems like a more reliable indicator for when it would be appropriate to display `passed` or `failed`.

```suggestion
      (scoreDisplayMode === 'binary'
        ? (score === 1
          ? 'passed'
          : 'failed')
        : undefined),
```

packages/plugin-lighthouse/src/lib/runner/utils.ts
hanna-skryl, line 47: ❗ An edge case: Audits like `"unsized-images"` whose values are not binary but that also don't have a `numericValue` now display an unformatted score in the value column (e.g., 0.5 when the score is 50%). Should it be reverted (a fallback of `0`)?

BioPhoton (dismissed):
Nice improvement for readability! 🏆

packages/plugin-lighthouse/src/lib/runner/utils.ts
matejchalk: I think this is fine for the raw value. But it could be a nice improvement to also set `displayValue` using our `formatReportScore` function in this case :thinking:

hanna-skryl: Done ✅ 

An audit `displayValue` should now be a formatted score when `displayValue` is missing and `scoreDisplayMode` is not binary.

matejchalk (approved):

PR #796, opened 2024-08-15
refactor: polish publish code

BioPhoton:


global-setup.ts
```diff
@@ -1,3 +1,5 @@
+import { setupTestFolder } from './testing/test-setup/src';
+
```
getlarge, line 2: `setupTestFolder` is imported but not called, was it an oversight? Or has this import a side effect?

tools/scripts/publish.plugin.ts
```diff
@@ -0,0 +1,47 @@
...
+export const createNodes: CreateNodes = [
+  '**/project.json',
+  (projectConfigurationFile: string, opts, context: CreateNodesContext) => {
+    const root = dirname(projectConfigurationFile);
+    const projectConfiguration: ProjectConfiguration = readJsonFile(
+      projectConfigurationFile,
+    );
+
+    const isPublishable = Boolean(projectConfiguration?.targets?.publish);
+    const { name: projectName } = projectConfiguration;
```
getlarge, line 14: The publishable projects should contain a `package.json`, right?
Wouldn't `**/package.json` be a better glob?
It would remove the need to keep an empty `publish` target.

tools/scripts/start-local-registry.ts
```diff
@@ -37,70 +49,55 @@ function startLocalRegistry({
...
+          }
+        },
+        onComplete: code => {
+          console.info('local registry onComplete', code);
+        },
+      },
+    }).catch(error => {
+      if (error.message !== 'Failed to start verdaccio: undefined') {
+        reject(error);
       }
```
getlarge, line 101: The promise would remain unfulfilled `if error.message !== 'Failed to start verdaccio: undefined'`.

tools/scripts/start-local-registry.ts
```diff
@@ -37,70 +49,55 @@ function startLocalRegistry({
...
+              stop: () => {
+                // this makes the process throw
+                childProcess.kill();
+                unconfigureRegistry(registryData);
+                execSync(
+                  `npm config delete ${registryData.registryNoProtocol}/:_authToken`,
+                );
+              },
+            });
+          }
```
getlarge, line 87: I wonder if instead of having to call `stop` manually, this could be replaced by an [Disposable](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html) ...  in any case, this logic would be called

tools/scripts/utils.ts
```diff
@@ -0,0 +1,159 @@
...
+  const registry = `${protocol}:${registryNoProtocol}`;
+
+  return {
+    protocol,
+    host,
+    port,
+    registryNoProtocol,
+    registry,
+  };
+}
```
getlarge, line 82: good one! I guess the registry parsing was missing in the original refactor.

getlarge (commented):
A pleasure to read... 
But I'm worried a newcomer might struggle to catch the e2e setup flow.
Maybe a diagram could help them?

tools/scripts/publish.plugin.ts
BioPhoton: Also buildable targets have a `package.json`. I updated the PR description.

tools/scripts/utils.ts
```diff
@@ -0,0 +1,159 @@
...
+  const protocol = protocolMatch?.groups?.proto;
+
+  if (!protocol || !['http', 'https'].includes(protocol)) {
+    throw new Error(
+      `Invalid protocol ${protocol}. Only http and https are allowed.`,
+    );
+  }
+
+  const host = 'localhost';
+  //const protocol = 'http';
```
BioPhoton, line 71:
```suggestion
```

tools/scripts/utils.ts
BioPhoton: Thanks!

global-setup.ts
BioPhoton, line 2:
```suggestion
```

BioPhoton: removed

tools/scripts/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+  return {
+    protocol,
+    host,
+    port,
+    registryNoProtocol,
+    registry,
+  };
+}
+
+export type ProcessResult = {
```
matejchalk, line 83: From here on down it's basically duplicating `@code-pushup/utils`.

Why not `import { executeProcess } from '../../packages/utils/src'` and add the extra callback parameters to the main implementation?

tools/scripts/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+  return execSync('git describe --tags --abbrev=0')
+    .toString()
+    .trim()
+    .replace(/^v/, '');
+}
+
+export function parseRegistryData(stdout: string): RegistryData {
+  const port = parseInt(
+    stdout.toString().match(/localhost:(?<port>\d+)/)?.groups?.port,
+  );
```
matejchalk, line 58: There's a type error here because of potentially assigning `undefined` to a `string` parameter.

image


One solution is to throw for invalid:

```suggestion
  const port = parseInt(
    stdout.toString().match(/localhost:(?<port>\d+)/)?.groups?.port ?? '',
  );
  if (isNaN(port)) {
    throw new Error('Invalid port number');
  }
```

Another option is to define a fallback port:

```suggestion
  const port = parseInt(
    stdout.toString().match(/localhost:(?<port>\d+)/)?.groups?.port ?? '4873',
  );
```

tools/scripts/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+  // yarnv1
+  process.env.YARN_REGISTRY = registry;
+  // yarnv2
+  process.env.YARN_NPM_REGISTRY_SERVER = registry;
+  console.info(`Set NPM and yarn registry process.env`);
+  /**
+   * TODO
+   */
+  process.env.YARN_UNSAFE_HTTP_WHITELIST = host;
+  console.info(`Set yarn whitelíst process.env`);
```
matejchalk, line 28: Could you add some more info to the TODO comments? Otherwise it's unclear what is missing exactly.

tools/scripts/start-local-registry.ts
BioPhoton: done

tools/scripts/start-local-registry.ts
BioPhoton: will move this comment in the next PR #798

nx.json
```diff
@@ -109,5 +109,6 @@
       }
     },
     "releaseTagPattern": "v{version}"
-  }
+  },
+  "plugins": ["./tools/scripts/publish.plugin.ts"]
```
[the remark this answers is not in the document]
BioPhoton: I would intro a `publish` and a `npm` folder with 2 different plugins in the followup PR

tools/scripts/utils.ts
BioPhoton: done

nx.json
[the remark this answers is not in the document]
BioPhoton: I will move this refactor into #798

tools/scripts/utils.ts
BioPhoton: extended existing code and reused it.

packages/utils/src/lib/execute-process.ts
```diff
@@ -133,33 +139,38 @@ export type ProcessObserver = {
...
   const date = new Date().toISOString();
   const start = performance.now();
 
   return new Promise((resolve, reject) => {
     // shell:true tells Windows to use shell command for spawning a child process
-    const process = spawn(command, args, { cwd, shell: true });
+    const spawnedProcess = spawn(command, args ?? [], {
+      shell: true,
+      ...options,
+    }) as ChildProcessByStdio<Writable, Readable, Readable>;
```
matejchalk, line 152: :warning: This change introduces a potential safety hole, the `stdout` and `stderr` streams can be made `null` if `stdio` is set to anything other than `'pipe'` (previously this wasn't exposed). 

image

If you want to enable callers to customizes the streams, then you should make sure it won't result in a runtime error:

image

If you want to restrict the caller to only the default `stdout`/`stderr` streams, then you should tighten the type definitions:

image

I would go with the 2nd option, since `executeProcess` is unlikely to be useful for custom `stdio` streams.

BioPhoton: went with second option and added the restricted type to the options.

tools/scripts/publish.plugin.ts
BioPhoton: Afterthought:

I will maybe use the target name `publishable` to identify publishable libs

matejchalk (approved):

PR #793, opened 2024-08-14
feat(utils): implement report-diff.md formatting for monorepos

matejchalk:
Part of #731 (only Markdown formatting)

### Scope

- Merged separate groups and audits details element into single details element to make the Markdown more compact. Applies to both standalone and monorepo comments.
  image
- Implemented function to format array of project diffs as single comment.
  - Section per project with status summary.
  - Unchanged categories are omitted from table.
  - Unchanged projects only mentioned in footer.
  - Projects sorted from most changed. 
  image
- Extracted diff formatting utilities into separate file and unit tested them.

### Next steps

- Implement CLI command to merge multiple `report-diff.json`s into single `report-diff.md`.
- Use command in GitHub Action (https://github.com/code-pushup/github-action/issues/35)

Issue #731, opened 2024-07-04
Multi-project comparison comment


matejchalk:
### User story

When running Code PushUp separately on different projects in a monorepo, I want a single comparison comment in my PRs which summarizes changes to all affected projects. 

### Acceptance criteria

- [x] implement CLI command to merge multiple report diffs into single markdown file
- [x] markdown has sections per project
- [x] project name is included in each section heading
- [x] more compact layout - only categories table shown by default, groups and audits in expandable details
- [x] code-pushup/github-action#35

### Related links

- https://github.com/code-pushup/cli/pull/706#discussion_r1660478690

PR #800, opened 2024-08-16
feat(cli): implement `merge-diffs` command

matejchalk:
Part of #731 (CLI)

New `merge-diffs` command implemented, which parses list of `report-diff.json` paths from required `--files` option and uses Markdown formatter from #793 to generate single `report-diff.md`.

Also added a `--label` option to `compare` command, which can be used to set project names for the JSON diffs (parent directory used as fallback).

### DIFF
--- a/e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
+++ b/e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
@@ -14,7 +14,7 @@ Commands:
   code-pushup compare       Compare 2 report files and create a diff file
 
   code-pushup print-config  Print config
-  code-pushup merge-diffs   Combine many report diffs into single report-diff.md
+  code-pushup merge-diffs   Combine many report diffs into a single diff file
 
 
 Global Options:
--- a/e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
+++ b/e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
@@ -14,6 +14,8 @@ Commands:
   code-pushup compare       Compare 2 report files and create a diff file
 
   code-pushup print-config  Print config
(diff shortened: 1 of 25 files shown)


packages/core/src/lib/merge-diffs.ts
```diff
@@ -0,0 +1,44 @@
...
+} from '@code-pushup/utils';
+
+export async function mergeDiffs(
+  files: string[],
+  persistConfig: Required<PersistConfig>,
+): Promise<string> {
+  const results = await Promise.allSettled(
+    files.map(async file => {
+      const json = await readJsonFile(file);
+      const diff = await reportsDiffSchema.parseAsync(json);
```
BioPhoton, line 20: I suggest adding using parseSafe to be able to add a better readable error message.

It would already help to tell the user that the error occurred because the content of the file xyz did not contain a valid report diff or similar. In the future I will ship a helper function to turn the Error into a readable text. PoC already exists.

BioPhoton (changes requested):
Nice to have a command for it and not just loose scripts!
Left comments for a slight adoption of the docs.

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
```diff
@@ -14,6 +14,8 @@ Commands:
   code-pushup compare       Compare 2 report files and create a diff file
 
   code-pushup print-config  Print config
+  code-pushup merge-diffs   Combine many report diffs into single report-diff.md
```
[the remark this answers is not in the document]
matejchalk: Updated description.

packages/core/src/lib/merge-diffs.ts
matejchalk: Agreed, improved the error messages :heavy_check_mark: Added a basic helper function in `utils` which could serve as a good basis for safely formatting error messages.

PR #810, opened 2024-08-30
refactor: derive NPM dependencies from implicit dependencies of e2e project

BioPhoton:
This PR refactors the new E2E setup and made the existing codebase more robust.

**It includes:** 
- 🔌 leverage task graph to set up dependencies for e2e 🔥🔥🔥
- 📏 reduce `global-setup.verdaccio.ts`. Now we have 1 target, `setup-deps`,  to set up e2e dependencies.
- ⛑️ deflake deletion on files with `rm(dirName, { recursive: true, force: true, maxRetries: 2 });`
- ⛑️ deflake killing verdaccio as for some reasons pid is NaN occasionally 
- ⛑️ simplify kill process in favour of OS bug

**It unfortunately also includes:** 
-  changing the existing tests in `cli-e2e` to make it work. We lost testing of the registered command 
   [before: `code-pushup ...` after:   `npx @code-pushup/cli ...`](https://github.com/code-pushup/cli/pull/810/files#r1739409924)
- existing [snapshots changed in formatting](https://github.com/code-pushup/cli/pull/810/files#r1739412166)
- commenting out compare e2e test due to [unknown diff in snapshot](https://github.com/code-pushup/cli/pull/810/files#r1739418147).  

### Implementation details

In my past attempts I was missing that params forwarding is off by default. Also the setup was too stable to make meaningful changes.

Targets are now correctly configures by using object to describe dependencies:

```json
{
    "TARGET_NAME": {
        "dependsOn": [
            {
                "target": "DEPENDING_TARGET_NAME",
                "projects": "dependencies",
                "params": "forward"
            }
        ]
    }
}
```

#### Task graph before

| cli-e2e | nx-plugin-e2e | create-cli-e2e |
| ------- | -------------- | --------------  |
| Screenshot 2024-08-30 at 18 23 50 | Screenshot 2024-08-30 at 18 23 26 | Screenshot 2024-08-30 at 18 23 36 |

#### Task graph after

| cli-e2e | nx-plugin-e2e | create-cli-e2e |
| ------- | -------------- | --------------  |
| Screenshot 2024-08-30 at 18 12 00 | Screenshot 2024-08-30 at 18 11 34 | Screenshot 2024-08-30 at 18 12 07 |


e2e/cli-e2e/tests/collect.e2e.test.ts
```diff
@@ -46,7 +46,7 @@ describe('CLI collect', () => {
   /* eslint-enable @typescript-eslint/no-unused-vars */
 
   beforeEach(async () => {
-    await cleanTestFolder('tmp/e2e');
+    await cleanTestFolder('tmp/e2e/react-todos-app');
```
BioPhoton, line 49: This would delete other parallel running tests

e2e/create-cli-e2e/project.json
```diff
@@ -18,6 +18,6 @@
       }
     }
   },
-  "implicitDependencies": ["create-cli"],
+  "implicitDependencies": ["models", "utils", "core", "cli", "create-cli"],
```
matejchalk, line 21: Shouldn't it be enough to set `cli` as implicit dependency? The `core`, `models` and `utils` dependencies would then automatically become indirect dependencies based on Nx graph.

Same goes for other `e2e/*-e2e/project.json`s.

e2e/cli-e2e/project.json
```diff
@@ -19,6 +19,10 @@
     }
   },
   "implicitDependencies": [
+    "models",
+    "utils",
+    "core",
+    "cli",
```
matejchalk, line 25: The `cli` is duplicated here, BTW.

tools/src/verdaccio/verdaccio.plugin.ts
```diff
@@ -54,6 +63,7 @@ export const createNodes: CreateNodes = [
             config,
             storage,
             preTargets,
+            deps: projectConfiguration.implicitDependencies,
```
matejchalk, line 66: Ah, this is why they're all explicit. It's confusing the `implicitDependencies` now work differently than is usual for an Nx workspace (also only for e2e projects, which further complicates things) :/

The `@nx/devkit` has helper functions for listing all project dependencies, which could be used here instead of the `implicitDependencies` array from `project.json`. That would be much easier to maintain than having to think about listing all direct and indirect dependencies in `project.json`s.

matejchalk (changes requested):

e2e/create-cli-e2e/project.json
matejchalk: Related to https://github.com/code-pushup/cli/pull/810#discussion_r1738527881

BioPhoton: ATM we don't leverage the graph for that. The reason is I did not manage to pass the args to dependencies as well as some edge case for the existing CLI-e2e tests. 

Have it on my roadmap!

BioPhoton: Updated PR, managed finally to forward target params 🎉

tools/src/verdaccio/verdaccio.plugin.ts
BioPhoton: resolved over task dependencies

global-setup.verdaccio.ts
matejchalk, line 1: Getting publishing errors locally when running `npx nx run-many -t e2e --parallel 1` :confused: 

```
       npm notice Publishing to http://localhost:4873/ with tag e2e and public access
       npm ERR! code E401
       npm ERR! Unable to authenticate, your authentication token seems to be invalid.
```

BioPhoton: i have problems with the cli-e2e ... seems i broke the old logic with the last changes :D 

On it...

BioPhoton: Hm i can revert to a green state? ... nothing else works :D

e2e/cli-e2e/tests/compare.e2e.test.ts
```diff
@@ -12,10 +12,11 @@ describe('CLI compare', () => {
         'Unstaged changes found in examples/react-todos-app, please stage or commit them to prevent E2E tests interfering',
       );
     }
-    await cleanTestFolder('tmp/e2e');
+    await cleanTestFolder('tmp/e2e/react-todos-app');
     await executeProcess({
-      command: 'code-pushup',
```
BioPhoton, line 17: Did not know how to fix thi. Would suggest just keep it for now. (outside of `react-todos-app` it is easier to test this in isolation)

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
```diff
@@ -7,50 +7,38 @@ Commands:
...
   code-pushup upload        Upload report results to the portal
   code-pushup history       Collect reports for commit history
   code-pushup compare       Compare 2 report files and create a diff file
-
   code-pushup print-config  Print config
-  code-pushup merge-diffs   Combine many report diffs into a single diff file
-
 
 Global Options:
-      --progress     Show progress bar in stdout.
```
BioPhoton, line 21: I am not sure why this is different now, but it looks better IMO :D

e2e/cli-e2e/tests/compare.e2e.test.ts
```diff
@@ -43,10 +45,12 @@ describe('CLI compare', () => {
     await cleanTestFolder('tmp/e2e');
   });
 
-  it('should compare report.json files and create report-diff.json and report-diff.md', async () => {
+  // eslint-disable-next-line vitest/no-disabled-tests
+  it.skip('should compare report.json files and create report-diff.json and report-diff.md', async () => {
```
BioPhoton, line 49: It is unclear to me why this happens. 

See error on my local machine
Screenshot 2024-08-30 at 21 42 06

matejchalk: I've figured out where the problem is - the test aren't running `@code-pushup/cli` from the local registry :disappointed: The `npx @code-pushup/cli` command installs and executes the latest version from https://registry.npmjs.org/. This explains why the `report-diff.md` output is different - the updated formatting from #793 hasn't been released yet.

e2e/cli-e2e/tests/compare.e2e.test.ts
matejchalk: Using `npx @code-pushup/cli` instead of `code-pushup` is masking the root problem - local package installation isn't working, see https://github.com/code-pushup/cli/pull/810#discussion_r1751773344

BTW the tests should probably use `npx code-pushup`, the plain `code-pushup` command would only work if it's installed globally, which is an atypical setup.

e2e/cli-e2e/tests/__snapshots__/help.e2e.test.ts.snap
matejchalk: I guess it's also because you're not running the CLI from source, as is the case for https://github.com/code-pushup/cli/pull/810#discussion_r1751773344

e2e/cli-e2e/tests/print-config.e2e.test.ts
```diff
@@ -11,8 +11,8 @@ describe('CLI print-config', () => {
   it.each(extensions)(
     'should load .%s config file with correct arguments',
     async ext => {
-      const { code, stderr, stdout } = await executeProcess({
-        command: 'npx',
+      const { code, stdout } = await executeProcess({
+        command: 'code-pushup',
```
vmasek, line 15: Is this change correct with `'@code-pushup/cli'` kept in args?

e2e/cli-e2e/vite.config.e2e.ts
```diff
@@ -16,7 +16,7 @@ export default defineConfig({
     },
     environment: 'node',
     include: ['tests/**/*.e2e.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
-    globalSetup: ['../../global-setup.e2e.ts'],
+    globalSetup: ['../../global-setup.verdaccio.ts'],
```
BioPhoton, line 19:
```suggestion
    globalSetup: ['../../global-setup.e2e.ts'],
```

.gitignore
```diff
@@ -48,3 +48,8 @@ Thumbs.db
 
 # Nx workspace cache
 .nx/cache
+
+# e2e tests in examples
+/examples/react-todos-app/storage
+/examples/react-todos-app/package-lock.json
+/examples/react-todos-app/node_modules
```
BioPhoton, line 55:
```suggestion

```

e2e/cli-e2e/tests/print-config.e2e.test.ts
BioPhoton: it was trying to use the new setup also in the cli-e2e, but there where too many edge cases.
Now all reverted.

PR #811, opened 2024-09-04
build: update Nx 19 & other dependencies

### LINKED ISSUE #771: ci: migrate to Nx 19
### User story

Nx 19 provides many helper functions for plugin generators. To not reimplement most of this code we have to migrate.

### Acceptance criteria

- [ ] Nx is on v19 with all targets working

### Implementation details

_No response_

### LINKED ISSUE #776: reduce dependencies
### User story

For our current changelog flow we use conventional cimmits and the `conventional-changelog-angular@^7.0.0`package.
This has a dependency to `@angular-devkit/build-angular@^17`. 

### Acceptance criteria

- [ ] Check if it is even in use
- [ ] Get rid of angular as dependency by switching or re-creting the conventions

### Implementation details

_No response_

### LINKED ISSUE #835: Don't bundle build output
## The problem

Currently, our packages are bundled with ESBuild. This has some strange consequences.

For example, although `@code-pushup/cli` depends on other packages like `@code-pushup/utils` via `dependencies` in `package.json`, the distributed `cli` package actually doesn't `import` any of our other packages, their code is included in the `index.js` or `bin.js` bundle. The reason ESBuild doesn't treat other `@code-pushup` packages as external dependencies is because the imports are treated as relative paths after resolving TypeScript path aliases.

This leads to weird cases like the #834 bug - the `@code-pushup/cli` source code doesn't directly import `vscode-material-icons` (hence why it should be OK that it's not in its `package.json`), but in the bundled code it is imported directly (hence why Yarn 4 complains about it). 

Also, bundling the package into one huge file makes it harder to diagnoze errors reported by customers, because the error stack just shows some random line in the massive `index.js` file, so it's harder to link to our source code.

## Proposal

Our packages are for Node only, not the browser. Therefore it doesn't make sense to use a bundler. All we need is to transpile `.ts` files to `.js` and `.d.ts` files, so plain old `tsc` should suffice.

## Acceptance criteria

- [x] replace [`@nx/esbuild:esbuild`](https://nx.dev/nx-api/esbuild/executors/esbuild) executor with [`@nx/js:tsc`](https://nx.dev/nx-api/js/executors/tsc)
- [x] make sure package JSON of each sub-package is correct

vmasek:
closes #771 
closes #776 
closes #835


vmasek:
Investigation of failing CI in progress 👀

e2e/cli-e2e/tests/print-config.e2e.test.ts
```diff
@@ -6,13 +6,14 @@ const extensions = ['js', 'mjs', 'ts'] as const;
 export const configFilePath = (ext: (typeof extensions)[number]) =>
   join(process.cwd(), `e2e/cli-e2e/mocks/fixtures/code-pushup.config.${ext}`);
 
-describe('print-config', () => {
+describe('CLI print-config', () => {
   it.each(extensions)(
     'should load .%s config file with correct arguments',
     async ext => {
-      const { code, stderr, stdout } = await executeProcess({
```
vmasek, line 13: @BioPhoton here the check for `stderr` can be skipped or we could check it for just errors (ignoring warning or any other pollution)

BioPhoton: Thanks!

e2e/cli-e2e/vite.config.e2e.ts
```diff
@@ -7,6 +7,7 @@ export default defineConfig({
   test: {
     reporters: ['basic'],
     testTimeout: 140_000,
+    hookTimeout: 20_000,
```
vmasek, line 10: @BioPhoton this you need to bump, I found out it was not the tests set time outing but the individual hooks because of the cleanup on windows

testing/test-setup/src/lib/test-folder.setup.ts
```diff
@@ -10,5 +12,20 @@ export async function cleanTestFolder(dirName: string) {
...
 
 export async function teardownTestFolder(dirName: string) {
-  await rm(dirName, { recursive: true, force: true });
+  try {
+    await rm(dirName, {
+      recursive: true,
+      force: true,
+      // eslint-disable-next-line no-magic-numbers
+      maxRetries: 3,
+      retryDelay: 100,
```
vmasek, line 21: @BioPhoton I also made the retry delay smaller and I would even say, lets have just two of `maxRetries` because if it is the filesystem lock problem on windows, then nothing will help and we just unnecessarily wait and if it is something else 2nd retry should pass.

BioPhoton: Thx! also implemented it! It's a good improvement.

BioPhoton:
Ima run e2e in parallel is not a must have for this pr

esbuild.config.js
```diff
@@ -45,6 +45,7 @@ module.exports = {
 
           try {
             await promisify(exec)(
+              // `npx tsc --emitDeclarationOnly --project ${tsConfig} --outDir dist`,
```
BioPhoton, line 48: Left over?

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           node-version-file: .nvmrc
           cache: npm
       - name: Audit production dependencies
-        run: npm audit --omit=dev --omit=optional
+        run: npm audit --omit=dev --omit=optional --log-level=high
```
BioPhoton, line 25: Left over?

nx.json
```diff
@@ -114,15 +114,22 @@
...
     {
       "plugin": "./tools/src/verdaccio/verdaccio.plugin.ts",
-      "options": { "verbose": true }
+      "options": {
+        "verbose": true
+      }
     }
-  ]
+  ],
+  "useInferencePlugins": false
```
BioPhoton, line 134: What does it do?

packages/nx-plugin/src/plugin/plugin.ts
```diff
@@ -1,11 +1,16 @@
...
+  CreateNodesContext,
+  CreateNodesResult,
+} from '@nx/devkit';
 import { PROJECT_JSON_FILE_NAME } from '../internal/constants';
 import { createTargets } from './target/targets';
 import type { CreateNodesOptions } from './types';
 import { normalizedCreateNodesContext } from './utils';
 
 // name has to be "createNodes" to get picked up by Nx
+// FIXME: refactor this to use the V2 api & remove the eslint disable on the whole file
```
BioPhoton, line 13: I opened #881

BioPhoton (changes requested):
Found small left overs. Merging main will also remove a couple of changed files. Would be good to do that after #873 is in main.

nx.json
vmasek: Leftover from failed migration https://nx.dev/concepts/inferred-tasks

e2e/nx-plugin-e2e/tests/plugin-create-nodes.e2e.test.ts
```diff
@@ -158,6 +158,7 @@ describe('nx-plugin', () => {
         configurations: {},
         executor: `@code-pushup/nx-plugin:cli`,
         options: {},
+        parallelism: true,
```
BioPhoton, line 161: Why this?

vmasek: It seems to be expected there by the unit test result

vmasek:
I am stuck on e2e failing, not really sure how to proceed. The logs in GH actions do not load but you can download them as zip

e2e/nx-plugin-e2e/tests/plugin-create-nodes.e2e.test.ts
vmasek: Nx 19 added is as default flag that signify that target can be run in parallel (I assume we want to keep it `true`)

vmasek:
As there were some issues with ESBuild and replacing some build outputs in dist we decided with @matejchalk to dig deep and replace ESBuild with pure TSC already (as we anyway plan to do it for better error stacks & simpler build).

Most of the codebase is migrated now and we'll be fixing the rest of the issues till the pipeline gets green 🟢 🙏🏻

vmasek:
@matejchalk @BioPhoton this is ready for review

packages/plugin-eslint/src/lib/eslint-plugin.integration.test.ts
```diff
@@ -18,8 +18,8 @@ describe('eslintPlugin', () => {
     ...plugin,
     runner: {
       ...(plugin.runner as RunnerConfig),
-      args: (plugin.runner as RunnerConfig).args?.map(arg =>
-        toUnixPath(arg.replace(thisDir, '<dirname>')),
+      args: (plugin.runner as RunnerConfig).args?.map(() =>
+        toUnixPath('<dirname>/bin.js'),
```
matejchalk, line 22: This change means the `args` effectively aren't being tested. If you only adjust the replaced path (directory is 1 level up now because `..` was added), we can still meaningfully assert that the correct arguments were created.

```suggestion
      args: (plugin.runner as RunnerConfig).args?.map((arg) =>
        toUnixPath(arg.replace(dirname(thisDir), '<dirname>')),
      ),
```

packages/utils/mocks/fixtures/execute-progress.mock.mjs
```diff
@@ -1,5 +1,5 @@
 import { bold, gray } from 'ansis';
-import { getProgressBar } from '../../../../dist/packages/utils/index.js';
+import { getProgressBar } from '../../../../dist/packages/utils/index.js.js';
```
matejchalk, line 2: Double-extension here. The `index.js` can actually be omitted since there's a `package.json`.

```suggestion
import { getProgressBar } from '../../../../dist/packages/utils';
```

packages/cli/package.json
```diff
@@ -47,6 +47,7 @@
     "@code-pushup/utils": "0.56.0",
     "yargs": "^17.7.2",
     "ansis": "^3.3.0",
+    "tslib": "^2.6.2",
```
matejchalk, line 50: Why does `tslib` need to be added as a production dependency everywhere? :raised_eyebrow:

.prettierrc
```diff
@@ -2,11 +2,12 @@
...
   "arrowParens": "avoid",
   "trailingComma": "all",
-  "plugins": ["@trivago/prettier-plugin-sort-imports"],
+  "plugins": ["@miller-svt/prettier-plugin-sort-imports"],
   "importOrder": [
     "^@code-pushup/portal-client$",
     "^@code-pushup/(.*)$",
     "^[./]"
   ],
+  "importOrderParserPlugins": ["typescript", "jsx", "importAttributes"],
```
matejchalk, line 11: Since we don't use the import attributes in the end, wouldn't it be better to switch back to `@trivago/prettier-plugin-sort-imports` instead of the fork?

code-pushup.config.ts
```diff
@@ -35,30 +36,30 @@ const config = {
...
+  //   {
+  //     slug: 'custom-checks',
+  //     title: 'Custom checks',
+  //     refs: [
+  //       ...fileSizeRecommendedRefs,
+  //       packageJsonPerformanceGroupRef,
+  //       packageJsonDocumentationGroupRef,
+  //     ],
+  //   },
+  // ],
```
matejchalk, line 62: I'd be fine with just deleting these from the config, to be honest :grin: I don't think they serve much purpose anymore.

testing/test-nx-utils/src/lib/utils/nx.ts
```diff
@@ -45,14 +46,14 @@ export async function generateWorkspaceAndProject(
...
-    directory: 'libs',
+    directory: join('libs', name),
     tags: 'scope:plugin',
     linter: 'none',
     unitTestRunner: 'none',
     testEnvironment: 'node',
     buildable: false,
     publishable: false,
-    projectNameAndRootFormat: 'derived',
+    projectNameAndRootFormat: 'as-provided',
```
matejchalk, line 56: Nice, no more Nx warnings :raised_hands:

packages/plugin-eslint/vite.config.integration.ts
```diff
@@ -10,6 +10,7 @@ export default defineConfig({
     cache: {
       dir: '../../node_modules/.vitest',
     },
+    testTimeout: 10_000,
```
matejchalk, line 13: Do we know which is the slow test specifically? The timeout could be set for an indivual test instead of all tests at once. 10 seconds isn't so bad though, so no biggie.

matejchalk (commented):
Amazing effort! :clap:

packages/plugin-eslint/vite.config.integration.ts
vmasek: it was just leftover, thanks for pointing it out

code-pushup.config.ts
```diff
@@ -14,7 +14,6 @@ import {
   packageJsonPlugin,
 } from './dist/examples/plugins';
 import { mergeConfigs } from './dist/packages/utils';
-import type { CoreConfig } from './packages/models/src';
```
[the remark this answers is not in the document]
vmasek: I was failing on that import previously but I guess it was just some red herring, reverted back

packages/cli/package.json
vmasek: It seems to be fixed by setting `"importHelpers": false` in `tsconfig.base.json` as we do not need import helpers anyway.

matejchalk (approved):

e2e/cli-e2e/mocks/fixtures/dummy-setup/code-pushup.config.ts
```diff
@@ -1,4 +1,4 @@
-import dummyPlugin, { dummyCategory } from './dummy.plugin';
+import dummyPlugin, { dummyCategory } from './dummy.plugin.js';
```
BioPhoton, line 1: Why is this needed?

e2e/cli-e2e/mocks/fixtures/existing-reports/code-pushup.config.ts
```diff
@@ -1,4 +1,4 @@
-import dummyPlugin, { dummyCategory } from './dummy.plugin';
+import dummyPlugin, { dummyCategory } from './dummy.plugin.js';
```
BioPhoton, line 1: Why is this needed?

examples/plugins/project.json
```diff
@@ -5,14 +5,13 @@
   "projectType": "library",
   "targets": {
     "build": {
-      "executor": "@nx/esbuild:esbuild",
+      "executor": "@nx/js:tsc",
```
BioPhoton, line 8: Why did we switch the build tool form `esbuild` to `tsc`?

vmasek: in general #835 and in examples for consistency and eventually getting rid of esbuild dependency (when we can drop `bundle-require` in future)

BioPhoton: is there another way to use `tsc` than change all imports?

vmasek:
> Do we really need to touch all imports in the code base to migrate to nx v19?
> 
> I would really want to avoid this.

I know it is quite big change but is seems like the best direction we could go. 
We are dropping a bundler and going for pure TS compilation.

BioPhoton (approved):

PR #812, opened 2024-09-04
feat(cli): do not show help on error

vmasek:


matejchalk (approved):

packages/cli/src/lib/yargs-cli.ts
```diff
@@ -71,8 +71,9 @@ export function yargsCli<T = unknown>(
     // take minimum of TERMINAL_WIDTH or full width of the terminal
     .wrap(Math.max(TERMINAL_WIDTH, cli.terminalWidth()))
     .help('help', descriptionStyle('Show help'))
-    .version('version', dim`Show version`, version)
     .alias('h', 'help')
+    .showHelpOnFail(false)
+    .version('version', dim`Show version`, version)
```
BioPhoton, line 76: I would remove version at all as it occupies other args named version.

PR #813, opened 2024-09-11
fix(utils): handle rounding of small trend values

### LINKED ISSUE #804: Prevent displaying ±0 changes in `report-diff.md`
### User story

As a user, it confuses me to see  or  in PR comments - did the score/value change or not?

Example in https://github.com/code-pushup/cli/pull/803#issuecomment-2312464609
- the `report-diff.json` ([download `code-pushup-report-diff` artifact](https://github.com/code-pushup/cli/actions/runs/10578277804/artifacts/1859518332) to view it) has the precise values - _Code coverage_'s `diff` is `-0.00000999999999995449`, _Code style_'s `diff` is `0.00007368110816385887` (scores are in 0-1 range here)
- the `report-diff.md` rounds the diffs to `0` because it only shows max 1 decimal (this also causes inappropriate `+` sign)

### Acceptance criteria

- [ ] adjust rounding logic for `diff` values in `formatScoreChange` function
  - [ ] value should never be displayed as 0 if it isn't exactly 0
  - [ ] values between -0.1 and 0 (after multiplying by 100) should be displayed as `−0.1` 
  - [ ] values between 0 and 0.1 (after multiplying by 100) should be displayed as `+0.1` 
- [ ] adjust rounding logic in `formatValueChange` function
  - [ ] if value change percentage would be rounded to 0, display rounded up to 1 decimal instead
  - [ ] if 1 decimal isn't enough to differentiate from 0, display as `+0.1 %` or `−0.1 %`

### Implementation details

Add unit tests for this rounding logic.

Suggested test cases for scores:
- `diff: -0.4` :arrow_right: `↑ −40`
- `diff: 0.015` :arrow_right: `↑ +0.2`
- `diff: -0.007` :arrow_right: `↑ −0.1`
- `diff: 0.002` :arrow_right: `↑ +0.1`
- `diff: -0.0000001` :arrow_right: `↑ −0.1`

Suggested test cases for values:
- `diff: -150, before: 600` :arrow_right: `↑ −25  %`
- `diff: 1, before: 100` :arrow_right: `↑ +1  %`
- `diff: 1, before: 1000` :arrow_right: `↑ +0.1  %`
- `diff: -1, before: 500` :arrow_right: `↑ −0.2  %`
- `diff: 42, before: 123_456` :arrow_right: `↑ +0.1  %`

hanna-skryl:
Fixes #804 

This PR replaces `Math.round()` with the custom `roundValue` function to handle rounding of score and value changes.

Previously, scores could be floats rounded to one decimal place, while values were always rounded to whole numbers. With the new `roundValue` function, both scores and values are now consistently rounded to their first decimal, ensuring uniform and accurate rounding throughout.


packages/utils/src/lib/reports/utils.ts
```diff
@@ -97,9 +97,21 @@ export function severityMarker(severity: 'info' | 'warning' | 'error'): string {
...
 
+const MIN_NON_ZERO_RESULT = 0.1;
+
+export function roundValue(value: number): number {
+  const roundedValue = Math.round(value * 10) / 10; // round with max 1 decimal
+  if (Math.abs(value) < 1 && value !== 0) {
+    return roundedValue === 0
+      ? MIN_NON_ZERO_RESULT * Math.sign(value)
+      : roundedValue;
+  }
```
matejchalk, line 108: :sewing_needle: Nitpick - I think as single condition would be a bit easier to follow:

```suggestion
  if (Math.abs(value) < 1 && value !== 0 && roundedValue === 0) {
    return MIN_NON_ZERO_RESULT * Math.sign(value);
  }
```

In fact, I'm not sure that checking if it's in -1 to 1 range is necessary:

```suggestion
  if (roundedValue === 0 && value !== 0) {
    return MIN_NON_ZERO_RESULT * Math.sign(value);
  }
```

matejchalk (dismissed):

packages/utils/src/lib/reports/utils.ts
hanna-skryl: You are absolutely right! That check was unnecessary

matejchalk (approved):

PR #814, opened 2024-09-12
fix(plugin-js-packages): filter out warnings

### LINKED ISSUE #787: JS Packages plugin parsing error when warning is printed
### What happened?

My PNPM repository uses an `.npmrc` file for publishing, and for many PNPM commands there is a warning printed regarding a missing environment variable. Although the log doesn't doesn't really affect anything (`pnpm audit` and `pnpm outdated` work as usual), it causes a parsing error in the Code PushUp plugin, which expects pure JSON in stdout.

I would expect that additional unrelated logs don't break the Code PushUp plugin. Since those logs have quite a well-defined format, perhaps we could add some logic to filter out those lines before calling `JSON.parse`?

### What steps did you take?

Described above.

0.48.0

Linux

### Node version

20.16.0

### Relevant log output

```shell
Code PushUp CLI
[ info ] Run autorun...
[ warn ] Plugins failed: 
[ warn ] Error: - Plugin JS Packages (js-packages) produced the following error:
  - undefined:1
 WARN  Issue while reading "/home/runner/work/radar-kit-fu/radar-kit-fu/packages/radar/.npmrc". Failed to replace env in config: ${NODE_AUTH_TOKEN}
^

SyntaxError: Unexpected token ' ', " WARN  Iss"... is not valid JSON
    at JSON.parse (<anonymous>)
    at Object.pnpmToOutdatedResult [as unifyResult] (file:///home/runner/work/radar-kit-fu/radar-kit-fu/node_modules/.pnpm/@code-pushup+js-packages-plugin@0.48.0/node_modules/@code-pushup/js-packages-plugin/bin.js:1509:29)
    at processOutdated (file:///home/runner/work/radar-kit-fu/radar-kit-fu/node_modules/.pnpm/@code-pushup+js-packages-plugin@0.48.0/node_modules/@code-pushup/js-packages-plugin/bin.js:1971:40)
    at async executeRunner (file:///home/runner/work/radar-kit-fu/radar-kit-fu/node_modules/.pnpm/@code-pushup+js-packages-plugin@0.48.0/node_modules/@code-pushup/js-packages-plugin/bin.js:1952:57)
    at async file:///home/runner/work/radar-kit-fu/radar-kit-fu/node_modules/.pnpm/@code-pushup+js-packages-plugin@0.48.0/node_modules/@code-pushup/js-packages-plugin/bin.js:2026:1

Node.js v20.16.0
```

hanna-skryl:
Fixes #787 

- Create a `filterOutWarnings` function to remove warnings from the output before passing it to `JSON.parse`.
- Add unit tests to verify that warnings are filtered out and valid JSON is correctly parsed.

Additionally, the updated functionality was manually tested in a real project using pnpm by directly editing the plugin's source code in the `node_modules` directory and executing the command. The command ran successfully, confirming that the functionality works as expected.

### DIFF
--- a/packages/plugin-js-packages/src/lib/package-managers/pnpm/audit-result.ts
+++ b/packages/plugin-js-packages/src/lib/package-managers/pnpm/audit-result.ts
@@ -1,9 +1,10 @@
 import type { AuditResult, Vulnerability } from '../../runner/audit/types';
 import { getVulnerabilitiesTotal } from '../../runner/audit/utils';
 import type { PnpmAuditResultJson } from './types';
+import { filterWarnings } from './utils';
 
 export function pnpmToAuditResult(output: string): AuditResult {
-  const pnpmResult = JSON.parse(output) as PnpmAuditResultJson;
+  const pnpmResult = JSON.parse(filterWarnings(output)) as PnpmAuditResultJson;
 
   const vulnerabilities = Object.values(pnpmResult.advisories).map(
     ({
--- a/packages/plugin-js-packages/src/lib/package-managers/pnpm/audit-result.unit.test.ts
+++ b/packages/plugin-js-packages/src/lib/package-managers/pnpm/audit-result.unit.test.ts
@@ -76,6 +76,62 @@ describe('pnpmToAuditResult', () => {
       summary: { critical: 0, high: 0, moderate: 0, low: 0, info: 0, total: 0 },
     });
   });
+
+  it('should filter out warnings from the input', () => {
+    /* eslint-disable no-irregular-whitespace */
+    const inputWithWarnings = `
+       WARN  Unsupported engine
+       WARN  Issue while reading file
+      {
+        "advisories": {
+          "123": {
+            "module_name": "@cypress/request",
+            "id": 123,
+            "severity": "high",
+            "vulnerable_versions": "<2.88.12",
+            "recommendation": "Upgrade to version 2.88.12 or later",
+            "title": "SSR forgery",
+            "url": "https://github.com/advisories",
+            "findings": [
+              { "paths": [". > @cypress/request@2.88.5"] }
+            ]
+          }
+        },
+        "metadata": {
+          "vulnerabilities": {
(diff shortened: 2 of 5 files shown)


matejchalk (commented):
I checked out NPM and Yarn in case their JSON output needs to handled similarly. When I create an invalid `.npmrc` (e.g. `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}` when the environment variable doesn't exist), then `npm audit --json` ignores it completely and succesfully outputs pure JSON, while `yarn audit --json` reports it as an error and returns a non-zero exit code.

So my conclusion is that only PNPM needs to handled this way.

vmasek:
I was thinking, wouldn't option like `--silent` be what we need to run these commands with?

https://pnpm.io/cli/dlx#--silent--s

matejchalk (approved):

hanna-skryl:
> I was thinking, wouldn't option like `--silent` be what we need to run these commands with?
> 
> https://pnpm.io/cli/dlx#--silent--s

@vmasek @matejchalk Sorry, guys, for the confusion. I just tried running the `outdated` and `audit` commands with the `--silent` option in a pnpm project, and it does remove warnings from the output.

So, the fix would be to add `--silent` to `COMMON_AUDIT_ARGS` and `COMMON_OUTDATED_ARGS`.

hanna-skryl:
> > I was thinking, wouldn't option like `--silent` be what we need to run these commands with?
> > https://pnpm.io/cli/dlx#--silent--s
> 
> @vmasek @matejchalk Sorry, guys, for the confusion. I just tried running the `outdated` and `audit` commands with the `--silent` option in a pnpm project, and it does remove warnings from the output.
> 
> So, the fix would be to add `--silent` to `COMMON_AUDIT_ARGS` and `COMMON_OUTDATED_ARGS`.

UPD: `--silent` filters out warnings, but for some reason, not all of them. So, let's stick to the current solution and keep an eye on updates on the `pnpm` end in case they fix this behavior.

PR #818, opened 2024-09-18
feat(cli): add command option aliases

### LINKED ISSUE #672: Add aliases to command arguments
### User story

As a developer I want to have descriptive command arguments for readability and shorthands for productivity. ATM there are no short aliases present

### Acceptance criteria

- [ ] `--onlyPlugins` => `-p`

### Implementation details

_No response_

hanna-skryl:
Closes #672 

This PR adds aliases for `--onlyPlugins` and `--skipPlugins`.

### DIFF
--- a/packages/cli/src/lib/implementation/skip-plugins.options.ts
+++ b/packages/cli/src/lib/implementation/skip-plugins.options.ts
@@ -7,6 +7,7 @@ export const skipPluginsOption: Options = {
   type: 'array',
   default: [],
   coerce: coerceArray,
+  alias: 'P',
 };
 
 export function yargsSkipPluginsOptionsDefinition(): Record<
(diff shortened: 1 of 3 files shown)


hanna-skryl:
Using `-c` as an alias for `--config` is tricky. `-c` conflicts with Nx options and such options need to be wrapped into `--args="..."` to work (see more on this topic [here](https://nx.dev/recipes/running-tasks/pass-args-to-commands#pass-args-when-running-the-command-in-the-terminal)).

If anyone knows how to do that, I would appreciate any help!

vmasek (approved):

packages/cli/src/lib/implementation/skip-plugins.options.ts
```diff
@@ -7,6 +7,7 @@ export const skipPluginsOption: Options = {
   type: 'array',
   default: [],
   coerce: coerceArray,
+  alias: 'P',
```
BioPhoton, line 10: could we get a unit test for this? should be doable in the middleware tests.

hanna-skryl: There were no issues with adding the suggested `config` alias (`-c`). The problem was that it ended up being ignored during the command execution as `-c` conflicted with the Nx CLI option.

As discussed, let's keep the `config` option without an alias for now.

PR #820, opened 2024-09-19
refactor(cli): improve plugin filter option validation

### LINKED ISSUE #565: If `--onlyPlugins` filters out all plugins an empty result shows without a warning
### What happened?

I ran `code-pushup collec --onlyPlugins non-existing-plugin` and an empty report appeared in the terminal

A warning ahould show up in the terminal

### What steps did you take?

Run `code-pushup collec --onlyPlugins non-existing-plugin` 

**Output:**
Screenshot 2024-03-14 at 08 42 06

_No response_

MacOS

### Node version

_No response_

### Relevant log output

_No response_

hanna-skryl:
Fixes #565 

- Ensure warnings for invalid plugin filter options are logged regardless of the `--verbose` flag.
- Improve message clarity by handling both singular and plural cases.
- Update unit tests to reflect changes in logging behavior and message wording.

### DIFF
--- a/packages/cli/src/lib/implementation/validate-plugin-filter-options.utils.ts
+++ b/packages/cli/src/lib/implementation/validate-plugin-filter-options.utils.ts
@@ -1,4 +1,3 @@
-import { yellow } from 'ansis';
 import type { CategoryConfig, PluginConfig } from '@code-pushup/models';
 import { filterItemRefsBy, ui } from '@code-pushup/utils';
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/lib/implementation/validate-plugin-filter-options.utils.ts
hanna-skryl, line 1: ❓ Should we abort execution on invalid plugin filters?

Currently, if an invalid value is passed to the `onlyPlugins` or `skipPlugins` filter options, a warning is logged, but all plugins still run as if no filter was applied. This could confuse users, who might expect the process to stop when they provide an invalid filter.

Would it make sense to abort execution when an invalid filter option is detected so users can correct the input before proceeding?

Curious to hear your thoughts on this.

vmasek (approved):

packages/cli/src/lib/implementation/validate-plugin-filter-options.utils.ts
hanna-skryl: It was agreed during the discussion that the command execution should be aborted when none of the values provided to the `onlyPlugins` filter option is valid. It's OOS for this PR, will be addressed later.

PR #824, opened 2024-09-25
feat(utils): support local report links

### LINKED ISSUE #745: Issues sections should link to local file in your IDE at given commit with highlighted line(s)
### User story

As a user of the CLI I want to have click through experience for all actionable feedback a get from the audit issues.

At the moment the provided issues often show exactly where the error needs to get fixed, but a manual navigation to the line of code is required.

**Example:**

|Severity|Message|Source file|Line(s)|
|:--:|:--:|:--:|:--:|
|⚠️ <i>warning</i>|Lines 1-18 are not covered in any test case.|<code>src/lib/index.ts</code>|1-18|
|⚠️ <i>warning</i>|Line 104 is not covered in any test case.|<code>src/lib/index.ts</code>|104|

|Severity|Message|Source file|Line(s)|
|:--:|:--:|:--:|:--:|
|🚨 <i>error</i>|1st branch is not taken in any test case.|<code>src/lib/index.ts</code>|101|
|🚨 <i>error</i>|1st branch is not taken in any test case.|<code>src/lib/index.ts</code>|1|

### Acceptance criteria

- [ ] The report contains links to the IDE
- [ ] When the link is clicked it opens the source file in the IDE
   - Webstorm - only file is linkable over relative path (`../src/lib/index.ts`)
   - Code - file and line are linkable over relative path (`../src/lib/index.ts#L9`)
   - GitHub - file and line rang is linkable over a relative path (`../src/lib/index.ts#L4-L8`)
- [ ] When no location of the report is present (location of report file unknown) the static file path is displayed. No link is used.

**Folder:**  
```
/
├── .code-pushup
│   └── report.json
└── src
    └── lib
        └── index.ts

```

**Link in report:**  
`[src/lib/index.ts](../src/lib/index.ts)`

**Example Report:**

|Severity|Message|Source file|Line(s)|
|:--:|:--:|:--:|:--:|
|⚠️ <i>warning</i>|Lines 1-18 are not covered in any test case.| <a href="../src/lib/index.ts#L1-L18"><code>src/lib/index.ts</code></a> |1-18|
|⚠️ <i>warning</i>|Line 104 is not covered in any test case.| <a href="src/lib/index.ts#L104"><code>src/lib/index.ts</code></a> |104 |

| Severity|Message|Source file|Line(s) |
|:--:|:--:|:--:|:--:|
|🚨 <i>error</i>|1st branch is not taken in any test case.|<a href="../src/lib/index.ts#L101"><code>src/lib/index.ts</code></a>|101|
|🚨 <i>error</i>|1st branch is not taken in any test case.|<a href="../src/lib/index.ts#L1"><code>src/lib/index.ts</code></a>|1|

### Implementation details

Essential logic and tests are present in the following PR, the over all configuration handling is not fully thought through.

**https://github.com/code-pushup/cli/tree/add-link-to-local-filesystem-in-reports** 

**Related Issues:**

https://github.com/code-pushup/cli/issues/149

hanna-skryl:
This PR introduces support for generating links from markdown reports to local source files. The links are formatted based on the detected environment (VS Code, GitHub, etc.), allowing direct navigation to specific lines within the IDE.

Closes #745


packages/utils/src/lib/reports/generate-md-report.unit.test.ts
```diff
@@ -135,31 +135,39 @@ describe('auditDetailsIssues', () => {
...
-        } as Issue,
-      ])?.toString(),
-    ).toMatch(/\|\s*4\s*\|/);
+            severity: 'warning',
+          } as Issue,
+        ],
+        { outputDir: '/.code-pushup' },
+      )?.toString(),
+    ).toMatch('[`/src/index.js`](../src/index.js#L4)');
+    vi.unstubAllEnvs();
```
matejchalk, line 162: This won't be reached if the `expect` assertion fails. Use an `afterAll`/`afterEach` hook to make the cleanup happens regardless of test result.

The same problem is present in `packages/utils/src/lib/reports/formatting.unit.test.ts`, so it might be best to add this hook to [`testing/test-setup/src/lib/reset.mocks.ts`](https://github.com/code-pushup/cli/blob/main/testing/test-setup/src/lib/reset.mocks.ts), it seems reasonable all include this for all tests automatically.

packages/utils/src/lib/reports/generate-md-report.unit.test.ts
```diff
@@ -135,31 +135,39 @@ describe('auditDetailsIssues', () => {
   it('should include source file', () => {
     expect(
       auditDetailsIssues([
-        { source: { file: 'index.js' }, severity: 'error' } as Issue,
+        { source: { file: '/index.js' }, severity: 'error' } as Issue,
       ])?.toString(),
-    ).toMatch('`index.js`');
+    ).toMatch('`/index.js`');
```
matejchalk, line 140: All issue file paths will be relative to git root directory at this point, so won't start with `/`. The core [converts all issues source files](https://github.com/code-pushup/cli/blob/main/packages/core/src/lib/normalize.ts#L13) right after [plugin execution](https://github.com/code-pushup/cli/blob/main/packages/core/src/lib/implementation/execute-plugin.ts#L85).

So realistic examples would be `index.js` or `src/index.js`.

packages/utils/src/lib/reports/formatting.unit.test.ts
```diff
@@ -112,3 +118,108 @@ describe('metaDescription', () => {
...
+  ])('should format position %o as "%s"', (position, expected) => {
+    expect(
+      formatSourceLine({ file: '/packages/utils/src/index.ts', position }),
+    ).toBe(expected);
+  });
+
+  it('should return an empty string when position is missing', () => {
+    expect(formatSourceLine({ file: '/packages/utils/src/index.ts' })).toBe('');
+  });
+});
```
matejchalk, line 137: Wouldn't it make more sense for the `formatSourceLine` to receive only the position? The source object includes irrelevant properties like `file`.

packages/utils/src/lib/reports/formatting.ts
```diff
@@ -58,3 +66,58 @@ export function metaDescription(
...
+  }
+  const ide = ideEnvironment();
+  const { startLine, endLine } = position;
+  return ide === 'vscode'
+    ? `${file}#L${startLine}`
+    : ide === 'github'
+    ? endLine
+      ? `${file}#L${startLine}-L${endLine}`
+      : `${file}#L${startLine}`
+    : file;
```
matejchalk, line 122: I think refactoring this using `switch (ide)` would be more readable.

packages/utils/src/lib/reports/formatting.ts
```diff
@@ -58,3 +66,58 @@ export function metaDescription(
...
+  const { file, position } = source;
+  const { outputDir } = options ?? {};
+  const unixPath = toUnixPath(file);
+
+  // NOT linkable
+  if (!outputDir) {
+    return md.code(unixPath);
+  }
+
+  const relativePath = pathPosix.relative(outputDir, unixPath);
```
matejchalk, line 91: Why use `posix.relative` instead of `relative`?

packages/utils/src/lib/reports/formatting.ts
```diff
@@ -58,3 +66,58 @@ export function metaDescription(
...
+ * @example
+ * linkToLocalSourceInIde({ file: '/src/index.ts'}, {outputDir: '/.code-pushup'}) // [/src/index.ts](../src/index.ts)
+ */
+export function linkToLocalSourceForIde(
+  source: SourceFileLocation,
+  options?: Pick<MdReportOptions, 'outputDir'>,
+): InlineText {
+  const { file, position } = source;
+  const { outputDir } = options ?? {};
+  const unixPath = toUnixPath(file);
```
matejchalk, line 84: It will already be in Unix format, the [`formatGitPath`](https://github.com/code-pushup/cli/blob/main/packages/utils/src/lib/git/git.ts#L10-L14) function takes care of that.

packages/utils/src/lib/reports/formatting.ts
```diff
@@ -58,3 +66,58 @@ export function metaDescription(
...
+  file: string,
+  position?: SourceFileLocation['position'],
+): string {
+  if (!position) {
+    return file;
+  }
+  const ide = ideEnvironment();
+  const { startLine, endLine } = position;
+  return ide === 'vscode'
+    ? `${file}#L${startLine}`
```
matejchalk, line 117: Tested by running Code PushUp CLI in VSCode terminal - the links are created in Markdown and open the correct file and line :tada:

packages/utils/src/lib/reports/formatting.ts
```diff
@@ -58,3 +66,58 @@ export function metaDescription(
...
+    : ide === 'github'
+    ? endLine
+      ? `${file}#L${startLine}-L${endLine}`
+      : `${file}#L${startLine}`
```
matejchalk, line 121: I don't understand how this is intended to work for GitHub Actions. The `report.md` can be downloaded as an artifact, but we don't have any control over where they'll open it, so the relative links are likely to be broken.

It would make more sense to link to remote URL in this case. Since we already know we're running in a GitHub Actions environment, we can use its [built-in environment variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables) to reliably construct the correct GitHub URL. Something like:

```ts
`${process.env.GITHUB_SERVER_URL}/${process.env.GITHUB_REPOSITORY}/blob/${process.env.GITHUB_SHA}/${issue.source.file}`
````

**Review by matejchalk — CHANGES_REQUESTED** [review]

File: packages/utils/src/lib/reports/formatting.ts (hunk 5 above)
Comment (hanna-skryl) [reply packages/utils/src/lib/reports/formatting.ts:91]: I tried out several other configurations, and none of those using regular `path.relative()` could normalize Windows path correctly

File: packages/utils/src/lib/reports/formatting.ts (hunk 6 above)
Comment (hanna-skryl) [reply packages/utils/src/lib/reports/formatting.ts:84]: The Windows tests kept failing without this conversion (see [here](https://github.com/code-pushup/cli/actions/runs/11040734885/job/30669397342)). Do you think the test itself was incorrect and it wasn't the code problem?

File: packages/utils/src/lib/reports/formatting.ts (hunk 5 above)
Comment (matejchalk) [reply packages/utils/src/lib/reports/formatting.ts:91]: Oh, so it's to prevent conversion to Windows paths? OK, then POSIX makes sense.

File: packages/utils/src/lib/reports/formatting.ts (hunk 6 above)
Comment (matejchalk) [reply packages/utils/src/lib/reports/formatting.ts:84]: Oh. Well, then I guess the tests input was in Windows format? You could always call `toUnixPath` in the test to prepare the input if need be.

Comment (hanna-skryl) [reply packages/utils/src/lib/reports/formatting.ts:84]: The `linkToLocalSourceForIde` test input uses forward slashes, but when running tests for Windows, it gets internally changed to the Windows-style path format. As a result, when tested in isolation, it didn't produce the expected results.

I am not very familiar with the code base, so I added `toUnixPath` as a precaution. If you confirm that it will work as expected without `toUnixPath`, then I will add it to the test instead.

Comment (matejchalk) [reply packages/utils/src/lib/reports/formatting.ts:84]: Yes, you can assume that `issue.source.file` was been normalized to relative Git path.

File: packages/utils/src/lib/reports/formatting.unit.test.ts [hunk 9]
```diff
@@ -126,100 +128,161 @@ describe('formatSourceLine', () => {
...
       }).toString(),
     ).toBe('`/packages/utils/src/index.ts`');
   });
 
-  it('should format file path as link when outputDir is provided', () => {
+  it('should format the file path as a link when outputDir is defined (VS Code)', () => {
+    vi.stubEnv('TERM_PROGRAM', 'vscode');
+    vi.stubEnv('GITHUB_ACTIONS', 'false');
+    const filePath = toUnixPath('/packages/utils/src/index.ts');
+    const outputDir = toUnixPath('/.code-pushup');
```
matejchalk, line 155: Please drop the leading slash, so that the test inputs use the expected format:

```suggestion
    const filePath = toUnixPath('packages/utils/src/index.ts');
    const outputDir = toUnixPath('.code-pushup');
```

Applies to multiple places in this file, and to `packages/utils/src/lib/reports/generate-md-report.unit.test.ts` as well.

packages/utils/src/lib/reports/formatting.unit.test.ts
```diff
@@ -126,100 +128,161 @@ describe('formatSourceLine', () => {
...
-    (file, position, expected) => {
-      vi.stubEnv('GITHUB_ACTIONS', 'true');
-      vi.stubEnv('TERM_PROGRAM', '');
-      expect(formatFilePosition(file, position)).toBe(expected);
-      vi.unstubAllEnvs();
+    'should generate a GitHub repository link for the file with position %o',
+    (position, expected) => {
+      expect(
+        getGitHubLink(
+          toUnixPath('src/index.ts'),
```
matejchalk, line 206: The URL is incorrect. The issue's file path is relative to the root git directory, not the report output directory, so it should be `https://github.com/user/repo/blob/1234567890abcdef/src/index.ts#L2`. The `outputDir` is actually irrelevant for the GitHub URL.

packages/utils/src/lib/reports/formatting.unit.test.ts
```diff
@@ -126,100 +128,161 @@ describe('formatSourceLine', () => {
...
+      { startLine: 2 },
+      'https://github.com/user/repo/blob/1234567890abcdef/.code-pushup/src/index.ts#L2',
+    ],
+    [
+      { startLine: 2, endLine: 5 },
+      'https://github.com/user/repo/blob/1234567890abcdef/.code-pushup/src/index.ts#L2-L5',
+    ],
+    [
+      { startLine: 2, startColumn: 1 },
+      'https://github.com/user/repo/blob/1234567890abcdef/.code-pushup/src/index.ts#L2',
```
matejchalk, line 199: I actually just found out that GitHub URLs support columns as well as line numbers :open_mouth: E.g.: https://github.com/code-pushup/cli/blob/main/packages/utils/src/lib/merge-configs.ts#L38C60-L40C10

Do you think you could include the columns, too? :pray:

packages/utils/src/lib/reports/formatting.unit.test.ts
hanna-skryl: Makes total sense! Don't know why I assumed `outputDir` was part of the GitHub link

packages/utils/src/lib/reports/formatting.unit.test.ts
hanna-skryl: Done ✅

packages/utils/src/lib/reports/formatting.ts
```diff
@@ -103,21 +103,24 @@ export function formatSourceLine(
...
-  const { startLine, endLine } = position;
-  const line = endLine ? `#L${startLine}-L${endLine}` : `#L${startLine}`;
-  return `${baseUrl}/${fullPath}${line}`;
+  const start = startColumn ? `L${startLine}C${startColumn}` : `L${startLine}`;
+  const end = endLine
+    ? endColumn
+      ? `L${endLine}C${endColumn}`
+      : `L${endLine}`
+    : '';
+  const lineRange = end ? (start === end ? start : `${start}-${end}`) : start;
```
matejchalk, line 122: (Nitpick) This would be a little easier to follow:

```suggestion
  const lineRange = end && start !== end ? `${start}-${end}` : start;
```

matejchalk (dismissed):

vmasek (dismissed):

matejchalk (approved):

PR #825, opened 2024-09-26
fix(utils): format reports with unchanged score

### LINKED ISSUE #819: Prevent empty categories table in PR comments
### What happened?

When only audits change, but categories, an empty categories table is displayed (only column headers).

image

If all categories are unchanged, then the table shouldn't be shown, only the _All of 4 categories are unchanged_ message should be displayed

### What steps did you take?

Running Code PushUp GitHub Action in monorepo mode. Fixed a few lint errors, but not all, so there is no change to categories.

0.50.0

Linux

### Node version

20

### Relevant log output

_No response_

hanna-skryl:
This PR resolves an issue with the report formatting logic where changes that did not affect the score were not being handled correctly. Specifically, it ensures that a categories table is excluded from the monorepo report when none of the category scores have been affected.

Closes #819


BioPhoton (approved):
Nice fix! Reduced the irrelevant information.

vmasek (approved):

matejchalk (approved):

PR #831, opened 2024-10-03
feat(cli): add skip and only category filters

### LINKED ISSUE #610: add CLI flag `--skipCategories`
### User story

As a developer sometimes I want to skip the execution of certain categories. E.g. to save time.

ATM there is no way to do this but comment them out in the `code-pushup.config.ts`

### Acceptance criteria

- [ ] the CLI supports a terminal argument `skipCategories` that is. used as string array that is `undefined` by default
- [ ] if `skipCategories` is given exclude the provided categories form the core config
- [ ] notify the user similar to `onlyPlugins`

### Implementation details

_No response_

### LINKED ISSUE #611: add CLI flag `--onlyCategories`
### User story

As a developer sometimes I want to skip the execution of certain categories. E.g. to save time.

ATM there is no way to do this but comment them out in the `code-pushup.config.ts`

### Acceptance criteria

- [ ] the CLI supports a terminal argument `onlyCategories` that is. used as string array that is `undefined` by default
- [ ] if `onlyCategories` is given execute only the provided categories form the core config
- [ ] notify the user similar to `onlyPlugins`

### Implementation details

_No response_

hanna-skryl:
Closes #610 closes #611 

This pull request adds two new CLI flags, `--skipCategories` and `--onlyCategories`, to extend the filtering capabilities of the CLI, allowing users to skip or execute only specific categories.

The existing `filterPluginsMiddleware` has been renamed to `filterMiddleware` as it now handles filtering for both plugins and categories in a unified manner. This ensures that all validation and filtering logic happens in the same place.

The user is notified about skipped or included categories in the same way as they are notified about skipped or included plugins, maintaining a consistent user experience across both filtering options.


packages/cli/src/lib/implementation/filter.model.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+export type FilterOptions = Partial<GlobalOptions> &
+  Pick<CoreConfig, 'categories' | 'plugins'> &
+  FilterCliOptions;
+
+export type FilterOptionType =
+  | 'skipCategories'
+  | 'onlyCategories'
+  | 'skipPlugins'
+  | 'onlyPlugins';
```
matejchalk, line 22: `FilterCliOptions` and `FilterOptionType` could derived from each other, makes it easier to keep in sync (you'd only need to change or add an option in one place). Either direction would work:

`FilterCliOptions` :arrow_right: `FilterOptionType`:

```ts
export type FilterCliOptions = {
  skipCategories?: string[];
  onlyCategories?: string[];
  skipPlugins?: string[];
  onlyPlugins?: string[];
};

export type FilterOptionType = keyof FilterCliOptions;
```

`FilterOptionType` :arrow_right: `FilterCliOptions`:

```ts
export type FilterCliOptions = Partial<Record<FilterOptionType, string[]>>;

export type FilterOptionType =
  | 'skipCategories'
  | 'onlyCategories'
  | 'skipPlugins'
  | 'onlyPlugins';
```

matejchalk (dismissed):
I like that you generalized the `{only|skip}{Plugins|Categories}` options into a single implementation, nice work :+1:

matejchalk (approved):

Issue #829, opened 2024-10-03
Shared CI code between GitHub and GitLab


matejchalk:
The `code-pushup/github-action` code has grown in complexity since monorepo mode was added.
Since we want to have very similar functionality for GitLab also (see https://github.com/code-pushup/github-action/issues/58), it would be much more maintainable to make the core logic CI-agnostic and extract it to a shared NPM package, so that both the GitHub Action and the GitLab template can import it and build on top of it.

The proposal is to create a dedicated `@code-pushup/ci` package for this purpose, which would be distributed from `code-pushup/cli` repo.

## Acceptance criteria

- [x] new `@code-pushup/ci` package is created in CLI monorepo
- [x] create abstraction layer from source code in `code-pushup/github-action` which accepts configuration for CI provider
- [x] publish `@code-pushup/ci` package and import it in `code-pushup/github-action`
  - [x] prerequisite: code-pushup/github-action#69
- [x] test `code-pushup/github-action` still works and publish new version to marketplace

PR #839, opened 2024-10-14
feat(ci): new package with code extracted from github-action

matejchalk:
Bulk of #829 

See new package's [README](https://github.com/code-pushup/cli/tree/ci-package/packages/ci).


e2e/ci-e2e/mocks/fixtures/code-pushup.config.ts
```diff
@@ -0,0 +1,49 @@
...
+        const jsFileCount = jsPaths.length;
+        const tsFileCount = tsPaths.length;
+        const ratio = tsFileCount / (jsFileCount + tsFileCount);
+        const percentage = Math.round(ratio * 100);
+        return [
+          {
+            slug: 'ts-files',
+            value: percentage,
+            score: ratio,
+            displayValue: `${percentage}% converted`,
```
BioPhoton, line 34: Nice test data

e2e/ci-e2e/tests/ci.e2e.test.ts
```diff
@@ -0,0 +1,180 @@
...
+} from '@code-pushup/ci';
+
+describe('CI package', () => {
+  const fixturesDir = join(
+    fileURLToPath(dirname(import.meta.url)),
+    '..',
+    'mocks',
+    'fixtures',
+  );
+  const workDir = join(process.cwd(), 'tmp', 'ci-e2e', 'test-repo');
```
BioPhoton, line 33: For test folder we follow the pattern: `tmp/e2e/<project-identifier>/<test-case>`

I suggest:
1. `tmp/e2e/ci-e2e/<test-case>` for a setup per it block
2. `tmp/e2e/ci-e2e/test-repo/<test-case>` for a setup where it blocks depend on `test-repo` but not modify it
3. `tmp/e2e/ci-e2e/<test-case>-test-repo` where the test repo folder gets modified in any way

I suggest 3.

e2e/ci-e2e/tests/ci.e2e.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    vi.spyOn(git, 'fetch').mockResolvedValue({} as FetchResult);
+    vi.spyOn(git, 'diffSummary').mockResolvedValue({
+      files: [{ file: 'index.ts', binary: false }],
+    } as DiffResult);
+    vi.spyOn(git, 'diff').mockResolvedValue('');
+
+    await git.init();
+    await git.addConfig('user.name', 'John Doe');
+    await git.addConfig('user.email', 'john.doe@example.com');
+    await git.branch(['-M', 'main']);
```
BioPhoton, line 62: We have helper functions to setup git repos. 
Have a look at: [testing/test-utils/src/lib/utils/git.ts](https://github.com/code-pushup/cli/blob/main/testing/test-utils/src/lib/utils/git.ts)

e2e/ci-e2e/vite.config.e2e.ts
```diff
@@ -0,0 +1,22 @@
...
+    globals: true,
+    alias: tsconfigPathAliases(),
+    pool: 'threads',
+    poolOptions: { threads: { singleThread: true } },
+    cache: {
+      dir: '../../node_modules/.vitest',
+    },
+    environment: 'node',
+    include: ['tests/**/*.e2e.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
+    globalSetup: ['../../global-setup.e2e.ts'],
```
BioPhoton, line 19: we use `../../global-setup.verdaccio.ts` for new every test setup

e2e/cli-e2e/project.json
```diff
@@ -23,7 +23,6 @@
     "utils",
     "core",
     "cli",
-    "cli",
     "plugin-eslint",
     "plugin-coverage",
     "plugin-js-packages",
```
BioPhoton, line 28: the new `../../global-setup.verdaccio.ts` will derive the dependencies. so you just need to add what you directly test. It's deps are installed automatically from the graph

package.json
```diff
@@ -32,6 +32,7 @@
     "cli-table3": "^0.6.3",
     "esbuild": "^0.19.12",
     "eslint": "~8.56.0",
+    "glob": "^10.4.5",
```
BioPhoton, line 35: We have a benchmark for glob libs here [packages/utils/perf/glob](https://github.com/code-pushup/cli/tree/main/packages/utils/perf/glob).

Lets try to use the fastest and the one with least deps

package.json
```diff
@@ -97,7 +100,6 @@
     "eslint-plugin-unicorn": "^48.0.1",
     "eslint-plugin-vitest": "^0.3.8",
     "fast-glob": "^3.3.2",
-    "glob": "^10.3.10",
     "globby": "^14.0.1",
```
BioPhoton, line 103: why do we need 2 glob libs?

packages/ci/src/lib/cli/commands/collect.ts
```diff
@@ -0,0 +1,29 @@
+import { executeProcess } from '@code-pushup/utils';
+import type { CommandContext } from '../context';
+import {
+  type PersistedCliFiles,
+  persistCliOptions,
+  persistedCliFiles,
+} from '../persist';
+
+export async function collect({
```
BioPhoton, line 9: maybe `runCollect` to avoid confusion with the `collect` function in `core` and the `collect` command in `cli`?

packages/ci/src/lib/cli/persist.ts
```diff
@@ -0,0 +1,102 @@
...
+  return {
+    ...filePaths,
+    artifactData: {
+      rootDir,
+      files: Object.values(filePaths),
+    },
+  };
+}
+
+export function projectToFilename(project: string): string {
```
BioPhoton, line 92: could this life in `utils`?

packages/ci/src/lib/comment.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  const createdComment = await api.createComment(body);
+  logger.info(`Created new comment ${createdComment.url}`);
+  return createdComment.id;
+}
+
+function truncateBody(body: string, max: number, logger: Logger): string {
+  const truncateWarning = '...*[Comment body truncated]*';
+  if (body.length > max) {
```
BioPhoton, line 42: This would be good to unit test

packages/ci/src/lib/monorepo/list-projects.ts
```diff
@@ -0,0 +1,121 @@
...
+    task: settings.task,
+    cwd: settings.directory,
+    ...(!settings.silent && {
+      observer: {
+        onStdout: stdout => {
+          console.info(stdout);
+        },
+        onStderr: stderr => {
+          console.warn(stderr);
+        },
```
BioPhoton, line 71: This configuration could be a new option `verbose` and we would still be able to add custom logs outside.

BioPhoton (changes requested):
Gitlab integration ahead it seems! Nice you started this PR.
More unit tests and some integration tests would round it up. :)

e2e/cli-e2e/project.json
matejchalk: I wish that were the case, but if I omit indirect dependencies they're not published, it seems :disappointed: 

image

matejchalk: Ah, nevermind, started working after I moved the working directory.

matejchalk: I've decided to change `ci-e2e` `project.json` to limit the scope of this PR. There are redundant implicit dependencies all in the E2E projects, but I'd clean that up in some follow up PR.

package.json
matejchalk: I did try `fast-glob`. But for some reason, it wasn't playing nicely with `memfs`, so it was blocking me from unit testing the monorepo handlers properly :disappointed: I couldn't figure out why, in the GitHub Action codebase I was able to test `fast-glob` using `memfs` without any problems (maybe Jest vs Vitest difference). But I spent a few hours trying get it working here without success, then I tried switching to `glob` as the next best option, and at it all started working :man_shrugging: 

I know that `fast-glob` is faster than `glob` in the benchmark, but performance isn't the only factor. The potential milliseconds of runtime difference is outweighted by the hours saved on development time, in my opinion :nerd_face: Otherwise I don't see much difference between them - both have relatively few dependencies and are widely used.

package.json
matejchalk: Explained why I went with `glob` over `fast-glob` in https://github.com/code-pushup/cli/pull/839#discussion_r1800915711

The other glob libs were here as dev dependencies because of performance benchmarks (not sure it makes sense to have them in main branch for that reason only, though). None of them were being used in any of our packages until now, so I moved `glob` to dependencies now that there's a production use case.

package.json
vmasek: I'd propose replacing the blob benchmark example with some synthetic example of some non-trivial math function like Fibonacci or similar. Meaning we could get rid of these unnecessary multiple blob libs deps (could be part of #776 effort)

packages/ci/src/lib/monorepo/list-projects.ts
matejchalk: Good point, although I'm wondering if it shouldn't be on by default :thinking: Instead of _opting in_ to printing the logs with `verbose` (default is off by convention), we would _opt out_ of those logs using a `silent` flag (inverted logic).

The GitHub Actions toolkit exports its own versions of our `executeProcess` ([`exec`](https://github.com/actions/toolkit/blob/main/packages/exec/src/exec.ts#L7-L31) and [`getExecOutput`](https://github.com/actions/toolkit/blob/main/packages/exec/src/exec.ts#L33-L90)), where their [`silent` flag is `false` by default](https://github.com/actions/toolkit/blob/main/packages/exec/src/interfaces.ts#L12-L13).

But this is maybe beyond the scope of this PR.

packages/ci/src/lib/cli/persist.ts
matejchalk: Moved it and added unit tests.

packages/ci/src/lib/cli/commands/collect.ts
matejchalk: Renamed all these command helpers.

e2e/ci-e2e/tests/ci.e2e.test.ts
matejchalk: Used the shared utils and refactored them a little.

package.json
matejchalk: Agreed, let's make that part of the other issue.

packages/ci/src/lib/comment.ts
matejchalk: Done :heavy_check_mark: Unit tested as part of `commentOnPR`.

packages/ci/src/lib/monorepo/list-projects.ts
matejchalk: Moved into new issue:
- #843

package.json
matejchalk: Created a new issue for it:
- #845

BioPhoton (approved):
🤩🚀🚀

vmasek (approved):

PR #868, opened 2024-11-07
refactor: standardize undefined handling for categories

### LINKED ISSUE #518: Resolve categories inconsistencies...again
## Current behaviour
Currently, the categories become an empty array if not passed inside core config middleware. TypeScript complains about this inconsistency later on because in `CoreConfig` type, they are always optional.

## Expected behaviour
Either the CLI should handle them as `undefined` and skip the section in the report instead.
Or there should be two? types, one for the config provided by user and one for the refined config after middlewares.

## Implementation details
Remove the default value from both middlewares (core config and only plugins).

hanna-skryl:
Closes #518 

This refactor standardizes the handling of categories as undefined throughout the CLI pipeline until conversion to the GQL report. Previously, an empty array was assigned by default in middlewares, leading to inconsistencies with `CoreConfig` where `categories` are optional.

By removing the default empty array, `categories` remain undefined unless explicitly provided. This update improves type consistency, ensuring that categories are only included in the final report if defined. When no categories are provided, the JSON report omits the categories field, and the Markdown report summary table indicates that there are 0 categories.

### DIFF
--- a/packages/models/src/lib/implementation/utils.ts
+++ b/packages/models/src/lib/implementation/utils.ts
@@ -64,10 +64,10 @@ export function exists<T>(value: T): value is NonNullable<T> {
  * @returns Array of missing references.
  */
 export function getMissingRefsForCategories(
-  categories: CategoryConfig[],
   plugins: PluginConfig[] | PluginReport[],
+  categories?: CategoryConfig[],
 ) {
-  if (categories.length === 0) {
+  if (!categories || categories.length === 0) {
     return false;
   }
 
@@ -108,10 +108,10 @@ export function getMissingRefsForCategories(
 }
 
 export function missingRefsForCategoriesErrorMsg(
-  categories: CategoryConfig[],
   plugins: PluginConfig[] | PluginReport[],
+  categories?: CategoryConfig[],
 ) {
-  const missingRefs = getMissingRefsForCategories(categories, plugins);
+  const missingRefs = getMissingRefsForCategories(plugins, categories);
   return `The following category references need to point to an audit or group: ${errorItems(
     missingRefs,
   )}`;
--- a/packages/utils/src/lib/reports/generate-md-report.ts
+++ b/packages/utils/src/lib/reports/generate-md-report.ts
@@ -13,7 +13,11 @@ import {
   categoriesDetailsSection,
   categoriesOverviewSection,
 } from './generate-md-report-categoy-section';
-import type { MdReportOptions, ScoredReport } from './types';
+import type {
+  MdReportOptions,
+  ScoredCategoryConfig,
+  ScoredReport,
+} from './types';
 import { formatReportScore, scoreMarker, severityMarker } from './utils';
 
 export function auditDetailsAuditValue({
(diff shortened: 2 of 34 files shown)


packages/ci/src/lib/issues.unit.test.ts
```diff
@@ -271,4 +275,79 @@ describe('issues sorting', () => {
       ),
     ).toBe(0.09); // 1% + 8% = 9%
   });
+
+  it('should return 0 when there are no categories', () => {
```
vmasek, line 279: thanks for this unit test :shipit:

vmasek (approved):

packages/utils/src/lib/reports/generate-md-report.ts
```diff
@@ -26,19 +30,25 @@ export function auditDetailsAuditValue({
...
     .heading(HIERARCHY.level_1, REPORT_HEADLINE_TEXT)
-    .$if(report.categories.length > 0, doc =>
-      doc.$concat(
-        categoriesOverviewSection(report),
-        categoriesDetailsSection(report),
-      ),
+    .$concat(
+      ...(hasCategories(report)
+        ? [categoriesOverviewSection(report), categoriesDetailsSection(report)]
+        : []),
```
hanna-skryl, line 48: Unfortunately, I had to refactor the `build-md`'s `$if` out as it only accepted a boolean condition, so it did not narrow types in the body. As a result, `categories` were still interpreted as potentially undefined inside the `$if` block, leading to TypeScript errors.

packages/ci/src/lib/issues.unit.test.ts
```diff
@@ -271,4 +275,79 @@ describe('issues sorting', () => {
...
+  });
+
+  it('should return 0 if the group has no refs', () => {
+    const emptyGroupRef = {
+      ...mockCategoryRef,
+      plugin: 'eslint',
+      slug: 'nonexistent-group',
+    };
+    expect(calculateGroupImpact(emptyGroupRef, mockAudit, mockReport)).toBe(0);
+  });
```
matejchalk, line 352: Empty refs isn't really a realistic scenario, our schema validations demand a non-empty array. The `!group?.refs.length` condition is more concerned with `group` being `undefined`, i.e. when no matching plugin and group could be found.

packages/models/src/lib/implementation/utils.ts
```diff
@@ -108,10 +108,10 @@ export function getMissingRefsForCategories(
 }
 
 export function missingRefsForCategoriesErrorMsg(
-  categories: CategoryConfig[],
   plugins: PluginConfig[] | PluginReport[],
+  categories?: CategoryConfig[],
 ) {
-  const missingRefs = getMissingRefsForCategories(categories, plugins);
+  const missingRefs = getMissingRefsForCategories(plugins, categories);
```
matejchalk, line 114: The parameters are quite unintuitive now. Both functions are primarily with categories (going by their name), yet the categories are an optional second parameter.

I would expect `categories: CategoryConfig[] | undefined, plugins: PluginConfig[] | PluginReport[]`.  It would be a mistake to call this function without passing in anything for categories, so `| undefined` is more appropriate than `?`.  And since categories are most relevant, they should be specified first, the plugins are more like additional context here.

packages/utils/src/lib/reports/generate-md-report.ts
```diff
@@ -26,19 +30,25 @@ export function auditDetailsAuditValue({
   )} (score: ${formatReportScore(score)})`;
 }
 
+function hasCategories(
+  report: ScoredReport,
+): report is ScoredReport & { categories: ScoredCategoryConfig[] } {
```
matejchalk, line 35: It would be safer to derive categories type from `ScoredReport` (TS doesn't validate if the type predicate matches the implementation):

```suggestion
): report is ScoredReport & Required<Pick<ScoredReport, 'categories'>> {
```

packages/ci/src/lib/issues.unit.test.ts
hanna-skryl: Interesting 🤔 `!group?.refs.length` check was a bit misleading, but now I get it

matejchalk (approved):

PR #880, opened 2024-11-22
feat(plugin-eslint): add support for flat config and v9

### LINKED ISSUE #875: ESLint 9 (flat config) support for Code PushUp plugin
## User story

As a developer, I want to update to the latest ESLint version and use the new flat configs instead of the legacy format. Because `@code-pushup/eslint-plugin` only supports ESLint 8, I'm unable to do that if want to reuse my ESLint configuration for Code PushUp.

## Related issues

- https://github.com/code-pushup/eslint-config/issues/24

## Acceptance criteria

- [x] `@code-pushup/eslint-plugin` should support flat config format and ESLint 9+
- [x] legacy format should still be supported

matejchalk:
Closes #875

- The `@code-pushup/eslint-plugin` now supports ESLint v9 and the flat config format.
- The format is autodetected from `process.env.ESLINT_USE_FLAT_CONFIG` and `ESLint.version`.
- The config path is optional, the default configuration is discovered using a new `findNearestFile` utility function.
- Dropped support for inline config object, which wasn't documented or used anyway. The reason is the new format isn't serializable as JSON anymore.
- E2E tested both config formats.
- Also tested manually (installed via Verdaccio registry) in external Nx example project using ESLint 9 combined with new version of `@code-pushup/eslint-config`.
- This repo still uses ESLint 8 (blocked by Nx migration), which serves to verify the plugin still works with v8 via Code PushUp job in CI.


packages/plugin-eslint/src/lib/meta/versions/detect.unit.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    vol.fromJSON({}, MEMFS_VOLUME);
+  });
+
+  it('should assume flat config if explicitly enabled using environment variable', async () => {
+    vi.stubEnv('ESLINT_USE_FLAT_CONFIG', 'true');
+    await expect(detectConfigVersion()).resolves.toBe('flat');
+  });
+
+  it('should assume legacy config if explicitly disabled using environment variable', async () => {
+    vi.stubEnv('ESLINT_USE_FLAT_CONFIG', 'false');
```
BioPhoton, line 17: Learned something new

packages/utils/src/lib/file-system.unit.test.ts
```diff
@@ -57,9 +58,9 @@ describe('crawlFileSystem', () => {
   beforeEach(() => {
     vol.fromJSON(
       {
-        ['README.md']: '# Markdown',
-        ['src/README.md']: '# Markdown',
-        ['src/index.ts']: 'const var = "markdown";',
+        'README.md': '# Markdown',
+        'src/README.md': '# Markdown',
+        'src/index.ts': 'const var = "markdown";',
```
BioPhoton, line 63: Thx!! Is there maybe a lint rule for such cases? I remember I mad the same mistake somewhere else too.

BioPhoton (approved):
Thx for the upgrade!

vmasek (approved):

packages/utils/src/lib/file-system.unit.test.ts
matejchalk: Found it! Will add it to shared config:
- https://github.com/code-pushup/eslint-config/issues/26

PR #898, opened 2024-12-21
refactor(plugin-js-packages): provide package json path + e2e test

vmasek:
- [x] option to provide `package.json` path
- [x] remove confusing parameter of `packageJsonPahts` that was used just to count outdated deps 
- [x] create e2e tests
- [x] fallback from `current` to `wanted` (`current` is missing if `node_modules` folder does not exist and we audit just using package log)

### DIFF
--- a/e2e/plugin-js-packages-e2e/mocks/fixtures/npm-repo/code-pushup.config.ts
+++ b/e2e/plugin-js-packages-e2e/mocks/fixtures/npm-repo/code-pushup.config.ts
@@ -1,0 +1,7 @@
+import jsPackagesPlugin from '@code-pushup/js-packages-plugin';
+import type { CoreConfig } from '@code-pushup/models';
+
+export default {
+  persist: { outputDir: './' },
+  plugins: [await jsPackagesPlugin({ packageManager: 'npm' })],
+} satisfies CoreConfig;
--- a/e2e/plugin-js-packages-e2e/mocks/fixtures/npm-repo/code-pushup.config.ts
+++ b/e2e/plugin-js-packages-e2e/mocks/fixtures/npm-repo/code-pushup.config.ts
@@ -1,7 +1,16 @@
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
 import jsPackagesPlugin from '@code-pushup/js-packages-plugin';
 import type { CoreConfig } from '@code-pushup/models';
 
+const thisConfigFolder = fileURLToPath(dirname(import.meta.url));
+
 export default {
-  persist: { outputDir: './' },
-  plugins: [await jsPackagesPlugin({ packageManager: 'npm' })],
+  persist: { outputDir: thisConfigFolder, format: ['json'] },
+  plugins: [
+    await jsPackagesPlugin({
+      packageManager: 'npm',
+      packageJsonPath: join(thisConfigFolder, 'package.json'),
+    }),
+  ],
 } satisfies CoreConfig;
--- a/e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
+++ b/e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
@@ -1,0 +1,98 @@
+import { cp } from 'node:fs/promises';
+import path from 'node:path';
+import { afterAll, beforeAll, expect, it } from 'vitest';
+import {
+  type AuditReport,
+  type Report,
+  reportSchema,
+} from '@code-pushup/models';
+import { nxTargetProject } from '@code-pushup/test-nx-utils';
+import { teardownTestFolder } from '@code-pushup/test-setup';
+import { E2E_ENVIRONMENTS_DIR, TEST_OUTPUT_DIR } from '@code-pushup/test-utils';
+import { executeProcess, readJsonFile } from '@code-pushup/utils';
+
+describe('plugin-js-packages', () => {
+  const fixturesDir = path.join(
+    'e2e',
+    'plugin-js-packages-e2e',
+    'mocks',
+    'fixtures',
+  );
+  const fixturesNPMDir = path.join(fixturesDir, 'npm-repo');
+
+  const envRoot = path.join(
+    E2E_ENVIRONMENTS_DIR,
+    nxTargetProject(),
+    TEST_OUTPUT_DIR,
+  );
+  const npmRepoDir = path.join(envRoot, 'npm-repo');
+
+  beforeAll(async () => {
+    await cp(fixturesNPMDir, npmRepoDir, { recursive: true });
+  });
+
+  afterAll(async () => {
+    await teardownTestFolder(npmRepoDir);
+  });
+
+  it('should run JS packages plugin for NPM and create report.json', async () => {
+    const { code } = await executeProcess({
+      command: 'node',
+      args: [
+        '../../node_modules/@code-pushup/cli/src/index.js',
+        'collect',
+        '--no-progress',
+      ],
+      cwd: npmRepoDir,
+    });
+
+    expect(code).toBe(0);
+
+    const report = await readJsonFile<Report>(
+      path.join(npmRepoDir, 'report.json'),
+    );
+
+    const plugin = report.plugins[0]!;
+    const npmAuditProd = plugin.audits.find(
+      ({ slug }) => slug === 'npm-audit-prod',
+    )!;
+    const npmOutdatedProd = plugin.audits.find(
+      ({ slug }) => slug === 'npm-outdated-prod',
+    )!;
+
+    expect(plugin.packageName).toBe('@code-pushup/js-packages-plugin');
+    expect(plugin.audits).toHaveLength(4);
+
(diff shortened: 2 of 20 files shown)


e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        expect.objectContaining({ slug: 'updates' }),
+      ]),
+      plugins: expect.arrayContaining([
+        expect.objectContaining({
+          packageName: '@code-pushup/js-packages-plugin',
+          audits: expect.arrayContaining([
+            expect.objectContaining({
+              slug: 'npm-audit-prod',
+              displayValue: expect.stringMatching(/\d vulnerabilities/),
+              value: expect.closeTo(7, 10), // there are 7 vulnerabilities (6 high, 1 low) at the time
```
matejchalk, line 73: This will fail as soon as a new vulnerability is reported. Using `expect.closeTo` doesn't solve anything here, because the values are integers, so the difference won't be less than 10 decimals.

A more future proof assertion would be something like `expect(audit.value).toBeGreaterThanOrEqual(7)`, but that doesn't have an inline variant, so you'll need to break this up into multiple assertions.

e2e/plugin-js-packages-e2e/mocks/fixtures/npm-repo/code-pushup.config.ts
```diff
@@ -0,0 +1,32 @@
...
+      refs: [
+        {
+          type: 'group',
+          plugin: 'js-packages',
+          slug: 'npm-outdated',
+          weight: 1,
+        },
+      ],
+    },
+  ],
```
matejchalk, line 31: I don't think the categories need to be included here. The test only asserts the slugs are in the resulting report, but that's not the responsibility of this plugin.

e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
```diff
@@ -0,0 +1,99 @@
...
+              details: expect.objectContaining({
+                issues: expect.any(Array),
```
matejchalk, line 75: How about checking for one of the expected vulnerabilities?

```suggestion
                issues: expect.arrayContaining([
                  {
                    "message": "`express`'s dependency `connect` has a **high** vulnerability in versions **<=2.30.2**. Fix available: Update `express` to version **4.21.2** (breaking change). More information: [methodOverride Middleware Reflected Cross-Site Scripting in connect](https://github.com/advisories/GHSA-3fw8-66wf-pr7m)",
                    "severity": "error"
                  },
                ]),
```

Might require a regex for the message though, since the suggested `4.21.2` update is probably just the latest `express` version :thinking:

e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
```diff
@@ -0,0 +1,99 @@
...
+            expect.objectContaining({
+              slug: 'npm-outdated-prod',
+              displayValue: '1 major outdated package version',
+              value: 1,
+              score: 0,
+              details: {
+                issues: expect.arrayContaining([
+                  expect.objectContaining({
+                    message: expect.stringMatching(
+                      /Package \[`express`].*\*\*major\*\* update from \*\*\d+\.\d+\.\d+\*\* to \*\*\d+\.\d+\.\d+\*\*/,
```
matejchalk, line 87: Only the new version will change over time, otherwise you could check the full message :thinking: 

```suggestion
                      /^Package \[`express`]\(http:\/\/expressjs\.com\) requires a \*\*major\*\* update from \*\*3.0.0\*\* to \*\*\d+\.\d+\.\d+\*\*\.$/,
```

e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
```diff
@@ -0,0 +1,99 @@
...
+                    ),
+                    severity: 'error',
+                  }),
+                ]),
```
matejchalk, line 91: Since there aren't any other issues or issue fields than the ones you mention, you can omit the `expect.arrayContaining` and `expect.objectContaining` wrappers.

```suggestion
                issues: [
                  {
                    message: expect.stringMatching(
                      /Package \[`express`].*\*\*major\*\* update from \*\*\d+\.\d+\.\d+\*\* to \*\*\d+\.\d+\.\d+\*\*/,
                    ),
                    severity: 'error',
                  },
                ],
```

e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    expect(code).toBe(0);
+    expect(stderr).toBe('');
+
+    const report = await readJsonFile<Report>(
+      path.join(npmRepoDir, 'report.json'),
+    );
+
+    expect(() => reportSchema.parse(report)).not.toThrow();
+    expect.objectContaining({
```
matejchalk, line 61: Umm ... this doesn't assert anything :sweat_smile: Assymetrics matchers like `expect.objectContaining` need to be used with `expect(...).toEqual(...)` or `expect(...).toHaveBeenCalledWith(...)`. On their own they don't do anything - how could they, when the value (in this case `report`) isn't being specified anywhere.

The assertion would need to be:

```ts
expect(report).toEqual(expect.objectContaining<Partial<Report>>({
  // ...
}));
```

(BTW without adding a type parameter, everything in the matcher is typed as `any`.)

However, you may need to break this into multiple assertions - see my other comment about replacing `expect.closeTo(7)` with `expect(...).toBeGreaterThan(7)`. So it'll probably be something along these lines in the end:

```ts
const plugin = report.plugins[0]!;
expect(plugin.packageName).toBe('@code-pushup/js-packages-plugin');
expect(plugin.audits).toHaveLength(2);
expect(plugin.audits[0]).toEqual<AuditReport>({
  // ...
  value: expect.any(Number),
}));
expect(plugin.audits[0]!.value).toBeGreaterThan(7);
expect(plugin.audits[1]).toEqual<AuditReport>({
  // ...
}));
```

e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  it('should run JS packages plugin for NPM and create report.json', async () => {
+    const { code: installCode } = await executeProcess({
+      command: 'npm',
+      args: ['install'],
+      cwd: npmRepoDir,
+    });
+
+    expect(installCode).toBe(0);
```
matejchalk, line 45: Since neither `npm audit` or `npm outdated` rely on packages actually being installed in `node_modules`, I'm guessing running `npm install` only serve to create the `package-lock.json`? It would be more optimal to just commit the `package-lock.json` in the test fixture, that we way we don't have to run `npm install` and the test will be faster.

matejchalk (changes requested):

e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
```diff
@@ -0,0 +1,99 @@
+import { cp } from 'node:fs/promises';
+import path from 'node:path';
+import { afterAll, beforeAll, expect, it } from 'vitest';
+import { type Report, reportSchema } from '@code-pushup/models';
+import { nxTargetProject } from '@code-pushup/test-nx-utils';
+import { teardownTestFolder } from '@code-pushup/test-setup';
+import {
+  E2E_ENVIRONMENTS_DIR,
+  TEST_OUTPUT_DIR,
+  omitVariableReportData,
```
matejchalk, line 10: This is unused.

e2e/plugin-js-packages-e2e/tests/plugin-js-packages.e2e.test.ts
vmasek: I uderstood the assertion purpose differently, thanks dor explanation

packages/plugin-js-packages/src/lib/config.unit.test.ts
```diff
@@ -8,18 +8,6 @@ import {
...
-    expect(() =>
-      jsPackagesPluginConfigSchema.parse({
-        auditLevelMapping: { moderate: 'error' },
-        checks: ['audit'],
-        packageManager: 'yarn-classic',
-        dependencyGroups: ['prod'],
-        packageJsonPaths: ['./ui-app/package.json', './ui-e2e/package.json'],
-      } satisfies JSPackagesPluginConfig),
-    ).not.toThrow();
-  });
```
matejchalk, line 21: I don't see why we shouldn't keep this test. Why not just adjust `packageJsonPath`?

vmasek: We can, I deleted it as it felt too similar to the one bellow, but this one uses just `prod` and different package manager, I'll add it back

PR #911, opened 2025-01-07
fix: handle skipped audits and groups

### LINKED ISSUE #903: Fix `onlyAudits` and `onlyGroups` errors in Lighthouse plugin
### What happened?

When using `onlyAudits` or `onlyGroups` in the Lighthouse plugin configuration, unexpected errors are thrown, indicating that category references are invalid due to missing audits or groups.

Examples of errors:
- For `onlyGroups: ['performance']`:

```
custom: The following category references need to point to an audit or group: lighthouse#accessibility (group), lighthouse#best-practices (group), lighthouse#seo (group)
```

- For `onlyAudits: ['csp-xss']`:

```
custom: The following category references need to point to an audit or group: lighthouse#performance (group), lighthouse#accessibility (group), lighthouse#seo (group)
```

- Categories should dynamically adjust their references to match the filtered audits and groups.
- Invalid references should be removed without throwing errors.
- Groups with no valid references should be excluded from the final configuration.
- Informational logs should detail which audits, groups, and categories were filtered out.

### What steps did you take?

Set `onlyGroups` in `lighthouseCoreConfig`:
   ```ts
   export const lighthouseCoreConfig = async (
     url: string,
   ): Promise<CoreConfig> => ({
     plugins: [
       await lighthousePlugin(url, {
         onlyGroups: ['performance'],
       }),
     ],
     categories: lighthouseCategories,
   });
```

Set `onlyAudits` in `lighthouseCoreConfig`:
```ts
export const lighthouseCoreConfig = async (
  url: string,
): Promise<CoreConfig> => ({
  plugins: [
    await lighthousePlugin(url, {
      onlyAudits: ['csp-xss'],
    }),
  ],
  categories: lighthouseCategories,
});
```

### Proposed Fix

- Update the Lighthouse plugin to dynamically filter invalid category references based on `isSkipped` audits and groups.
- Remove categories with no valid references.
- Log skipped audits, groups, and categories for transparency.

_No response_

MacOS

### Node version

v22.10.0

### Relevant log output

```shell

```

hanna-skryl:
Fixes #903 

This fix enhances the filtering of report results by audits and groups through the core configuration (currently implemented for the Lighthouse plugin only).

The `isSkipped` property is introduced in the `PluginConfig` metadata for audits and groups, ensuring the complete set of plugin audits and groups is retained during `CoreConfig` validation to prevent unexpected errors. Filtering is now deferred to the filter middleware, which accurately handles skipped items by removing categories that contain only skipped or zero-weight references.

### DIFF
--- a/packages/cli/src/lib/implementation/filter.middleware.ts
+++ b/packages/cli/src/lib/implementation/filter.middleware.ts
@@ -4,13 +4,19 @@ import type {
   PluginConfig,
 } from '@code-pushup/models';
 import { filterItemRefsBy } from '@code-pushup/utils';
+import {
+  applyFilters,
+  extractSkippedItems,
+  filterPluginsFromCategories,
+  filterSkippedItems,
+  isValidCategoryRef,
+} from './filter.middleware.utils.js';
 import type { FilterOptions, Filterables } from './filter.model.js';
 import {
   handleConflictingOptions,
-  isValidCategoryRef,
   validateFilterOption,
-  validateFilteredCategories,
   validateFinalState,
+  validateSkippedCategories,
 } from './validate-filter-options.utils.js';
 
 // eslint-disable-next-line max-lines-per-function
@@ -27,16 +33,8 @@ export function filterMiddleware<T extends FilterOptions>(
     verbose = false,
   } = originalProcessArgs;
 
-  const plugins = processPlugins(rcPlugins);
+  const plugins = filterSkippedInPlugins(rcPlugins);
   const categories = filterSkippedCategories(rcCategories, plugins);
-
-  if (rcCategories && categories) {
-    validateFilteredCategories(rcCategories, categories, {
-      onlyCategories,
-      skipCategories,
-      verbose,
-    });
-  }
 
   if (
     skipCategories.length === 0 &&
@@ -44,6 +42,9 @@ export function filterMiddleware<T extends FilterOptions>(
     skipPlugins.length === 0 &&
     onlyPlugins.length === 0
   ) {
+    if (rcCategories && categories) {
+      validateSkippedCategories(rcCategories, categories, verbose);
+    }
     return {
       ...originalProcessArgs,
       ...(categories && { categories }),
@@ -54,17 +55,18 @@ export function filterMiddleware<T extends FilterOptions>(
   handleConflictingOptions('categories', onlyCategories, skipCategories);
   handleConflictingOptions('plugins', onlyPlugins, skipPlugins);
 
+  const skippedPlugins = extractSkippedItems(rcPlugins, plugins);
+  const skippedCategories = extractSkippedItems(rcCategories, categories);
+
   const filteredCategories = applyCategoryFilters(
     { categories, plugins },
-    skipCategories,
-    onlyCategories,
-    verbose,
+    skippedCategories,
+    { skipCategories, onlyCategories, verbose },
   );
   const filteredPlugins = applyPluginFilters(
     { categories: filteredCategories, plugins },
-    skipPlugins,
-    onlyPlugins,
-    verbose,
+    skippedPlugins,
+    { skipPlugins, onlyPlugins, verbose },
   );
   const finalCategories = filteredCategories
     ? filterItemRefsBy(filteredCategories, ref =>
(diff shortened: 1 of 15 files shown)


packages/plugin-lighthouse/src/lib/lighthouse-plugin.unit.test.ts
```diff
@@ -17,47 +17,36 @@ describe('lighthousePlugin-config-object', () => {
...
-    expect(pluginConfig.audits[0]).toEqual(
+    expect(
+      pluginConfig.audits.find(({ slug }) => slug === 'first-contentful-paint'),
+    ).toEqual(
       expect.objectContaining({
-        slug: 'first-contentful-paint',
+        isSkipped: false,
       }),
     );
   });
```
matejchalk, line 33: I'm missing some assertion that the other audits have `isSkipped: true`. (Same for groups.)

packages/cli/src/lib/implementation/validate-filter-options.utils.ts
```diff
@@ -76,6 +146,30 @@ export function validateFinalState(
...
+  plugin.groups.forEach(group => {
+    group.refs.forEach(ref => {
+      weightMap.set(ref.slug, (weightMap.get(ref.slug) ?? 0) + ref.weight);
+    });
+  });
+  const totalWeight = plugin.audits.reduce(
+    (sum, audit) => sum + (weightMap.get(audit.slug) ?? 0),
+    0,
+  );
+  return totalWeight === 0;
```
matejchalk, line 172: This implementation doesn't take into account that an audit can be referenced by multiple groups within the plugin. I would simply sum up the weights for each group, and then check if some group has a total weight of 0.

packages/cli/src/lib/implementation/validate-filter-options.utils.ts
```diff
@@ -55,6 +63,68 @@ export function validateFilterOption(
...
+  if (invalidArgs.length > 0) {
+    throw new OptionValidationError(
+      invalidArgs
+        .map(
+          ({ option, args }) =>
+            `The --${option} argument references skipped categories: ${args.join(', ')}`,
+        )
+        .join('. '),
+    );
+  }
```
matejchalk, line 102: I wouldn't throw validation errors in this case. Skipping an already skipped category seems harmless to me. And we could simply remove skipped categories from `onlyCategories`. Only if all of `onlyCategories` are skipped would we consider that a validation error, because that leaves us with nothing to run. But other than that, I would just log a warning. The "skipping already skipped" warning should probably be behind verbose flag as well. :thinking:

packages/cli/src/lib/implementation/filter.middleware.ts
```diff
@@ -141,3 +163,40 @@ function filterPluginsFromCategories({
...
+
+function filterSkippedItems<T extends { isSkipped?: boolean }>(
+  items: T[] | undefined,
+): Omit<T, 'isSkipped'>[] {
+  return (items ?? [])
+    .filter(({ isSkipped }) => isSkipped !== true)
+    .map(({ isSkipped, ...props }) => props);
+}
+
+export function processPlugins(plugins: PluginConfig[]): PluginConfig[] {
```
matejchalk, line 175: The name `processPlugins` isn't very self-explanatory:

```suggestion
export function filterSkippedInPlugins(plugins: PluginConfig[]): PluginConfig[] {
```

packages/models/src/lib/implementation/schemas.ts
```diff
@@ -96,6 +101,9 @@ export function metaSchema(options?: {
       docsUrl: docsUrlDescription
         ? docsUrlSchema.describe(docsUrlDescription)
         : docsUrlSchema,
+      isSkipped: isSkippedDescription
+        ? isSkippedSchema.describe(isSkippedDescription)
+        : isSkippedSchema,
```
matejchalk, line 106: Could you please also re-generate the Markdown docs (`npx nx generate-docs models`)?

packages/cli/src/lib/implementation/filter.middleware.utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+  switch (ref.type) {
+    case 'audit':
+      return plugin.audits.some(({ slug }) => slug === ref.slug);
+    case 'group':
+      return plugin.groups?.some(({ slug }) => slug === ref.slug) ?? false;
+  }
+}
+
+export function filterPluginsFromCategories({
```
hanna-skryl, line 56: I noticed a bug with the `onlyCategories` and `skipCategories` arguments. Groups and audits that are associated with removed categories still appear in the report. Although the `categories` and `plugins` arrays contain the correct items, plugins still include all available groups and audits. I'm not sure whether to fix this issue in this PR.

packages/cli/src/lib/implementation/validate-filter-options.utils.ts
```diff
@@ -1,81 +1,85 @@
...
-      : [];
-  const invalidItems = itemsToFilter.filter(
-    item => !validItems.some(({ slug }) => slug === item),
+  const skippedItemsSet = new Set(skippedItems);
+  const validItemsSet = new Set(
+    isCategoryOption(option)
+      ? categories.map(({ slug }) => slug)
+      : isPluginOption(option)
+        ? plugins.map(({ slug }) => slug)
+        : [''],
```
matejchalk, line 29: An array with an empty string? :raised_eyebrow: That looks like a mistake:

```suggestion
        : [],
```

packages/cli/src/lib/implementation/filter.middleware.utils.ts
matejchalk: I wouldn't consider that a bug. The CLI has no control over which audits a plugin measures, it can only skip the whole plugin (via `--onlyPlugins` or `--skipPlugins`). So since the groups and audits are measured anyway, why hide the results from the user? The audits and groups themselves could still be useful, even if they're not referenced by any category.

hanna-skryl: Makes sense 👍

matejchalk (approved):

PR #916, opened 2025-01-10
feat(utils): add string helper

BioPhoton:
This PR includes helper for the TypeScript plugin #902


packages/utils/src/lib/string.ts
```diff
@@ -0,0 +1,58 @@
...
+    .replace(/[\s_]+/g, '-') // Replace spaces and underscores with hyphens
+    .toLowerCase() as CamelCaseToKebabCase<T>;
+}
+
+/**
+ * Formats a slug to a readable title.
+ * @param string - The slug to format.
+ * @returns The formatted title.
+ */
+export function kebabCaseToSentence(string = '') {
```
aramirezj, line 40: Since  ```export function kebabCaseToCamelCase(string: string) {``` the param it's not optional, I would not do this optional either
```suggestion
export function kebabCaseToSentence(string: string) {
```

aramirezj (changes requested):

packages/utils/src/lib/string.ts
matejchalk: Agreed. There would be no sense in calling `kebabCaseToSentence()` without arguments. Same goes for `camelCaseToSentence()`.

packages/utils/src/lib/string.ts
```diff
@@ -0,0 +1,58 @@
...
+    .replace(/-/g, ' ')
+    .replace(/\b\w/g, letter => letter.toUpperCase());
+}
+
+/**
+ * Formats a slug to a readable title.
+ * @param string - The slug to format.
+ * @returns The formatted title.
+ */
+export function camelCaseToSentence(string = '') {
```
matejchalk, line 51: This function is missing unit tests.

packages/utils/src/lib/string.unit.test.ts
```diff
@@ -0,0 +1,81 @@
...
+});
+
+describe('kebabCaseToSentence', () => {
+  it('should convert simple slug to title case', () => {
+    expect(kebabCaseToSentence('hello-world')).toBe('Hello World');
+  });
+
+  it('should handle multiple hyphens', () => {
+    expect(kebabCaseToSentence('this-is-a-title')).toBe('This Is A Title');
+  });
```
matejchalk, line 66: What you've implemented is called **title case**, not **sentence case**. See [Title Case vs. Sentence Case: What’s the Difference?](https://www.grammarly.com/blog/sentences/title-case-sentence-case/):

> Title case and sentence case are two different styles of capitalization used in writing titles, headings, and headlines. With title case, the first letter of every major word is capitalized, while articles, conjunctions, or prepositions are lowercase—unless they are the first word in the title. In sentence case, only the first word is capitalized, along with any proper nouns in the title or heading.
> 
> - Title case example: _The Mystery of the Missing Red Ruby_
> - Sentence case example: _The mystery of the missing red ruby_

packages/utils/src/lib/string.ts
matejchalk, line 1: Maybe `string.ts` a bit too broad? :thinking: How about `casings.ts` or `case-conversions.ts`?

matejchalk, line 1: We have a `capitalize` function in `transform.ts`. Since it's also about converting string case, it could be moved in here.

packages/utils/src/lib/types.ts
```diff
@@ -15,3 +15,10 @@ export type WithRequired<T, K extends keyof T> = Prettify<
 >;
 
 export type Prettify<T> = { [K in keyof T]: T[K] };
+
+export type CamelCaseToKebabCase<T extends string> =
+  T extends `${infer First}${infer Rest}`
+    ? Rest extends Uncapitalize<Rest>
+      ? `${Lowercase<First>}${CamelCaseToKebabCase<Rest>}`
+      : `${Lowercase<First>}-${CamelCaseToKebabCase<Rest>}`
+    : T;
```
matejchalk, line 24: The inferred type doesn't match the returned string from `camelCaseToKebabCase`:

image

It's important (for developer sanity) that the _type is consistent with reality_, even if that means making the function less "robust". For example, the `'hello_world test'` input string isn't in camelCase, so converting it need not be the function's responsibility. After all, the function is called `camelCaseToKebabCase`, it doesn't claim to convert _anything_ to kebab-case. Simplifying the runtime conversion will also simplify the type conversion - reimplementing the current conversion in types would very challenging otherwise.

packages/utils/src/lib/string.ts
```diff
@@ -0,0 +1,58 @@
+import type { CamelCaseToKebabCase } from './types.js';
+
+/**
+ * Converts a kebab-case string to camelCase.
+ * @param string - The kebab-case string to convert.
+ * @returns The camelCase string.
+ */
+export function kebabCaseToCamelCase(string: string) {
```
matejchalk, line 8: Since the inverse function uses a `CamelCaseToKebabCase<T>`, why not create a `KebabCaseToCamelCase<T>` type also? It's actually a lot easier to do the conversion in this direction:

```ts
export type KebabCaseToCamelCase<T extends string> =
  T extends `${infer First}-${infer Rest}`
    ? `${First}${Capitalize<KebabCaseToCamelCase<Rest>>}`
    : T;
```

packages/utils/src/lib/types.ts
matejchalk: BTW, Vitest supports [Testing Types](https://vitest.dev/guide/testing-types). We should consider that for these complex types. I actually used it recently [here](https://gitlab.com/code-pushup/gitlab-pipelines-template/-/blob/main/src/env.test-d.ts?ref_type=heads) and it works really well, all I had to do was add `--typecheck` to `vitest` command. We could leave that for another PR, though.

packages/utils/src/lib/string.ts
BioPhoton: fixed

packages/utils/src/lib/string.ts
BioPhoton: refactored and implemented

packages/utils/src/lib/string.unit.test.ts
BioPhoton: I will refactor based on the info

packages/utils/src/lib/string.ts
BioPhoton: renamed!

BioPhoton: moved. good point

packages/utils/src/lib/string.ts
BioPhoton: Thanks! added and used

packages/utils/src/lib/types.ts
BioPhoton: Fixed + added type tests

matejchalk (approved):
Good job with the refactor and the tests :+1: 

I think the type tests need some more setup (see comments), but I don't see it as a blocker for this PR.

PR #926, opened 2025-01-24
test: add custom matcher for markdown table rows

### LINKED ISSUE #757: Custom matcher for Markdown tables
## Motivation

As a developer, I want a convenient way to test if a Markdown string contains a table with specific cell content. 

Because table cells are padded with spaces according to columns alignments (since #755), matching the content of all tables in a row is a bit tricky:

```ts
// 👇 this doesn't take into account padding within `| ... |` cells
expect(md).toMatch('|🏷 Category|⭐ Score|🛡 Audits|');

// 👇 this regex ignores the padding, but it isn't very readable
expect(md).toMatch(/\|\s*🏷 Category\s*\|\s*⭐ Score\s*\|\s*🛡 Audits\s*\|/);
```

## Proposed solution

Create a [custom matcher](https://vitest.dev/guide/extending-matchers) which abstracts away the formatting details.

So instead of the `/\|\s*...\s*\|\s*...\s*\|\s*...\s*\|/` regular expression above, the tests would be simplified to:

```ts
// 👇 checks that there is some MD table row with these cell contents, regardless of spaces between bars 
expect(md).toContainMarkdownTableRow(['🏷 Category', '⭐ Score', '🛡 Audits']);
```

The matcher should live in the `test-setup` Nx project and be strongly typed using a declaration file linked via `tsconfig.test.json` (see [docs](https://vitest.dev/guide/extending-matchers)):

<details><summary><code>vitest.d.ts</code> example</summary>
<p>

```ts
/* eslint-disable @typescript-eslint/consistent-type-definitions, @typescript-eslint/no-empty-interface, @typescript-eslint/no-shadow, @typescript-eslint/no-unused-vars */
import type { Assertion, AsymmetricMatchersContaining } from 'vitest';

// see Vitest guide: https://vitest.dev/guide/extending-matchers

interface CustomMatchers<R = unknown> {
  toContainMarkdownTableRow: (cells: string[]) => void;
  // ... declarations of other matchers ...
}

declare module 'vitest' {
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  interface Assertion<T = any> extends CustomMatchers<T> {}
  interface AsymmetricMatchersContaining extends CustomMatchers {}
}

// HACK: prevents IDEs removing "unused" types in import optimization
const x: Assertion | AsymmetricMatchersContaining = {};
```

</p>
</details> 

## Similar issues

- #552
- #744

hanna-skryl:
Closes #757 

This PR introduces a custom matcher, `toContainMarkdownTableRow`, designed to validate the presence of specific rows within Markdown tables during testing.

### DIFF
--- a/testing/test-setup/src/lib/extend/markdown-table.matcher.ts
+++ b/testing/test-setup/src/lib/extend/markdown-table.matcher.ts
@@ -1,0 +1,44 @@
+import type { SyncExpectationResult } from '@vitest/expect';
+import { expect } from 'vitest';
+
+export type CustomMarkdownTableMatchers = {
+  toContainMarkdownTableRow: (cells: string[]) => void;
+};
+
+expect.extend({
+  toContainMarkdownTableRow: assertMarkdownTableRow,
+});
+
+function assertMarkdownTableRow(
+  actual: string,
+  expected: string[],
+): SyncExpectationResult {
+  const rows = actual
+    .split('\n')
+    .map(line => line.trim())
+    .filter(line => line.startsWith('|') && line.endsWith('|'))
+    .map(line =>
+      line
+        .split('|')
+        .map(cell => cell.trim())
+        .filter(Boolean),
+    )
+    .filter(row => row.some(cell => /\w/.test(cell)));
+
+  const pass = rows.some(
+    row =>
+      row.length === expected.length &&
+      row.every((cell, i) => cell === expected[i]),
+  );
+  return pass
+    ? {
+        pass,
+        message: () =>
+          `Expected markdown not to contain a table row with cells: ${expected.join(', ')}`,
+      }
+    : {
+        pass,
+        message: () =>
+          `Expected markdown to contain a table row with cells: ${expected.join(', ')}`,
+      };
+}
--- a/testing/test-setup/src/lib/extend/markdown-table.matcher.ts
+++ b/testing/test-setup/src/lib/extend/markdown-table.matcher.ts
@@ -19,11 +19,10 @@ function assertMarkdownTableRow(
     .filter(line => line.startsWith('|') && line.endsWith('|'))
     .map(line =>
       line
-        .split('|')
-        .map(cell => cell.trim())
+        .split(/(?<!\\)\|/)
+        .map(cell => cell.replace(/\\\|/g, '|').trim())
         .filter(Boolean),
-    )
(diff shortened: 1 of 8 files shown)


testing/test-setup/src/lib/extend/markdown-table.matcher.ts
```diff
@@ -0,0 +1,44 @@
...
+    .split('\n')
+    .map(line => line.trim())
+    .filter(line => line.startsWith('|') && line.endsWith('|'))
+    .map(line =>
+      line
+        .split('|')
+        .map(cell => cell.trim())
+        .filter(Boolean),
+    )
+    .filter(row => row.some(cell => /\w/.test(cell)));
```
matejchalk, line 26: Not sure if these filters are a good idea. Empty cells or cells with non-word characters (e.g. `-`) are perfectly, so I don't see why we'd treat differently to other cells.

testing/test-setup/src/lib/extend/markdown-table.matcher.ts
```diff
@@ -0,0 +1,44 @@
...
+  actual: string,
+  expected: string[],
+): SyncExpectationResult {
+  const rows = actual
+    .split('\n')
+    .map(line => line.trim())
+    .filter(line => line.startsWith('|') && line.endsWith('|'))
+    .map(line =>
+      line
+        .split('|')
```
matejchalk, line 22: There's an edge case where this simple split by `|` might get tripped up. A Markdown cell could contain an escaped `\|`, which shouldn't be interpreted as a column separator.

E.g. this markdown:

```md
| Package  | Versions             |
| :------- | :------------------- |
| `eslint` | `^8.0.0 \|\| ^9.0.0` |
```

renders 2 columns and the escaped bars turn into cell content:

| Package  | Versions             |
| :------- | :------------------- |
| `eslint` | `^8.0.0 \|\| ^9.0.0` |

To handle these cases properly, you can use a regex with negative lookbehind as the separator:

```suggestion
        .split(/(?<!\\)\|/)
```

matejchalk (commented):
This looks so much nicer to use :+1:

testing/test-setup/src/lib/extend/markdown-table.matcher.ts
matejchalk: The `.filter(Boolean)` is still present, would ignore empty cells.

matejchalk (approved):

Issue #888, opened 2024-12-05
Improve logging and error handling


matejchalk:
## The problem

Users have a hard time setting up our CLI and plugins on their own.

### Error handling

- Users have a hard time understanding errors from the CLI.
- Errors don't give hints on how to fix them.
- It often isn't clear in which step an error occurred.
- Debugging errors using `--verbose` flag is of very little help, as this log level is barely used.
- Some errors are unnecessarily verbose (e.g. Zod schema validation).
- Output and error streams from executed processes are often silenced unless there's an error, hiding potentially valuable information.

### Logging
 
- The only progress indicator is a progress bar which is supposed to update discretely whenever a plugin is finished running (but has actually been broken for a long time, so it just skips to 100% immediately).
- For long-running plugin execution, there's no way for a user to tell if something is still happening or if it's hanging.
- Aside from the final report, very little information is logged about what the CLI is doing. This lack of transparency contributes to users struggling to understand where an error came from, as well as how the CLI is supposed to work in general.
- The progress bar also causes strange UI side-effects (scrolling logs out of viewport), is difficult to test and isn't suitable for logging in CI pipelines.

## Proposed solution

- [ ] replace progress bar with spinners
- [ ] spinners should be shown whenever some potentially long-running process is running
- [ ] CLI should log high-level steps being performed (e.g. parsing config, executing plugin 2 out of 4, formatting report)
- [ ] all plugin logs should be prefixed by plugin which emitted them
- [ ] plugins should use logger with same verbose settings as CLI
- [ ] Zod errors should be made more compact
- [ ] ensure every error is given proper context (in which step is occurred)
- [ ] add E2E test for what's being printed to stdout in `cli-e2e` (part of 1st contact)

Issue #695, opened 2024-06-05
Logs from plugins do not appear


Tlacenka:
### User story

As a user, I want to know what is going on inside the plugins. I want to see not only errors but also warnings and other valuable information.

### Acceptance criteria

- [ ] When an `info` or `warning` log is posted from within a plugin, the user can see it (can be behind verbose flag).
- [ ] As before, `error` logs are also visible to the user.

vmasek:
Consider extending logger to log into log files so all errors from plugins could appear in some temporary file.

Turn off logging `--help` options on error.

BioPhoton:
> Consider extending logger to log into log files so all errors from plugins could appear in some temporary file.

I would suggest to implement the stdout and stderr properly and log it directly to the console.

Not sure what the file can help more in comparison to styled and formatted logs in CI/locally.

PR #929, opened 2025-02-03
fix: nest files with timestamp for plugin runners

### LINKED ISSUE #930: Plugin runner config and output path gets overridden in parallel mode
### What happened?

When running plugins in parallel or running same plugin multiple times for different audits it overrides its runner config as the folder for each individual runner config is `node_modules/.code-pushup/<plugin-name>`.

Each runner of individual plugins should have its own unique folder to hold the config file and output file.
This can be achieved by adding one more level of nesting into runners folder as in `node_modules/.code-pushup/<timestamp>/<plugin-slug>`.

Also plugin slug should be used instead of hardcoded plugin name as it will be more versatile if people would reuse or extend existing plugins with their own variety (for example duplicating `coverage` plugin execution to get two reported coverages, one from Jest and one from Cypress component tests)

### What steps did you take?

1. Setup Code PushUp target with ESLint for multiple projects in one Nx monorepo
2. For better reproduction have different lint rules setup in the projects
3. Run plugins in parallel or via affected
4. Observe errors of rules not being found or other weird behavior related to overridden runner config

latest

Windows

### Node version

22

### Relevant log output

```shell

```

vmasek:
closes #930
part of #888 
part of #695
Additional changes:
- [x] makes sure all the logs from runner process will appear in CLI via logger `log`
- [x] adds some handy logs that help identify or reproduce issues
- [x] adds `--no-progress --verbose` to "code-pushup" command "node packages/cli/src/index.ts" that runns CP collection on this repo as it is very annoying to not have logs considering how frequently it is it breaks during PRs


packages/core/src/lib/implementation/runner.ts
```diff
@@ -26,7 +31,8 @@ export async function executeRunnerConfig(
   });
 
   // read process output from file system and parse it
-  const outputs = await readJsonFile(path.join(process.cwd(), outputFile));
+  const outputs = await readJsonFile(outputFile);
+  await removeDirectoryIfExists(path.dirname(outputFile));
```
matejchalk, line 35: This deserves a comment, took me a while to realize why you remove the directory after reading.

```suggestion
  // clean up plugin output directory
  await removeDirectoryIfExists(path.dirname(outputFile));
```

packages/plugin-eslint/src/bin.ts
```diff
@@ -1,3 +1,7 @@
+import process from 'node:process';
+import { Parser } from 'yargs/helpers';
 import { executeRunner } from './lib/runner/index.js';
 
-await executeRunner();
+const { runnerConfigPath, runnerOutputPath } = Parser(process.argv);
```
matejchalk, line 5: Why don't we declare the required Yargs options? Then the args will actually be validated, and we'll get the inferred types.

The logic could live in `@code-pushup/utils`, since many plugins are using the same pattern.

packages/plugin-eslint/src/lib/eslint-plugin.integration.test.ts
```diff
@@ -18,10 +19,19 @@ describe('eslintPlugin', () => {
...
+      args: (plugin.runner as RunnerConfig).args?.map(arg => {
+        if (arg.includes('bin.js')) {
+          return toUnixPath(arg.replace(path.dirname(thisDir), '<dirname>'));
+        }
+
+        return toUnixPath(arg).replace(
+          /\/eslint\/\d+\//,
+          '/eslint/<timestamp>/',
+        );
+      }),
```
matejchalk, line 31: This can be simplified:

```suggestion
      args: (plugin.runner as RunnerConfig).args?.map(arg =>
        toUnixPath(arg.replace(path.dirname(thisDir), '<dirname>')).replace(
          /\/eslint\/\d+\//,
          '/eslint/<timestamp>/',
        ),
      ),
```

packages/plugin-eslint/src/lib/runner.integration.test.ts
```diff
@@ -2,26 +2,36 @@ import os from 'node:os';
...
     const targets: ESLintTarget[] = [{ eslintrc, patterns }];
     const { audits } = await listAuditsAndGroups(targets);
-    await createRunnerConfig('bin.js', audits, targets);
+    const { outputFile } = await createRunnerConfig('bin.js', audits, targets);
+    return {
+      runnerOutputPath: outputFile,
+      runnerConfigPath: outputFile.replace(
+        'runner-output.json',
+        'plugin-config.json',
+      ),
```
matejchalk, line 33: Maybe we could add an optional `configFile` to the runner config schema? It would save you from doing this output file replacement in several tests. And the core could then also take care of deleting the config file as well :thinking:

packages/utils/src/lib/execute-process.ts
```diff
@@ -148,6 +150,9 @@ export function executeProcess(cfg: ProcessConfig): Promise<ProcessResult> {
   const date = new Date().toISOString();
   const start = performance.now();
 
+  const logCommand = [command, ...(args || [])].join(' ');
+  ui().logger.log(gray(`Executing command:\n${logCommand}`));
```
matejchalk, line 154: Nice improvement, adds a lot more transparency :+1:

project.json
```diff
@@ -5,7 +5,7 @@
     "code-pushup": {
       "executor": "nx:run-commands",
       "options": {
-        "command": "node packages/cli/src/index.ts",
+        "command": "node packages/cli/src/index.ts --no-progress --verbose",
```
matejchalk, line 8: :+1:

matejchalk (approved):
Thank you for taking care of this :slightly_smiling_face: 

Left some suggestions, but none of them are a blocker.

packages/plugin-eslint/src/lib/nx.integration.test.ts
```diff
@@ -11,7 +12,7 @@ import {
 
 type Project = 'cli' | 'core' | 'nx-plugin' | 'utils';
 
-describe('Nx helpers', () => {
+describe.skipIf(() => process.platform === 'win32')('Nx helpers', () => {
```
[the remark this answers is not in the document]
vmasek: For some reason type signature suggested the Fn, I'll fix it

matejchalk (approved):

PR #932, opened 2025-02-12
feat(plugin-typescript): add TS core logic

BioPhoton:


packages/plugin-typescript/src/lib/runner/runner.ts
```diff
@@ -0,0 +1,60 @@
...
+              issues: [...existingIssues, getIssueFromDiagnostic(diag)],
+            },
+          },
+        };
+      },
+      {} as unknown as Record<
+        CodeRangeName,
+        Pick<AuditOutput, 'slug' | 'details'>
+      >,
+    );
```
matejchalk, line 46: I see `as unknown as` brute-force casting :arrow_right: I complain :sweat_smile: 

The _true_ type of the `reduce` accumulator is actually a `Partial`, because not every code range need be emitted. Then the initial accumulator value is also valid (although a soft `as` would've been enough anyway).

There's also the question of where to put the accumulator type in `reduce`. In most situations, the type parameter is safest and most convenient (type used for implementation as well). There are some cases where the initial and final accumulator types are different, then it's a bit more tricky, but fortunately that's not the case here.

```suggestion
    const result = diagnostics.reduce<
      Partial<Record<CodeRangeName, Pick<AuditOutput, 'slug' | 'details'>>>
    >((acc, diag) => {
      const slug = tSCodeToAuditSlug(diag.code);
      const existingIssues: Issue[] =
        (acc[slug] && acc[slug].details?.issues) || ([] as Issue[]);
      return {
        ...acc,
        [slug]: {
          slug,
          details: {
            issues: [...existingIssues, getIssueFromDiagnostic(diag)],
          },
        },
      };
    }, {});
```

packages/plugin-typescript/src/lib/runner/runner.ts
```diff
@@ -0,0 +1,60 @@
...
+  return async (): Promise<AuditOutputs> => {
+    const diagnostics = await getTypeScriptDiagnostics({ tsconfig });
+    const result: Record<
+      CodeRangeName,
+      Pick<AuditReport, 'slug' | 'details'>
+    > = diagnostics.reduce(
+      (acc, diag) => {
+        const slug = tSCodeToAuditSlug(diag.code);
+        const existingIssues: Issue[] =
+          (acc[slug] && acc[slug].details?.issues) || ([] as Issue[]);
```
matejchalk, line 31: I see unnecessary casting :arrow_right: I complain :grin: 

The whole expression can be simplified, actually:

```suggestion
        const existingIssues: Issue[] = acc[slug]?.details?.issues ?? [];
```

packages/plugin-typescript/src/lib/runner/runner.ts
```diff
@@ -0,0 +1,60 @@
...
+        };
+      },
+      {} as unknown as Record<
+        CodeRangeName,
+        Pick<AuditOutput, 'slug' | 'details'>
+      >,
+    );
+
+    return expectedAudits.map(({ slug }) => {
+      const { details } = result[slug as CodeRangeName] ?? {};
```
matejchalk, line 49: Another unnecessary casting :sweat_smile:

The `AuditSlug` is actually an alias of `CodeRangeName`, so no conversion is needed: 

```suggestion
      const { details } = result[slug] ?? {};
```

packages/plugin-typescript/src/lib/runner/ts-runner.ts
```diff
@@ -0,0 +1,25 @@
...
+  tsconfig,
+}: DiagnosticsOptions): Promise<readonly Diagnostic[]> {
+  const { fileNames, options } = await loadTargetConfig(tsconfig);
+  try {
+    const program = createProgram(fileNames, options);
+    // @TODO use more fine-grained helpers like getSemanticDiagnostics instead of getPreEmitDiagnostics
+    return getPreEmitDiagnostics(program);
+  } catch (error) {
+    throw new Error(
+      `Can't create TS program in getDiagnostics. \n ${(error as Error).message}`,
```
matejchalk, line 22: We have a [utility function](https://github.com/code-pushup/cli/blob/main/packages/utils/src/lib/errors.unit.test.ts) that takes care of safely converting unknown errors to a string:

```suggestion
      `Can't create TS program in getDiagnostics.\n${stringifyError(error)}`,
```

packages/plugin-typescript/src/lib/runner/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+import { TS_CODE_RANGE_NAMES } from './ts-error-codes.js';
+import type { CodeRangeName } from './types.js';
+
+/**
+ * Transform the TypeScript error code to the audit slug.
+ * @param code - The TypeScript error code.
+ * @returns The audit slug.
+ * @throws Error if the code is not supported.
+ */
+export function tSCodeToAuditSlug(code: number): CodeRangeName {
```
matejchalk, line 22: Typo in function name:

```suggestion
export function tsCodeToAuditSlug(code: number): CodeRangeName {
```

packages/plugin-typescript/package.json
```diff
@@ -22,6 +22,10 @@
     "access": "public"
   },
   "type": "module",
-  "dependencies": {},
+  "dependencies": {
+    "@code-pushup/models": "0.58.0",
+    "@code-pushup/utils": "0.58.0",
+    "typescript": "5.5.4"
```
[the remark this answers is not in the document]
BioPhoton: made it <=4

packages/plugin-typescript/src/lib/runner/runner.ts
matejchalk: I think this change was overwritten by applying suggestion from https://github.com/code-pushup/cli/pull/932#discussion_r1952869368

packages/plugin-typescript/package.json
```diff
@@ -23,8 +23,8 @@
   },
   "type": "module",
   "dependencies": {
-    "@code-pushup/models": "0.58.0",
-    "@code-pushup/utils": "0.58.0"
+    "@code-pushup/models": "*",
+    "@code-pushup/utils": "*"
```
matejchalk, line 27: Only the current version should be specified here. Nx release takes care of bumping it usually, but I happened to make a release just now while this PR was still open :sweat_smile: If you run into problems, merging in `main` should fix it.

```suggestion
    "@code-pushup/models": "0.59.0",
    "@code-pushup/utils": "0.59.0"
```

matejchalk (approved):

PR #936, opened 2025-02-14
feat(plugin-typescript): add plugin logic

BioPhoton:
part of #902


packages/plugin-typescript/src/lib/schema.ts
```diff
@@ -0,0 +1,20 @@
...
+  onlyAudits: z
+    .array(z.enum(auditSlugs), {
+      description: 'Array with specific TsCodes to measure',
+    })
+    .optional(),
+});
+
+export type TypescriptPluginOptions = z.infer<
+  typeof typescriptPluginConfigSchema
+>;
```
matejchalk, line 20: OK, so there's quite a bit to improve with the types.

In addition to the inferred `TypescriptPluginOptions` here, there are 2 other exported types of the same name in `typescript-plugin.ts` and `types.ts`. The `FilterOptions` subtype is also duplicated.

I'm guessing these might be the reasons you don't you use the schema type in the `typescriptPlugin` function:
- The `tsconfig` is required in the inferred type, but for the user it's supposed to be optional.
- The `onlyAudits` is inferred as `string[]` from the `z.enum`, while `AuditSlug[]` gives the user better autocomplete.

There's a better way to solve both these problems:
- Zod can actually infer 2 different types from a schema:
   - `z.infer` infers the output type, i.e. the return type of `.parse()` (`.default`, `.transform`, etc.)
   - `z.input` infers the input type, i.e. what argument `.parse()` would consider (structurally) valid 
  So for the user-facing function, `z.input` is more appropriate - `tsconfig` will be optional here. But for internal code which runs on what's already been parsed, `z.infer` will be more accurate - `tsconfig` won't be optional anymore because the defaults been applied.
- You cast the values of `z.enum` as `[string, ...string[]]`. The casting is necessary because Zod's typings demand a non-empty, but a `string` is too wide. Cast it as `[AuditSlug, ...AuditSlug[]]` instead, then Zod will infer the union type.

```suggestion
import type { AuditSlug } from './types.js';

const auditSlugs = AUDITS.map(({ slug }) => slug) as [
  AuditSlug,
  ...AuditSlug[],
];
export const typescriptPluginConfigSchema = z.object({
  tsconfig: z
    .string({
      description: 'Path to a tsconfig file (default is tsconfig.json)',
    })
    .default(DEFAULT_TS_CONFIG),
  onlyAudits: z
    .array(z.enum(auditSlugs), {
      description: 'Filters TypeScript compiler errors by diagnostic codes',
    })
    .optional(),
});

export type TypescriptPluginOptions = z.input<
  typeof typescriptPluginConfigSchema
>;
export type TypescriptPluginConfig = z.infer<
  typeof typescriptPluginConfigSchema
>;
```

(Included small improvements to the descriptions.)

packages/plugin-typescript/src/lib/schema.unit.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it('accepts a configuration with tsConfigPath and empty onlyAudits', () => {
+    expect(() =>
+      typescriptPluginConfigSchema.parse({
+        tsConfigPath,
+        onlyAudits: [],
+      } satisfies TypescriptPluginOptions),
+    ).not.toThrow();
+  });
```
matejchalk, line 29: These tests are out of sync with the schema - expects `tsconfig`, not `tsConfigPath`.

packages/plugin-typescript/src/lib/typescript-plugin.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { AuditSlug } from './types.js';
+import { getAudits, getGroups, logSkippedAudits } from './utils.js';
+
+const packageJson = createRequire(import.meta.url)(
+  '../../package.json',
+) as typeof import('../../package.json');
+
+export type FilterOptions = { onlyAudits?: AuditSlug[] };
+export type TypescriptPluginOptions = Partial<DiagnosticsOptions> &
+  FilterOptions;
```
matejchalk, line 17: As mentioned before, these types are duplicated and can be replaced with the inferred schema types.

packages/plugin-typescript/src/lib/typescript-plugin.ts
```diff
@@ -0,0 +1,58 @@
...
+  tsPluginOptions: TypescriptPluginOptions,
+): TypescriptPluginOptions {
+  try {
+    return typescriptPluginConfigSchema.parse(tsPluginOptions) as FilterOptions;
+  } catch (error) {
+    throw new Error(
+      `Error parsing TypeScript Plugin options: ${stringifyError(error)}`,
+    );
+  }
+}
```
matejchalk, line 58: The inferred schema types should be used here, then no casting is needed:

```suggestion
function parseOptions(
  tsPluginOptions: TypescriptPluginOptions,
): TypescriptPluginConfig {
  try {
    return typescriptPluginConfigSchema.parse(tsPluginOptions);
  } catch (error) {
    throw new Error(
      `Error parsing TypeScript Plugin options: ${stringifyError(error)}`,
    );
  }
}
```

packages/plugin-typescript/src/lib/schema.ts
BioPhoton: Thanks for the explanation! 

I did not know about `z.input`, pretty helpful API. Will apply the suggestion!

packages/plugin-typescript/src/lib/typescript-plugin.ts
BioPhoton: removed

matejchalk (approved):

PR #937, opened 2025-02-17
fix(nx-plugin): adjust upload config handling

BioPhoton:
This PR includes:
- a fix to only use upload data id the API key is set
- updated unit tests
- updated e2e tests


matejchalk:
@BioPhoton The PR description only mentions a small adjustment to upload config, but most of the changed files are removing extensions from imports. Could you please remove those changes, or explain why they are needed?

I remember you mentioned something about import extensions before, but it's never made sense to me why that would solve anything. Unlike imports without extensions (which are CJS-only), explicit extensions work for both CJS and ESM. For my own sanity, I tried reverting the imports in ba918503a2216d3b89f943f01893d1a34eaf2812 and all `nx-plugin` tests still pass locally and in CI (see #972). So what is the reason for changing the imports?

PR #938, opened 2025-02-17
feat(ci): add skipComment option, ensure head ref is checked out

matejchalk:
- Runs `git checkout -f ${refs.head.ref}` in case set to other than current HEAD. This is a preparation for new custom source ref GitLab input.
- New `skipComment` option added (`false` by default), so that PR/MR comment can be turned off easily. Will follow-up with exposing this as input for GitLab and GitHub.


getlarge (approved):
Thank you!

vmasek (approved):

PR #942, opened 2025-02-19
feat(ci): disable nx/turbo cache for non-autorun code-pushup commands

matejchalk:
Monorepos like Nx and Turborepo support task caching. Since collecting a Code PushUp report may be slow, it makes sense to configure caching for the `code-pushup` task, with report files as the cache output. 

However, the CI integration also uses other CLI commands - `print-config`, `compare` and `merge-diffs` - which have different outputs. It would be more complex to configure cache outputs for these commands as well, and because these commands are typically only run "under the hood", we can't expect a user to know about them. Furthermore, these commands are much faster than `collect`/`autorun`, so caching them doesn't bring much benefit.

For the reasons outlined above, I've added CLI arguments to these commands that explicitly opt out of caching - `--skip-nx-cache` for Nx, `--no-cache --force` for Turborepo.


vmasek (approved):
Makes perfect sense 🚀

PR #950, opened 2025-02-26
feat(ci): allow refs without shas, fetch if needed

matejchalk:
Should unblock GitHub Action release (currently [fails](https://github.com/code-pushup/github-action/actions/runs/13545884942/job/37857864118#step:5:51) in https://github.com/code-pushup/github-action/pull/145).

### DIFF
--- a/packages/utils/src/lib/transform.ts
+++ b/packages/utils/src/lib/transform.ts
@@ -151,3 +151,11 @@ export function toOrdinal(value: number): string {
 
   return `${value}th`;
 }
+
+export function removeUndefinedAndEmptyProps<T extends object>(obj: T): T {
+  return Object.fromEntries(
+    Object.entries(obj).filter(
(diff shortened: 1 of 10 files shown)


packages/utils/src/lib/transform.ts
```diff
@@ -151,3 +151,11 @@ export function toOrdinal(value: number): string {
 
   return `${value}th`;
 }
+
+export function removeUndefinedAndEmptyProps<T extends object>(obj: T): T {
+  return Object.fromEntries(
+    Object.entries(obj).filter(
+      ([_, value]) => value !== undefined && value !== '',
```
vmasek, line 158: I think this is sufficient, but that's nitpick
```suggestion
      ([, value]) => value !== undefined && value !== '',
```

vmasek (dismissed):

vmasek (approved):

PR #952, opened 2025-03-04
fix(ci): always show execute process errors and log stdout if verbose

vmasek:
part of #888


packages/ci/src/lib/monorepo/list-projects.ts
```diff
@@ -95,16 +96,16 @@ function createMonorepoHandlerOptions(
...
         onStdout: stdout => {
           console.info(stdout);
         },
-        onStderr: stderr => {
-          console.warn(stderr);
-        },
-      },
-    }),
+      }),
+    },
```
matejchalk, line 108: This logic is duplicated in `createCommandContext`. How about extracting the observer creation?

packages/ci/src/lib/run.integration.test.ts
```diff
@@ -620,7 +692,7 @@ describe('runInCI', () => {
         diffMdString = await readFile(fixturePaths.diffs.merged.md, 'utf8');
       });
 
-      it('should collect and compare reports for all projects and comment merged diff', async () => {
+      it('event should collect and compare reports for all projects and comment merged diff', async () => {
```
matejchalk, line 695: This test renaming looks like a mistake, the sentence doesn't make sense:

```suggestion
      it('should collect and compare reports for all projects and comment merged diff', async () => {
```

vmasek: It is, thanks

matejchalk (approved):
LGTM :+1:

PR #953, opened 2025-03-04
fix(utils): remove partial from mergeConfigs return type

matejchalk:
Since the first parameter of `mergeConfigs` is a full `CoreConfig`, only the subsequent parameters are `Partial<CoreConfig>`s, then the result must be a full `CoreConfig`.

This removes the need to do `mergeConfigs(...) as CoreConfig` when using it.


vmasek (approved):
makes perfect sense, I do not recall why I left result partial (probably because all params were partial in initial design)

PR #954, opened 2025-03-04
fix: update nx-verdaccio pkg

BioPhoton:


vmasek (approved):
Does this release automatically skip install or should I try that myself?

PR #966, opened 2025-03-05
fix(plugin-js-packages): ignore non-json lines and warnings in stderr

matejchalk:
Ran into problem with JS packages when Node prints a deprecation warning to output. It doesn't have any effect on the audit and outdated commands, but it caused error nonetheless. Now such unrelated warnings should be ignored.


vmasek (approved):

vmasek (approved):
I think it is the same failing test, when it was happening to me I remember reading something about nx graph having some issues in nested git repositories (which we do have in test fixtures).

PR #971, opened 2025-03-09
feat: mvp nx-plugin with v2 format

aramirezj:


packages/nx-plugin/src/plugin/utils.ts
```diff
@@ -1,15 +1,15 @@
...
 import * as path from 'node:path';
-import { CP_TARGET_NAME } from './constants.js';
+import { CP_TARGET_NAME } from './constants';
 import type {
   CreateNodesOptions,
   NormalizedCreateNodesContext,
   ProjectConfigurationWithName,
 } from './types.js';
 
 export async function normalizedCreateNodesContext(
```
BioPhoton, line 11: We could add a unit test for a vs object

BioPhoton (changes requested):
Nice! Soon we have it working for latest Nx.
We could add a test for the normalize function.

vmasek:
From what I understood we are keeping both legacy and V2 for backwards compatibility.

Looks good so far, I'll watch this as this is now to only blocker for moving to latest Nx.
It will also unlock latest vitest which I'm looking forward updating to 👍🏻

BioPhoton:
> From what I understood we are keeping both legacy and V2 for backwards compatibility.
> 
> Looks good so far, I'll watch this as this is now to only blocker for moving to latest Nx. It will also unlock latest vitest which I'm looking forward updating to 👍🏻

There is one thing left and that is considering `projec.json` **AND** `package.json`

packages/nx-plugin/src/plugin/utils.ts
```diff
@@ -1,15 +1,22 @@
-import type { CreateNodesContext } from '@nx/devkit';
+import type { CreateNodesContext, CreateNodesContextV2 } from '@nx/devkit';
 import { readFile } from 'node:fs/promises';
 import * as path from 'node:path';
-import { CP_TARGET_NAME } from './constants.js';
+import { CP_TARGET_NAME } from './constants';
```
BioPhoton, line 4:
```suggestion
import { CP_TARGET_NAME } from './constants.js';
```

packages/nx-plugin/src/plugin/types.ts
```diff
@@ -1,6 +1,10 @@
-import type { CreateNodesContext, ProjectConfiguration } from '@nx/devkit';
+import type {
+  CreateNodesContext,
+  CreateNodesContextV2,
+  ProjectConfiguration,
+} from '@nx/devkit';
 import type { WithRequired } from '@code-pushup/utils';
-import type { DynamicTargetOptions } from '../internal/types.js';
+import type { DynamicTargetOptions } from '../internal/types';
```
BioPhoton, line 7:
```suggestion
import type { DynamicTargetOptions } from '../internal/types.js';
```

BioPhoton (changes requested):
added changes for file extensions

BioPhoton (dismissed):

e2e/nx-plugin-e2e/tests/plugin-derive-config.e2e.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  beforeEach(async () => {
+    tree = await generateWorkspaceAndProject();
+    registerPluginInWorkspace(tree, '@code-pushup/nx-plugin');
+    await generateProject(tree, projectName);
+    root = readProjectConfiguration(tree, projectName).root;
+    generateCodePushupConfig(tree, root);
+  });
+
+  afterEach(async () => {
+    // await teardownTestFolder(testFileDir);
```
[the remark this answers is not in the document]
aramirezj: oops

aramirezj:
All done! Ready to review again

BioPhoton:
The e2e tests fail in the `nx-plugin` this does not seem like flakeiness 

Screenshot 2025-03-24 at 11 25 19

BioPhoton (changes requested):
The e2e tests fail in the `nx-plugin` this does not seem like flakeiness 

Screenshot 2025-03-24 at 11 25 19

PR #972, opened 2025-03-10
[DO NOT MERGE] Nx plugin fix extensions

matejchalk:


BioPhoton:
@matejchalk we ran into the same error. What was the fix to not remove .js from all imports?

PR #973, opened 2025-03-10
refactor(utils): make execute process log compact

### LINKED ISSUE #967: More compact execute process logs
Currently, each executed process logs 4 lines:

Image

It would be more compact to just print the command and only print the working directory if it's different to current (and relative path would be shorter than absolute). 

A command run in CWD would then have one line like this.

```
$ eslint src
```

A command run in some other directory would be logged like this:

```
src $ eslint .
```

vmasek:
closes #967

### DIFF
--- a/packages/utils/src/lib/format-command-log.integration.test.ts
+++ b/packages/utils/src/lib/format-command-log.integration.test.ts
@@ -1,0 +1,61 @@
+import path from 'node:path';
+import { describe, expect, it } from 'vitest';
+import { removeColorCodes } from '@code-pushup/test-utils';
+import { formatCommandLog } from './format-command-log.js';
+
+describe('formatCommandLog', () => {
+  it('should format simple command', () => {
+    const result = removeColorCodes(
+      formatCommandLog('npx', ['command', '--verbose']),
+    );
+
+    expect(result).toBe('$ npx command --verbose');
+  });
+
+  it('should format simple command with explicit process.cwd()', () => {
+    const result = removeColorCodes(
+      formatCommandLog('npx', ['command', '--verbose'], process.cwd()),
+    );
+
+    expect(result).toBe('$ npx command --verbose');
+  });
+
+  it('should format simple command with relative cwd', () => {
+    const result = removeColorCodes(
+      formatCommandLog('npx', ['command', '--verbose'], './wololo'),
+    );
+
+    expect(result).toBe(`wololo $ npx command --verbose`);
+  });
+
+  it('should format simple command with absolute non-current path converted to relative', () => {
+    const result = removeColorCodes(
+      formatCommandLog(
+        'npx',
+        ['command', '--verbose'],
+        path.join(process.cwd(), 'tmp'),
+      ),
+    );
+    expect(result).toBe('tmp $ npx command --verbose');
+  });
+
+  it('should format simple command with relative cwd in parent folder', () => {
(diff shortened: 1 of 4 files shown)


packages/utils/src/lib/format-command-log.integration.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  it('should format simple command with relative cwd', () => {
+    const result = removeColorCodes(
+      formatCommandLog('npx', ['command', '--verbose'], './wololo'),
+    );
+
+    expect(result).toBe(
+      `${path.join(process.cwd(), 'wololo')} $ npx command --verbose`,
+    );
+  });
```
matejchalk, line 31: This is the opposite of what I would expect :sweat_smile: To reduce verbosity, the relative path should be printed, not the full path. In fact, if `cwd` is an absolute path, then it should be reduced to a relative path. Expanding a relative path to an absolute path (what's being tested here) is counter-productive.

matejchalk (changes requested):

packages/utils/src/lib/format-command-log.integration.test.ts
vmasek: I see, I'll switch the behavior.

packages/utils/src/lib/format-command-log.integration.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it('should format simple command with relative cwd', () => {
+    const result = removeColorCodes(
+      formatCommandLog('npx', ['command', '--verbose'], './wololo'),
+    );
+
+    expect(result).toBe(`wololo $ npx command --verbose`);
+  });
+
```
matejchalk, line 30: I'm missing a test case with an absolute path which isn't equal to CWD:

```suggestion

  it('should format simple command with absolute non-current path converted to relative', () => {
    const result = removeColorCodes(
      formatCommandLog(
        'npx',
        ['command', '--verbose'],
        path.join(process.cwd(), 'tmp')),
    );
    expect(result).toBe('tmp $ npx command --verbose');
  });

```

packages/utils/src/lib/format-command-log.integration.test.ts
```diff
@@ -0,0 +1,54 @@
...
+        'npx',
+        ['command', '--verbose'],
+        '/tmp/path/to/somewhere',
+      ),
+    );
+
+    expect(result).toBe(
+      `${path.relative(process.cwd(), '/tmp/path/to/somewhere')} $ npx command --verbose`,
+    );
+  });
```
matejchalk, line 53: It's not very good practice to use expressions instead of literals in test assertions. Especially when the assertion reuses internal implementation, the tests end up being a self-fulfilling prophecy, instead of describing behaviour from the outside. Yes, "the function that returns `path.relative(...)` does indeed match `path.relative(...)`", obviously, but what does that mean? What _value_ does it actually return?

I think both of these tests could be merged into one, with a literal assertion:

```suggestion
  it('should format simple command using relative path to parent directory', () => {
    const result = removeColorCodes(
      formatCommandLog(
        'npx',
        ['command', '--verbose'],
        path.dirname(process.cwd())),
    );

    expect(result).toBe('.. $ npx command --verbose');
  });
```

matejchalk (approved):

PR #977, opened 2025-03-10
test: nx plugin cfg from pkg

BioPhoton:


BioPhoton:
closed as merge into #971

PR #991, opened 2025-04-25
docs(cli): document minimal NodeJS support

### LINKED ISSUE #864: Lack of documentation on minimum Node version
### What happened?

Hi!

I was running into the following error when testing out the CLI, after configuring the ESLint plugin:

```
npx code-pushup


TypeError: groups.toSorted is not a function
    at groupsFromRuleCategories  ....
```

I believe this stems from the use of the `Array.prototype.toSorted()` method, which is supported from Node v20+, whereas I was using Node v18.

There is no mention in the README about minimum node version, so it would be a good idea to add this as a pre-requisite in the "getting started" guide somewhere.

Even the GitHub Actions example config here: https://github.com/code-pushup/cli/wiki/Code-PushUp-integration-guide-for-Nx-monorepos#ci-automation is using Node v18 - so that should also be updated too.

### What steps did you take?

As above - run `npx code-pushup` after following the [getting started guide](https://github.com/code-pushup/cli/blob/main/packages/cli/README.md#getting-started).

"@code-pushup/cli": "^0.53.1",     "@code-pushup/eslint-plugin": "^0.53.1",

Linux

### Node version

18.18.2

### Relevant log output

TypeError: groups.toSorted is not a function
    at groupsFromRuleCategories  ....

vmasek:
closes #864 

- [x] add minimal node of 20 to engine of the `@code-pushup/cli` `package.json`
- [x] add badge to the Readme
- [x] fix bump node version in the example CI action to v22 and update the actions used to v4 from deprecated v3

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 
 [![version](https://img.shields.io/github/v/release/code-pushup/cli)](https://github.com/code-pushup/cli/releases/latest)
 [![release date](https://img.shields.io/github/release-date/code-pushup/cli)](https://github.com/code-pushup/cli/releases)
+![NodeJS support](https://img.shields.io/node/v/%40code-pushup%2Fcli)
 [![license](https://img.shields.io/github/license/code-pushup/cli)](https://opensource.org/licenses/MIT)
 [![commit activity](https://img.shields.io/github/commit-activity/m/code-pushup/cli)](https://github.com/code-pushup/cli/pulse/monthly)
 [![CI](https://github.com/code-pushup/cli/actions/workflows/ci.yml/badge.svg?branch=main)](https://github.com/code-pushup/cli/actions/workflows/ci.yml?query=branch%3Amain)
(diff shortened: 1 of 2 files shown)


packages/cli/package.json
```diff
@@ -41,6 +41,9 @@
   "bin": {
     "code-pushup": "./src/index.js"
   },
+  "engines": {
+    "node": ">=20"
+  },
```
matejchalk, line 46: Should we add this to all our packages? :thinking:

README.md
matejchalk, line 1: > fix bump node version in the example CI action to v22 and update the actions used to v4 from deprecated v3

@vmasek What is this referring to? It's from your PR description. I don't see any matching file change :raised_eyebrow:

matejchalk (approved):

packages/cli/package.json
matejchalk: > I assume you always need to use CLI to run anything

That's true. Users also import plugins directly and may import from utils as well (when building their own plugins, typically), but in the end they'll run it through the CLI.

PR #1034, opened 2025-07-14
feat(nx-plugin): update unit test and int test cache with test-setup cache

### LINKED ISSUE #1029: Nx caching inputs missing for all projects using `setupFiles` or `globalSetup` in `vitest.config.ts`
**Update:**
We have to re-evaluate the issue after https://github.com/code-pushup/cli/pull/1142 is merged

---

### What happened?

If I change something in a testing lib the packages using them over `setupFiles` or `globalSetup` in `vitest.config.ts` have a false cache hit. 

**The problem:** 
Project `models` als all other projects miss the dependencies  to testing files referenced in  `vitest.config.ts`.

If I change something in a testing lib the packages using them over `setupFiles` or `globalSetup` in `vitest.config.ts` do not have a cache hit.

**Possible Solutions:**
- add `implicitDependencies` to all relevant testing targets - quick but not fine-grained also the biggest downside is our build and any other target in the project would be affected. ⚠️
- add the used files from `vitest.config.ts` as target `inputs/namedInputs` -  fine-grained but only maintainable over custom plugin that adds the inputs. But as the used filse in `setupFiles` or `globalSetup` change it has to be dynamically detected.
- create a config project handling the setupFiles and is directly depending on `testing-setup` (**best**)
  - #1065

### What steps did you take?

1. ✅ Test the project `modles` in `packages/models` and verify cache hit
   
```shell
  # [existing outputs match the cache, left as is] <= ✅ this is correct. It should have a cache hit
   nx run models:unit-test
```
2. ✅ Apply any code change in the project `testing-setup` in folder `testing/setup` and save the change to your filesystem
3. ❌ Test the project `modles` in  `packages/models` and verify NO cache hit - `nx unit-test models`

```shell
  # [existing outputs match the cache, left as is] <= ❌ this is not correct. It should not have a cache hit
   nx run models:unit-test
```

_No response_

MacOS

### Node version

24

### Relevant log output

```shell

```

### Related information

- https://nx.dev/reference/core-api/devkit/documents/ImplicitDependency
- https://nx.dev/deprecated/global-implicit-dependencies
- https://nx.dev/reference/inputs
- https://nx.dev/recipes/running-tasks/configure-outputs

AndriiSiuta:
Closes #1029

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -16,16 +16,28 @@
     "e2e": {
       "dependsOn": ["^build"]
     },
-    "nxv-pkg-install": {
-      "parallelism": false
-    },
-    "@nx/vite:test": {
+    "unit-test": {
       "cache": true,
-      "inputs": ["default", "^production"],
+      "inputs": ["default", "^production", "production", "test-setup"],
+      "executor": "@nx/vite:test",
       "options": {
+        "configFile": "{projectRoot}/vitest.unit.config.ts",
         "passWithNoTests": true,
         "watch": false
       }
+    },
+    "int-test": {
+      "cache": true,
+      "inputs": ["default", "^production", "test-setup"],
+      "executor": "@nx/vite:test",
+      "options": {
+        "configFile": "{projectRoot}/vitest.int.config.ts",
+        "passWithNoTests": true,
+        "watch": false
+      }
+    },
+    "nxv-pkg-install": {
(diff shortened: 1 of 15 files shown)


BioPhoton:
Should this be in draft state?

PR #1052, opened 2025-08-07
chore: add pkg new job

BioPhoton:


packages/nx-plugin/project.json
```diff
@@ -9,11 +9,11 @@
...
-        "tsConfig": "packages/nx-plugin/tsconfig.lib.json",
+        "outputPath": "{projectRoot}/dist",
+        "main": "{projectRoot}/src/index.ts",
+        "tsConfig": "{projectRoot}/tsconfig.lib.json",
         "assets": [
-          "packages/nx-plugin/*.md",
+          "{projectRoot}/*.md",
           {
             "input": "./packages/nx-plugin/src",
             "glob": "**/!(*.ts)",
```
matejchalk, line 19: I guess only `options.assets` need to be added here? The rest is the same as in `nx.json`.

packages/plugin-coverage/project.json
```diff
@@ -4,27 +4,8 @@
...
   "projectType": "library",
   "targets": {
-    "build": {
-      "executor": "@nx/js:tsc",
-      "outputs": ["{options.outputPath}"],
-      "options": {
-        "outputPath": "dist/packages/plugin-coverage",
-        "main": "packages/plugin-coverage/src/index.ts",
-        "tsConfig": "packages/plugin-coverage/tsconfig.lib.json",
-        "additionalEntryPoints": ["packages/plugin-coverage/src/bin.ts"],
```
matejchalk, line 14: I would've thought `additionalEntryPoints` had to be preserved (same for `plugin-eslint` and `plugin-js-packages`) :thinking: But E2E tests are passing, so I guess they aren't needed anymore :grin: Possibly because we switched from ESBuild bundle to TSC non-bundle.

.github/workflows/ci.yml
```diff
@@ -141,3 +141,5 @@ jobs:
         run: npm ci
       - name: Build affected projects
         run: npx nx affected:build --parallel=3
+      - name: Test-release packages
+        run: npx pkg-pr-new publish "packages/models/dist" "packages/utils/dist" "packages/core/dist" "packages/cli/dist" "packages/create-cli/dist"  "packages/ci/dist" "packages/plugin-coverage/dist" "packages/plugin-eslint/dist" "packages/plugin-js-packages/dist" "packages/plugin-jsdocs/dist" "packages/plugin-lighthouse/dist" "packages/plugin-typescript/dist"
```
matejchalk, line 145: A few things:

1. Only affected packages are built, so not all dist outputs will be available for publish.
2. This is a long list of folders to maintain, would be better to list projects dynamically.
3. The folders look wrong, the outputs are in `dist/packages/{name}`, not `packages/{name}/dist`.

One solution would be to use `nx show projects` along with `jq` to generate the package folders:

```sh
# tested locally, example output:
# dist/packages/core dist/packages/cli dist/packages/nx-plugin dist/packages/create-cli
npx nx show projects --affected --projects="packages/*" --json | jq -r 'map("dist/packages/"+.) | join(" ")'
```

However, it's probably simpler to just use shell wildcard expansion

```sh
npx pkg-pr-new publish dist/packages/*
```

packages/nx-plugin/project.json
BioPhoton: true!

BioPhoton: Done!

packages/plugin-coverage/project.json
BioPhoton: > Possibly because we switched from ESBuild bundle to TSC non-bundle.
Yes that is the reason! Still good catch.

.github/workflows/ci.yml
matejchalk: > This is intentional to be more in line for Nx release. Also mentioned in the issue description.

Ah, I didn't notice that you also changed `build`s `outputPath` to `{projectRoot}/dist`. OK, then all good.

matejchalk (dismissed):

PR #1056, opened 2025-08-08
feat(ci): add searchCommits option to extend portal cache range

matejchalk:
Reuses `latestReport` resolver from GraphQL API for a more wide-ranging search for previous report in Portal. Useful for Nx users who always run with `--affected`, increases chances of cache hit for projects which don't change much. The range can be configured, maximum number of commits is 100, default is 10.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
-        "@code-pushup/portal-client": "^0.14.3",
+        "@code-pushup/portal-client": "^0.15.0",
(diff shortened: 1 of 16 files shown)


matejchalk:
[Failing Code PushUp job](https://github.com/code-pushup/cli/actions/runs/16833772062/job/47687910911?pr=1056#step:5:2762) is due to breaking changes in `@code-pushup/portal-client` update. It works in current branch, but fails when checking out main (we'd have to re-run `npm install` in target branch). Shouldn't reoccur once this PR is merged.

vmasek (approved):

PR #1061, opened 2025-08-12
chore: refactor `unit-test` and `int-test` targets

BioPhoton:
**This PR includes:**
- add `defaultTargets` for `unit-test`, `int-test`
- adjust all `unit-test`, `int-test` targets to generate coverage by default
- remove `--coverage.enabled` and `--skipNxCache` form CP targets
- make CP reuse coverage data from `unit-test`, `int-test`
- added `passWithNoTests` to all targets
- removes unused `int-test` targets

> [!note]
> This PR brings the execution time **from 19m down to max 9m** for the whole cp report. 🔥


examples/plugins/project.json
```diff
@@ -4,37 +4,18 @@
...
       "executor": "@nx/vite:test",
-      "outputs": ["{options.reportsDirectory}"],
       "options": {
-        "configFile": "examples/plugins/vitest.int.config.ts",
-        "reportsDirectory": "../../coverage/examples-plugins/int-tests"
+        "configFile": "{projectRoot}/vitest.int.config.ts",
+        "coverage": {
+          "enabled": true
+        }
       }
```
matejchalk, line 19: The configuration covered by `nx.json`, I believe. So the `project.json` could be simplified to `"int-test": {}`.

Same goes for all other projects, as far as I can tell. It looks like you only reduced the `unit-test` configs, but not the `int-test` configs.

matejchalk (commented):
Nice cleanup :+1:

examples/plugins/project.json
BioPhoton: Done

matejchalk (approved):

Issue #721, opened 2024-06-23
feat: add `scoreTarget ` as CIL feature for audits


BioPhoton:
### User story

Followup of #713 just for audits.

## Acceptance criteria

- [ ] https://github.com/code-pushup/cli/issues/713 
- [ ] the icons should be rendered correctly on all displays 
  - [ ] **terminal** -  The "✅" and "❌" characters cause layout issues in the terminal. I suggest "✓" and "✗" instead with chalk styling.
    e.g. `chalk.bold(chalk.green('✓'))` and `chalk.bold(chalk.red('✗'))`
  - [ ] **readme** -  the "✅" and "❌" characters are used
  - [ ] **JSON** - n/a
  - ~**portal** - custom icons are used and they are already fine~

```mermaid
graph TD

    #713 --> CLI_Preparation
    CLI_Preparation --> CLI_Client_Update
    Portal_Model_Adoption --> Portal_UI_Update
    Portal_Model_Adoption --> CLI_Client_Update
    CLI_Client_Update --> CLI_Plugin_Update
```

**PR1 - CLI preparation**
- [ ] extend `auditSchema` with `targetScore`
- [ ] rename `applyMaxScoreAboveThreshold` to `scoreAuditWithScoreTarget` and move it form `plugin-coverage` into `utils`
- [ ] adjust `executePlugin` to apply the new scoring logic after the normalization
- [ ] adopt `categoryToGQL` functions to convert the name back and the floating number back to boolean and mark it with a comment

**PR2 - Portal model adoption**
- [ ] accept a number from 0-1 for new property `targetScore`
- [ ] release portal-client

**PR3 - CLI portal-client update**
- [ ] adopt `categoryToGQL` functions for new type upload
- [ ] add `targetScore` styling to all places
  - [ ] for `audits`, `scoreTarget` means score is changed to perfect score, because it affects the category/group score calculation

**PR4 - CLI plugin updates**
- [ ] refactor `plugin-coverage` for the new `targetScore` shape 

**PR5 - Portal UI updates**
- [ ] adopt score card for new `targetScore` logic

PR #1076, opened 2025-08-21
feat: implement scoreTargets for audits

hanna-skryl:
Part of #721 (CLI preparation)

This PR implements the CLI preparation phase for the `scoreTargets` feature for audits.

- Extend the plugin configuration with the `scoreTargets` property (single number or per-audit mapping)
- Create a generic `scoreAuditsWithTarget` utility function in the `utils` package
- Apply the `scoreTargets` transformation in `executePlugin` after the plugin execution
- Add unit tests for the new functionality

The following items from the ticket are intentionally deferred:

1) "extend `auditSchema` with `targetScore`" - After careful consideration, it's been concluded that `scoreTarget` belongs at the plugin configuration level rather than individual audit definitions. This approach provides greater flexibility, allowing for either a single value applicable to all audits or a per-audit mapping.
2) "rename `applyMaxScoreAboveThreshold`" - This function remains in the coverage plugin unchanged. Migration to the new generic system is planned for PR4 to maintain backward compatibility.
3) Portal integration - Changes to`auditToGQL` are not needed until PR3, after the portal model adoption (PR2) is complete.

### DIFF
--- a/packages/core/src/lib/implementation/execute-plugin.unit.test.ts
+++ b/packages/core/src/lib/implementation/execute-plugin.unit.test.ts
@@ -128,7 +128,7 @@ describe('executePlugin', () => {
   it('should apply a single score target to all audits', async () => {
     const pluginConfig: PluginConfig = {
       ...MINIMAL_PLUGIN_CONFIG_MOCK,
-      scoreTarget: 0.8,
+      scoreTargets: 0.8,
       audits: [
         {
           slug: 'speed-index',
@@ -169,7 +169,7 @@ describe('executePlugin', () => {
   it('should apply per-audit score targets', async () => {
     const pluginConfig: PluginConfig = {
       ...MINIMAL_PLUGIN_CONFIG_MOCK, // returns node-version audit with score 0.3
-      scoreTarget: {
+      scoreTargets: {
         'node-version': 0.2,
       },
     };


BioPhoton (changes requested):
Left a small naming change

BioPhoton (approved):
LGTM

packages/models/src/lib/plugin-config.ts
```diff
@@ -18,31 +19,42 @@ export const pluginContextSchema = z
...
+  )
+  .optional();
+
+export type PluginScoreTarget = z.infer<typeof pluginScoreTargetSchema>;
+
 export const pluginDataSchema = z.object({
   runner: z.union([runnerConfigSchema, runnerFunctionSchema]),
   audits: pluginAuditsSchema,
   groups: groupsSchema,
+  scoreTarget: pluginScoreTargetSchema,
```
matejchalk, line 52: Unlike a `scoreTarget` in a category, this applies to many audits and can also have multiple values. I wonder if it may be better to use a slightly different name - e.g. `scoreTargets` (plural) or `auditScoreTarget` (namespaced). :thinking:

matejchalk (approved):

matejchalk (approved):

packages/models/src/lib/plugin-config.ts
hanna-skryl: Agree! I prefer `scoreTargets`, as it complements `pluginScoreTargetsSchema` well.

Issue #1068, opened 2025-08-16
Use artifact generation options in `plugin-eslint`


BioPhoton:
To leverage parallelisation and caching of external tools like Nx we have to adjust the eslint plugin.

**Preparations:**

- [x] Pass persist options to `runnerConfig.args` and the `runnerFunction` to be able to know the `Persist.outputDir` in the runner. This enables to write files under `${persist.outputDir}/${PLUGIN_SLUG}/xy.z`, similar as the `cache.write` option 
- [x] Make runner create `eslint-report.<count>.json`
  - [x] Use `--output-path` to create `.json` file (loosing terminal output so we also write before sending)
- [x] Add artifice artefact generation logic 
  - [x] Accept globs for `artificePaths`
- [x] Remove additional bin execution and switch to runnerFunction

**Implementation:**

- [x] Update docs
- [x] add e2e tests
- [x] Add artifact options to `plugin-eslint`  
- [x] `lint` target in projects and `nx.json` should use formatter
- [x] Use Nx `run-many` to generate `eslint-report` artifacts 
  - [x] directly consume artifacts: `artifactsPaths: ['packages/**/.eslint/eslint-report.json']`

**Optional:**
- [x] Add Custom reporter package 
  - [x] Add Custom reporter to code 
  - [x] `lint-formatter` target in projects and `nx.json`
  - [x] Use custom reporter in `plugint-eslint` CP config artifact command  
- ~move custom formatter into separate project `@code-pushup/eslint-formatter-multi`~ (moved to separate issue)

PoC: #1069 

Related:
- #1077 
- #1078 
- #1087
- #1089 
- #1090 
- #1106

PR #1078, opened 2025-08-21
chore: add lint report target

BioPhoton:
This PR includes:
- adjust existing `lint` target in `nx.json#defaultTargtets`
- add `lint-report` target `nx.json#defaultTargtets`
- add empty `lint-report` target to all projects

Related:
- #1068

### DIFF
--- a/code-pushup.preset.ts
+++ b/code-pushup.preset.ts
@@ -7,9 +7,10 @@ import coveragePlugin, {
   getNxCoveragePaths,
 } from './packages/plugin-coverage/src/index.js';
 import eslintPlugin, {
-  eslintConfigFromAllNxProjects,
   eslintConfigFromNxProject,
 } from './packages/plugin-eslint/src/index.js';
+import type { ESLintTarget } from './packages/plugin-eslint/src/lib/config.js';
+import { nxProjectsToConfig } from './packages/plugin-eslint/src/lib/nx/projects-to-config.js';
 import jsPackagesPlugin from './packages/plugin-js-packages/src/index.js';
 import jsDocsPlugin from './packages/plugin-jsdocs/src/index.js';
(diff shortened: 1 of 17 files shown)


code-pushup.preset.ts
```diff
@@ -156,14 +157,25 @@ export const jsDocsCoreConfig = (
...
+
 export const eslintCoreConfigNx = async (
   projectName?: string,
 ): Promise<CoreConfig> => ({
   plugins: [
     await eslintPlugin(
       await (projectName
         ? eslintConfigFromNxProject(projectName)
-        : eslintConfigFromAllNxProjects()),
+        : eslintConfigFromPublishableNxProjects()),
```
matejchalk, line 178: What's wrong with linting non-publishable projects? :thinking:

BioPhoton: I would only include projects that are also tracked in the monorepo setup to have a good comparison base.

matejchalk: Agreed, the same projects should be included in both standalone and monorepo modes. Question is whether to include every project in both, or only published packages in both. 

For simplicity, I'm leaning towards collecting all projects. It also gives a little bit more scale to play with - caching impact will be more noticable when more projects are involved.

BioPhoton: What about tagging all projects with quality levels:
- their-a - released code
- their-b - released code secondary
- their-c - tooling, testing, pers, code-pushup, 
- their-d - examples

Example for a `project.json`:
```json
{
    tags: [ "checks:their-a" ] 
}
```

matejchalk: One more thing to consider, as this PR is concerned with ESLint. The rules configured from `@code-pushup/eslint-config` include many adjustments for "test code", so we already distinguish prod and non-prod with regards to how strict our rule set is.

PR #1079, opened 2025-08-22
feat(ci): add helper function for parsing configPatterns from json string

matejchalk:
Exporting some utilities from `@code-pushup/ci` which will be useful for parsing new `configPatterns` (#1046) and `searchCommits` (#1056) options in [GitHub Action](https://github.com/code-pushup/github-action) and [GitLab pipelines template](https://gitlab.com/code-pushup/gitlab-pipelines-template).


packages/ci/src/lib/schemas.unit.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  it('should reject an invalid slug that cannot be fixed by interpolation', () => {
+    expect(() => interpolatedSlugSchema.parse('Invalid Slug!')).toThrow(
+      ZodError,
+    );
+  });
+
+  it('should reject a non-string value', () => {
+    expect(() => interpolatedSlugSchema.parse(123)).toThrow(ZodError);
```
BioPhoton, line 24: Could those 2 live in a each block. I understand the description is lost, just an idea.

packages/models/src/lib/implementation/constants.ts
```diff
@@ -1,5 +1,12 @@
...
 export const DEFAULT_PERSIST_OUTPUT_DIR = '.code-pushup';
 export const DEFAULT_PERSIST_FILENAME = 'report';
 export const DEFAULT_PERSIST_FORMAT: Format[] = ['json', 'md'];
+
+export const DEFAULT_PERSIST_CONFIG: Required<PersistConfig> = {
+  outputDir: DEFAULT_PERSIST_OUTPUT_DIR,
+  filename: DEFAULT_PERSIST_FILENAME,
+  format: DEFAULT_PERSIST_FORMAT,
+  skipReports: false,
+};
```
BioPhoton, line 12:
```suggestion
export const DEFAULT_PERSIST_SKIP_REPORT = false;

export const DEFAULT_PERSIST_CONFIG: Required<PersistConfig> = {
  outputDir: DEFAULT_PERSIST_OUTPUT_DIR,
  filename: DEFAULT_PERSIST_FILENAME,
  format: DEFAULT_PERSIST_FORMAT,
  skipReports: DEFAULT_PERSIST_SKIP_REPORT,
};
```

This way we can reuse it in other areas like we do with `DEFAULT_PERSIST_OUTPUT_DIR`

packages/ci/src/lib/schemas.unit.test.ts
matejchalk: I would keep them separate, the slug which cannot be fixed by interpolation is a special case. The number is more of a general sanity check.

BioPhoton (changes requested):
Thanks for adding this helper. Added a suggestion for a const for export.

vmasek (approved):

PR #1087, opened 2025-08-27
feat(core): pass config to plugin runner

BioPhoton:
This PR includes:
- passing perts of the persist options to runner logic
- adding a plugin slug as const for `plugin-eslint`


packages/core/src/lib/implementation/execute-plugin.unit.test.ts
```diff
@@ -383,8 +394,8 @@ describe('executePlugins', () => {
           {
             ...MINIMAL_PLUGIN_CONFIG_MOCK,
             runner: {
-              command: 'node',
-              args: ['-v'],
+              command: 'echo',
```
BioPhoton, line 397: More stable and we anyway dont want to test `node -v` logic it self

packages/core/src/lib/implementation/runner.ts
```diff
@@ -32,12 +34,13 @@ export type ValidatedRunnerResult = Omit<RunnerResult, 'audits'> & {
...
 export async function executeRunnerConfig(
   cfg: RunnerConfig,
+  config: Required<Pick<PersistConfig, 'outputDir'>>,
 ): Promise<RunnerResult> {
   const { args, command, outputFile, outputTransform } = cfg;
 
   const { duration, date } = await executeProcess({
     command,
-    args,
+    args: [...(args ?? []), ...objectToCliArgs(config)],
```
BioPhoton, line 43: Here could be potential for a more strict parsing. If we pass persist options directly to args it might cause unexpected issues.

PR #1091, opened 2025-08-27
chore: use @code-pushup/nx-plugin in repo

BioPhoton:
**Precondition:**
- [x] #1095
- [x] #1105
- [x] #1109 
- [x] #1157
- [x] #1158
- [x] #1159 
- [x] #1161 
- [x] #1189

**This PR includes:**
- [x] use plugin for main and cached targets
- [x] use executor on all targets
- [x] remove existing target configurations in projects
- [ ] add docs
- [x] add validation helper script

**Related:** 
closes #971

**TODO Run local code**

- https://github.com/nrwl/nx/blob/3d1e54481224e329865e77e019e2dadb75d1a9a7/packages/nx/src/plugins/js/utils/register.ts#L13-L28

>  tsx is a utility to run TypeScript files in node which is growing in popularity:
>  https://tsx.is
> 
>  Behind the scenes it is invoking node with relevant --require and --import flags.
> 
>  If the user is invoking Nx via a script which is being invoked via tsx, then we
>  do not need to register any transpiler at all as the environment will have already
>  been configured by tsx. In fact, registering a transpiler such as ts-node or swc
>  in this case causes issues.
> 
>  Because node is being invoked by tsx, the tsx binary does not end up in the final
>  process.argv and so we need to check a few possible things to account for usage
>  via different package managers (e.g. pnpm does not set process._ to tsx, but rather
>  pnpm itself, modern yarn does not set process._ at all etc.).

**Changes where tested by:**

1. `npm i https://pkg.pr.new/code-pushup/cli/@code-pushup/nx-plugin@1091`  
2. add plugin inc options to `nx.json`
```json
{
  "pluginsConfig": {
    "@code-pushup/nx-plugin": {
      "projectPrefix": "cli",
      "bin": "packages/cli/src/index.ts",
      "env": {
        "NODE_OPTIONS": "--import tsx",
        "TSX_TSCONFIG_PATH": "tsconfig.base.json"
      }
    }
  },
  "plugins": [
    {
      "plugin": "@code-pushup/nx-plugin",
      "options": {
        "targetName": "cp"
      }
    }
  ]
}
```


e2e/nx-plugin-e2e/tests/plugin-create-nodes.e2e.test.ts
```diff
@@ -229,7 +229,7 @@ describe('nx-plugin', () => {
   });
 
   it('should consider plugin option projectPrefix in executor target', async () => {
-    const cwd = path.join(testFileDir, 'configuration-option-bin');
+    const cwd = path.join(testFileDir, 'configuration-option-projectPrefix');
```
BioPhoton, line 232: folder name was a duplicate

package.json
```diff
@@ -20,7 +20,11 @@
   "private": true,
   "dependencies": {
     "@axe-core/playwright": "^4.11.0",
+    "@code-pushup/cli": "^0.96.1",
+    "@code-pushup/models": "^0.96.1",
+    "@code-pushup/nx-plugin": "^0.96.1",
```
BioPhoton: We have to install it because ATM Nx plugins don't take import aliases in any way. This means our nx-plugin code is always 1 release behind the current PR

PR #1107, opened 2025-09-05
refactor: common vitest configurations

poltorak:


e2e/ci-e2e/vitest.e2e.config.ts
```diff
@@ -1,22 +1,10 @@
...
-    alias: tsconfigPathAliases(),
-    pool: 'threads',
-    poolOptions: { threads: { singleThread: true } },
-    cache: {
-      dir: '../../node_modules/.vitest',
-    },
-    environment: 'node',
-    include: ['tests/**/*.e2e.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
-    globalSetup: './global-setup.ts',
-    setupFiles: ['../../testing/test-setup/src/lib/reset.mocks.ts'],
```
BioPhoton, line 20: Did you forgot to keep this?

packages/core/vitest.unit.config.ts
```diff
@@ -1,36 +1,15 @@
...
-      '../../testing/test-setup/src/lib/extend/ui-logger.matcher.ts',
-      '../../testing/test-setup/src/lib/extend/markdown-table.matcher.ts',
-      '../../testing/test-setup/src/lib/extend/jest-extended.matcher.ts',
-    ],
-  },
+export default createUnitConfig('core', {
+  projectRoot: new URL('../../', import.meta.url),
+  setupFiles: [
+    'testing/test-setup/src/lib/cliui.mock.ts',
+    'testing/test-setup/src/lib/fs.mock.ts',
```
BioPhoton, line 8: This can be default

packages/core/vitest.unit.config.ts
```diff
@@ -1,36 +1,15 @@
...
+    'testing/test-setup/src/lib/git.mock.ts',
+    'testing/test-setup/src/lib/portal-client.mock.ts',
+    'testing/test-setup/src/lib/extend/ui-logger.matcher.ts',
+    'testing/test-setup/src/lib/extend/markdown-table.matcher.ts',
+    'testing/test-setup/src/lib/extend/jest-extended.matcher.ts',
```
BioPhoton, line 13: this could also be in defaults

packages/utils/vitest.unit.config.ts
```diff
@@ -1,38 +1,17 @@
...
+  projectRoot: new URL('../../', import.meta.url),
+  include: ['src/**/*.{unit,type}.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
+  typecheckInclude: ['**/*.type.test.ts'],
+  setupFiles: [
+    'testing/test-setup/src/lib/cliui.mock.ts',
+    'testing/test-setup/src/lib/fs.mock.ts',
+    'testing/test-setup/src/lib/extend/ui-logger.matcher.ts',
+    'testing/test-setup/src/lib/extend/markdown-table.matcher.ts',
+    'testing/test-setup/src/lib/extend/path.matcher.ts',
+    'testing/test-setup/src/lib/extend/jest-extended.matcher.ts',
```
BioPhoton, line 14: Those could be defaults

tools/README.md
```diff
@@ -0,0 +1,154 @@
+## Vitest config factory and setup presets
+
+This folder contains utilities to centralize and standardize Vitest configuration across the monorepo.
```
BioPhoton, line 3:
```suggestion
This project contains utilities to centralize and standardize Vitest configuration across a monorepo.
```

tools/README.md
```diff
@@ -0,0 +1,154 @@
...
+### Files
+
+- `vitest-config-factory.ts`: builds typed Vitest configs with sensible defaults.
+- `vitest-setup-presets.ts`: reusable groups of `setupFiles` paths by test kind.
+
+### Goals
+
+- Reduce duplication across `vitest.*.config.ts` files.
+- Keep per-package intent clear with minimal overrides.
+- Provide safe defaults and easy extension points.
```
BioPhoton, line 14: We could remove it as its already mentioned in first paragraph

tools/README.md
```diff
@@ -0,0 +1,154 @@
...
+});
+```
+
+Creators:
+
+- `createUnitConfig(projectKey, { projectRoot, ...options })`
+- `createIntConfig(projectKey, { projectRoot, ...options })`
+- `createE2eConfig(projectKey, { projectRoot, ...options })`
+
+`projectKey` is used for cache and coverage directories.
```
BioPhoton, line 35: You could give an example

tools/README.md
```diff
@@ -0,0 +1,154 @@
...
+Global setup:
+
+- Unit/Int: `['<projectRoot>/global-setup.ts']` by default
+- E2E: none by default (set per-suite if needed)
+
+Include patterns:
+
+- Unit: `src/**/*.unit.test.*`
+- Int: `src/**/*.int.test.*`
+- E2E: `tests/**/*.e2e.test.*`
```
BioPhoton, line 61: This information is tightly coupled to our implementation. MAybe link the perset directly here instead of repeating?

tools/README.md
```diff
@@ -0,0 +1,154 @@
...
+- E2E baseline: `reset.mocks.ts`
+
+Extend with additional files using `options.setupFiles` — they append after the baseline (paths are project-root-relative):
+
+```ts
+export default createUnitConfig('core', {
+  projectRoot: new URL('../../', import.meta.url),
+  setupFiles: ['testing/test-setup/src/lib/cliui.mock.ts'],
+});
+```
```
BioPhoton, line 78: Same here, strongly coupled to implementation that can change often.

tools/README.md
```diff
@@ -0,0 +1,154 @@
...
+- `kind` ('unit' | 'int' | 'e2e'): test kind.
+- `projectRoot` (string | URL): absolute root for all paths.
+- `include?: string[]`: override default include globs.
+- `setupFiles?: string[]`: extra setup files (appended to baseline; project-root-relative).
+- `overrideSetupFiles?: boolean`: skip baseline and use only provided list.
+- `globalSetup?: string[]`: override default global setup (project-root-relative).
+- `coverage?: { enabled?, exclude?, reportsSubdir? }`
+- `testTimeout?: number`: e.g., for E2E.
+- `typecheckInclude?: string[]`: include patterns for Vitest typecheck.
+- `cacheKey?: string`: custom cache dir suffix.
```
BioPhoton, line 123: The best DX is given if the user does no need to lear another type. 

I would suggest try to keep the types as close as possible to existing virest types. 

e.g. :  `createIntConfig({/*vitestTestConfig*/}, { projectKey: 'unit', /*other custom options we need that are not already inside vitest types */})`

tools/vitest-config-factory.ts
```diff
@@ -0,0 +1,151 @@
...
+    reportsSubdir?: string;
+  };
+  testTimeout?: number;
+  typecheckInclude?: string[];
+  cacheKey?: string;
+}
+
+export function createVitestConfig(
+  options: CreateVitestConfigOptions,
+): ViteUserConfig {
```
BioPhoton, line 32: As mentioned in [this comment](https://github.com/code-pushup/cli/pull/1107/files#r2324898492) I would use a API and more important data structure/types as lose as possible to vitest.

BioPhoton (changes requested):
Thanks for cleaning up our test config mess :) 

I left some comments. 

Many thing I could add in addition is, we should have the project under `testing/vitest-setup`. I updated the issue accordingly

tools/vitest-config-factory.ts
```diff
@@ -0,0 +1,151 @@
...
+export type TestKind = 'unit' | 'int' | 'e2e';
+
+export interface CreateVitestConfigOptions {
+  projectKey: string;
+  kind: TestKind;
+  projectRoot: string | URL;
+  include?: string[];
+  setupFiles?: string[];
+  /** If true, the factory will not inject the baseline setupFiles for the given kind. */
+  overrideSetupFiles?: boolean;
```
BioPhoton, line 18: I would only apply vitest helper and the special merge logic is handled in each project.

testing/test-setup-config/src/lib/vitest-setup-presets.unit.test.ts
```diff
@@ -0,0 +1,331 @@
...
+    );
+  });
+
+  it('should merge other overrides correctly while using default setupFiles', () => {
+    const testTimeout =
+      kind === 'unit'
+        ? TEST_TIMEOUTS.MEDIUM
+        : kind === 'int'
+          ? TEST_TIMEOUTS.LONG
+          : TEST_TIMEOUTS.E2E;
```
BioPhoton, line 132: Let's avoid conditionals in tests. It makes the test code harder to read and adds noise.

testing/test-setup-config/src/lib/vitest-setup-presets.unit.test.ts
```diff
@@ -0,0 +1,331 @@
...
+    expect(typeof setupPresets).toBe('object');
+  });
+});
+
+// Parameterized tests to eliminate duplication
+describe.each([
+  ['unit', setupPresets.unit.base, createUnitConfig],
+  ['int', setupPresets.int.base, createIntConfig],
+  ['e2e', setupPresets.e2e.base, createE2eConfig],
+] as const)('%s config creation', (kind, baseSetupFiles, createFn) => {
```
BioPhoton, line 80: The each block for describe is introducing conditional checks in the it blocks as well as const. It further makes the tests harder to read. Let's avoid conditionals in tests as good as possible. 

If you want to test basic values you could have a separate it block using the each for static things like defaults that are always the same. In the specific cases you can use `expect.objectContaining` to keep is focused.

se [example for conditionaly](https://github.com/code-pushup/cli/pull/1107/files#r2403979638)

testing/test-setup-config/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+- `setupPresets.unit.{base, git, portalClient, matchersCore, matcherPath}`
+- `setupPresets.int.{base, cliui, fs, git, portalClient, matcherPath, chromePath}`
+- `setupPresets.e2e.{base}`
+
+### Parameters
+
+- `projectKey`: Used for cache and coverage directory naming
+- `projectRoot`: Required path/URL to the project root for resolving all paths
+- Standard Vitest configuration options can be provided in the overrides parameter
```
BioPhoton, line 34: This is covered by the types. You could add TSDocs and remove it here.

e2e/ci-e2e/vitest.e2e.config.ts
```diff
@@ -1,22 +1,17 @@
...
-      dir: '../../node_modules/.vitest',
+export default createE2eConfig(
+  'ci-e2e',
+  {
+    projectRoot: new URL('../../', import.meta.url),
+    cacheKey: 'ci-e2e',
+  },
+  {
+    test: {
+      testTimeout: 60_000,
```
BioPhoton, line 12: this could be a default for the e2e target

testing/test-setup-config/src/lib/vitest-config-factory.ts
```diff
@@ -0,0 +1,211 @@
...
+        ? [...defaultExclude, ...params.overrideExclude]
+        : defaultExclude,
+  };
+}
+
+function buildBaseConfig(params: {
+  projectKey: string;
+  kind: TestKind;
+  projectRootUrl: URL;
+  cacheDirName: string;
```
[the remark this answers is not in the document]
poltorak: Is it suggestion or request change 😄 ?
Deriving `cacheDirName` from `projectKey` in form: `cache-${projectKey}` sounds ok but makes me wonder, aren't we making some hidden bond between project key and cache directory? 

Maybe we could consider `cacheDirName` as optional which fallbacks to `cache-${projectKey}` if not specified?
Let me know @BioPhoton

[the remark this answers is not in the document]
BioPhoton: > Maybe we could consider cacheDirName as optional which fallbacks to ...
Yes better!

> cache-${projectKey} if not specified?  
the "cache-" was just something to give example. In general the cache key (and directory) should be unique across the repository. That is the reason we are about it in the first place.

e2e/ci-e2e/vitest.e2e.config.ts
BioPhoton: Hmmm.... I would put 20 as default. Or if there are more projects with 40 lets go with 40.
Did not realise this is already an overwrite..

poltorak: I checked that and:

| Timeout | Occurences |
----------|-------------
| 20s | 6 |
| 40s | 1 |
| 60s | 1 |
| 80s | 2 |
| 120s | 1 |

I suggest then use 20s as default (20_000ms)

testing/test-setup-config/src/lib/vitest-config-factory.ts
```diff
@@ -0,0 +1,211 @@
...
+export type VitestOverrides = ViteUserConfig & { test?: InlineConfig };
+export type ConfigRestParams = Pick<
+  VitestConfigFactoryOptions,
+  'projectRoot' | 'cacheKey'
+>;
+
+export function createVitestConfig(
+  options: VitestConfigFactoryOptions,
+  overrides: VitestOverrides = {},
+): ViteUserConfig {
```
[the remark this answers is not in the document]
poltorak: Please let me know which direction we should follow:
- Easier override handling ( current solution ) - in this case we take care of most of the overrides but on the other hand we are exposed to vitest changes (if the happen) and we need to keep up it them
- We go easy way and move transfer the "burden" of overrides to end consumer / user.

Both approaches have pros and cons. I don't feel like I'm the right person to make a decision here 😅

.gitignore
```diff
@@ -32,6 +32,7 @@ node_modules
 /.sass-cache
 /connect.lock
 /coverage
+**/.coverage/**
```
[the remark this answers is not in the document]
BioPhoton: This one could be used to exclude colocated coverage folders of each project.

tools/vitest-tsconfig-path-aliases.ts
```diff
@@ -1,8 +1,13 @@
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 import { loadConfig } from 'tsconfig-paths';
 import type { Alias, AliasOptions } from 'vite';
 
-export function tsconfigPathAliases(): AliasOptions {
-  const result = loadConfig('tsconfig.base.json');
+export function tsconfigPathAliases(projectRootUrl?: URL): AliasOptions {
```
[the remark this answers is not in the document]
poltorak: We actually need both for some time. The one from `tool` is used in other not migrated config files.
Once we migrate all configs then we can remove it.

testing/test-setup-config/src/lib/vitest-config-factory.unit.test.ts
```diff
@@ -0,0 +1,356 @@
...
+            ]),
+            coverage: expect.objectContaining({
+              reporter: ['text', 'lcov'],
+              reportsDirectory: '../../coverage/test-package/unit-tests',
+              exclude: ['mocks/**', '**/types.ts', 'perf/**'],
+            }),
+            typecheck: { include: ['**/*.type.test.ts'] },
+          }),
+        }),
+      );
```
matejchalk: No properties other than `cacheDir` and `test` should be created, so `expect.objectContaining` shouldn't be needed.

testing/test-setup-config/src/lib/vitest-config-factory.unit.test.ts
```diff
@@ -0,0 +1,356 @@
...
+
+      expect((config as any).test.typecheck).toEqual({
+        include: ['**/*.type.test.ts'],
+      });
+    });
+
+    it('should always include perf/** in coverage exclusions', () => {
+      const config = createVitestConfig('test-package', 'unit');
+
+      expect((config as any).test.coverage.exclude).toContain('perf/**');
```
matejchalk: There are a lot of unnecessary `config as any` castings in this test suite. To access properties under `config.test`, use `!` [^1] or `?.`.

[^1]: Non-null assertion with `!` isn't safe to use in general, but in tests it doesn't really matter - a runtime error is much the same as a failing test.

matejchalk (changes requested):
I think this is headed in the right direction. I also reviewed the internal implementation this time, where I have some minor suggestions.

However, I think the pattern is now fairly clear, so please go ahead and apply it to other Vitest configs. 🙏

packages/plugin-eslint/mocks/fixtures/nx-monorepo/package-lock.json
matejchalk: I don't believe this lockfile is necessary for running the tests. Please remove it.

matejchalk (commented):
Almost there 🤞

packages/ci/vitest.unit.config.ts
```diff
@@ -1,32 +1,17 @@
...
+      ...(config.test!.setupFiles || []),
       '../../testing/test-setup/src/lib/logger.mock.ts',
-      '../../testing/test-setup/src/lib/reset.mocks.ts',
-      '../../testing/test-setup/src/lib/extend/jest-extended.matcher.ts',
     ],
   },
-});
+};
+
+export default config;
```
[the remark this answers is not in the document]
matejchalk: Ah, sorry, you've done it already :grin:

PR #1109, opened 2025-09-05
feat(nx-plugin): adjust bin path handling

BioPhoton:
Precondition:
- #1113

This PR includes:
- allows us to execute the repositories code of: 
  - `cli` project'
  - `nx-plugin` projects executors (not the target inference)
- removes pluginBin option as it would only help product development and is not working with dependent bundles 
- more intuitive naming

> [!note]
> After this is merged is should be possible to use it locally like this:
> 
> _nx.json.defaultTargets
> ```json
> "code-pushup": {
>       "dependsOn": ["cli:build", "nx-plugin:build"],
>       "executor": "@copde-pushup/nx-plugin:cli",
>       "options": {
>         "bin": "./packages/cli/dist",
>         "env": {
>           "NODE_OPTIONS": "--import tsx",
>           "TSX_TSCONFIG_PATH": "tsconfig.base.json"
>         }
>     }
> }
> ```
> or
>
> _nx.json.plugins_
> ```json
> {
>       "plugin": "@code-pushup/nx-plugin",
>       "options": {
>         "bin": "./packages/cli/dist",
>         "env": {
>           "NODE_OPTIONS": "--import tsx",
>           "TSX_TSCONFIG_PATH": "tsconfig.base.json"
>         }
>     }
> }
> ```
> Then run: `nx code-pushup`
> 
> You will see a log that prints the executed command as copy past version for you to re-ensure the setup.
> `NODE_OPTIONS="--import tsx" TSX_TSCONFIG_PATH="tsconfig.base.json" npx @code-pushup/cli --verbose --no-progress --config="code-pushup.config.ts"  --upload.project="cli-workspace"`
> 
> The only thing that is left but blocked by Nx, is using the local target inference code with external dependencies.    
    
Another precondition of: 
- #1091

### DIFF
--- a/packages/nx-plugin/src/plugin/target/executor-target.ts
+++ b/packages/nx-plugin/src/plugin/target/executor-target.ts
@@ -9,9 +9,9 @@ export function createExecutorTarget(options?: {
   const { bin, projectPrefix } = options ?? {};
 
   const executor = `${PACKAGE_NAME}:cli`;
-  const options = (bin || projectPrefix) && {    
+  const executorOptions = (bin || projectPrefix) && {
     ...(bin && { bin }),
     ...(projectPrefix && { projectPrefix }),
   };
-  return { executor, ...(options && { options })};
+  return { executor, ...(executorOptions && { options: executorOptions }) };
 }


packages/nx-plugin/src/internal/versions.ts
```diff
@@ -1,25 +1,4 @@
...
- * Load the package.json file from the given folder path.
- * @param folderPath
- */
-function loadPackageJson(folderPath: string): PackageJson {
-  return readJsonFile<PackageJson>(path.join(folderPath, 'package.json'));
-}
+export const cpNxPluginVersion = 'latest';
+export const cpModelVersion = 'latest';
+export const cpUtilsVersion = 'latest';
+export const cpCliVersion = 'latest';
```
BioPhoton, line 4: Here I'm unsure what the best replacement for accessing other packages is...

CONTRIBUTING.md
```diff
@@ -69,6 +69,58 @@ You can control the execution of long-running tests over the `INCLUDE_SLOW_TESTS
 To change this setup, open (or create) the `.env` file in the root folder.
 Edit or add the environment variable there as follows: `INCLUDE_SLOW_TESTS=true`.
 
+### Executing local code
+
+We use the current version of CodePushup and its plugins, to measure itself and its plugins.
```
matejchalk, line 74: Branding :sweat_smile: 

```suggestion
We use the current version of Code PushUp and its plugins, to measure itself and its plugins.
```

CONTRIBUTING.md
```diff
@@ -69,6 +69,58 @@ You can control the execution of long-running tests over the `INCLUDE_SLOW_TESTS
...
+    "exec-local-cli-source-and-local-plugin-source": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "node packages/cli/src/index.ts",
+        "args": ["--no-progress", "--verbose", "--onlyPlugins=js-packages"],
+        "env": {
+          "NODE_OPTIONS": "--import tsx",
+          "TSX_TSCONFIG_PATH": "tsconfig.base.json"
+        }
+      }
```
matejchalk, line 98: A couple of things:
- the `--help` in `exec-local-cli-source` target would not run any commands, probably not what you want
- `exec-local-cli-source` is missing the `tsx` env vars, so it wouldn't work
- it would be more convenient to have 1 reusable target than 2 specialized targets (won't scale), if you just expose the CLI, then any arguments can be appended (e.g. `-- --onlyPlugins=js-packages`)
- the target names are really long - why not just call the target `code-pushup`? that's what it's always called, and it does the same thing (just from source instead of npm)
 
```suggestion
    "code-pushup": {
      "executor": "nx:run-commands",
      "options": {
        "command": "node packages/cli/src/index.ts",
        "args": ["--no-progress", "--verbose"],
        "env": {
          "NODE_OPTIONS": "--import tsx",
          "TSX_TSCONFIG_PATH": "tsconfig.base.json"
        }
      }
```

CONTRIBUTING.md
```diff
@@ -69,6 +69,58 @@ You can control the execution of long-running tests over the `INCLUDE_SLOW_TESTS
...
+```jsonc
+// nx.json
+{
+  "plugins": [
+    {
+      "plugin": "@code-pushup/nx-plugin",
+      "options": {
+        "cliBin": "node ./packages/cli/src/index.ts",
+        "env": {
+          // nx.json
```
matejchalk, line 114: The `nx.json` comment is duplicated:

```suggestion
```

packages/nx-plugin/src/executors/cli/executor.ts
```diff
@@ -30,29 +32,30 @@ export default async function runAutorunExecutor(
...
+      ...(context.cwd ? { cwd: context.cwd } : {}),
+      env,
+      dryRun,
+    });
+  } catch (error) {
+    logger.error(error);
+    return {
+      success: false,
+      command: commandString,
+      error: error as Error,
```
matejchalk, line 55: I realize this wasn't introduced in this PR, but it's been bugging me :sweat_smile: Casting the error is unsafe, and given we have our `stringifyError` function, safe error handling should be easy:

```suggestion
      error: error instanceof Error ? error : new Error(stringifyError(error)),
```

packages/nx-plugin/package.json
```diff
@@ -37,7 +37,8 @@
     "@nx/devkit": ">=17.0.0",
     "ansis": "^3.3.0",
     "nx": ">=17.0.0",
-    "zod": "^4.0.5"
+    "zod": "^4.0.5",
+    "chalk": "5.3.0"
```
matejchalk, line 41: We already have `ansis` for terminal colors, please use it instead of introudcing a new dependency.

```suggestion
    "zod": "^4.0.5"
```

packages/nx-plugin/src/internal/execute-process.ts
```diff
@@ -138,40 +159,71 @@ export type ProcessObserver = {
...
-  const logCommand = [command, ...(args || [])].join(' ');
   ui().logger.log(
     gray(
-      `Executing command:\n${logCommand}\nIn working directory:\n${cfg.cwd ?? process.cwd()}`,
+      `Executing command:\n${formatCommandLog({
+        command,
+        args: args ?? [],
+        env,
+      })}\nIn working directory:\n${cfg.cwd ?? process.cwd()}`,
     ),
```
matejchalk, line 182: The `executeProcess` implemented in `utils` now has more compact command formatting that's easier to read.
- #967

https://github.com/code-pushup/cli/blob/c7d314e8c835b0332d3b0ddd1407987621d45179/packages/utils/src/lib/execute-process.ts#L154-L158

https://github.com/code-pushup/cli/blob/c7d314e8c835b0332d3b0ddd1407987621d45179/packages/utils/src/lib/format-command-log.ts#L12-L27

I suggest copying it here, so we don't reinvent the wheel and have consistent command logs across all our tools.

packages/nx-plugin/src/executors/internal/cli.ts
```diff
@@ -12,15 +13,49 @@ export function createCliCommandString(options?: {
...
+    logElements.push(...envVars);
+  }
+  // eslint-disable-next-line functional/immutable-data
+  logElements.push(chalk.cyan(command));
+  if (args.length > 0) {
+    // eslint-disable-next-line functional/immutable-data
+    logElements.push(chalk.dim.gray(args.join(' ')));
+  }
+  return logElements.join(' ');
+}
```
matejchalk, line 41: Following up on previous comment, there's a `formatCommandLog` in `utils` that could be copied over, so we don't diverge too much. I'd also prefer to stick to an immutable coding style - i.e. conditional array spread instead of mutating the array.

However, I see your version has an extra feature - prefixing environment variables :+1: Do you think you add that to the upstream implementation, then copy it over here? :slightly_smiling_face:

I'd prefer to maintain one robust implementation (albeit copy-pasted), rather than two diverging implementations which are incomplete in different ways.

packages/nx-plugin/src/executors/internal/cli.ts
```diff
@@ -12,15 +13,49 @@ export function createCliCommandString(options?: {
...
+
 export function createCliCommandObject(options?: {
   args?: Record<string, unknown>;
-  command?: string;
+  command?: 'autorun' | 'collect' | 'upload' | 'print-config' | string;
```
matejchalk, line 45: TypeScript resolves this as a string, I'm afraid.

image

But, there's a trick you can use get a loose autocomplete type, if that's what you want.

image

matejchalk (changes requested):

packages/nx-plugin/src/executors/internal/cli.ts
BioPhoton: Ah!! I was thinking of this hack! Seems I forgot. Nice reminder of this.

packages/nx-plugin/src/plugin/target/executor.target.unit.test.ts
```diff
@@ -9,8 +9,13 @@ describe('createExecutorTarget', () => {
   });
 
   it('should use bin if provides', () => {
```
matejchalk:
```suggestion
  it('should use bin if provided', () => {
```

PR #1138, opened 2025-11-07
feat(cli,core,utils): replace basic logs with new logger

matejchalk:
Part of #888

Replaces basic `@poppinss/cliui` logs with new `logger` in the core packages - `cli`, `core`, `utils`.

I'll tackle logs in plugins and the more complex `@poppinss/cliui` logs (tables, stickers) in other PRs.


packages/cli/src/lib/collect/collect-command.ts
```diff
@@ -40,12 +40,13 @@ export function yargsCollectCommandObject(): CommandModule {
 }
 
 export function renderUploadAutorunHint(): void {
+  // TODO: replace @poppinss/cliui
```
BioPhoton: Is this part of a followup PR or leftover?

PR #1142, opened 2025-11-09
refactor: fix cache invalidation for tests in monorepo setups

### LINKED ISSUE #1102: Invalidate test cache when fixtures change
Each test target - `unit-test`, `int-test`, `e2e-test` - should be invalidated whenever the tested code changes, or the tests themselves change. Currently, the cache isn't being invalidated when fixtures change - i.e. files which aren't `import`ed in code, but are read from file system with `node:fs` module (usually copied to a temporary test folder in `beforeEach`/`beforeAll`).

Because this, we run the risk of merging broken tests, which will be hard to track down.

## Steps to reproduce

1. Run `npx nx int-test plugin-eslint`. Tests will pass and Nx will save cache.
2. Change any file under `packages/plugin-eslint/mocks/fixtures` in a way that should break some test (e.g. delete `nx-monorepo/nx.json`).
3. Re-run `npx nx int-test plugin-eslint`. Tests will pass instantly, Nx logs that cache was used.
4. Run `npx nx int-test plugin-eslint --skipNxCache`. Observe that tests fail.

This is repeatable of any other type of test that relies on file system - i.e. some integration tests, most E2E tests.

## Acceptance criteria

- [ ] Any change to files under `{projectRoot}` should invalidate test targets in Nx cache.

BioPhoton:
Remove folders from `.nxignore` and rename files to not get detected by Nx

~~Followup PR:~~
- ~~one last folder is remaining under `packages/plugin-eslint/mocks/fixtures/nx-monorepo`~~

closes  #1102

### DIFF
--- a/testing/test-utils/src/lib/utils/test-folder-setup.unit.test.ts
+++ b/testing/test-utils/src/lib/utils/test-folder-setup.unit.test.ts
@@ -35,6 +35,7 @@ describe('restoreNxIgnoredFiles', () => {
       '/projects/lib2/project.json': '',
     });
   });
+
   it('should rename Nx ignored files in a folder', async () => {
     vol.fromJSON(
       {


e2e/ci-e2e/mocks/setup.ts
```diff
@@ -28,6 +30,7 @@ export async function setupTestRepo(folder: string) {
   );
 
   await cp(fixturesDir, baseDir, { recursive: true });
+  await restoreRenamedFiles(baseDir, NX_IGNORED_FILES_TO_RESTORE);
```
matejchalk: I see the second argument is always set to `NX_IGNORED_FILES_TO_RESTORE` constant in each usage. How about making it the default value? Or hard-coding it even? It would make usages less verbose.

testing/test-utils/src/lib/utils/test-folder-setup.ts
```diff
@@ -36,3 +37,88 @@ export async function teardownTestFolder(dirName: string) {
...
+        try {
+          await rename(fullPath, newPath);
+        } catch (error) {
+          // Ignore errors if file doesn't exist or can't be renamed
+        }
+      }
+    }
+  } catch (error) {
+    // Ignore errors if directory doesn't exist
+  }
```
matejchalk: Are you expecting these to fail? The paths are from `readdir`, so I would expect it's safe to traverse and rename any of these files. If there was an error, I'd be interested to know what it was, rather than silencing it.

testing/test-utils/src/lib/utils/test-folder-setup.ts
```diff
@@ -36,3 +37,88 @@ export async function teardownTestFolder(dirName: string) {
...
+}
```
matejchalk: The function is somewhat complex, so I think it would be worth unit testing it with `memfs`.

e2e/ci-e2e/mocks/setup.ts
BioPhoton: Will hardcode it.

testing/test-utils/src/lib/utils/test-folder-setup.ts
BioPhoton: Very good point!! I wanted to open a pr showing the error and suggested solution but did not manage to understand the issue yesterday. Expect comments today and a PR in the next days.

testing/test-utils/src/lib/utils/test-folder-setup.ts
```diff
@@ -36,3 +37,88 @@ export async function teardownTestFolder(dirName: string) {
...
+ * This is needed to restore files back to their prefixed state after tests,
+ * so they are excluded from Nx detection.
+ *
+ * @param dir - Directory to process recursively
+ * @param fileNames - Array of file names to prefix (e.g., ['package.json', 'nx.json', 'project.json'])
+ */
+export async function prefixRenamedFiles(
+  dir: string,
+  fileNames: readonly string[],
+): Promise<void> {
```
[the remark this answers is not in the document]
BioPhoton: Will remove it could be leftover of me trying to fix the last folder.

testing/test-utils/src/lib/utils/test-folder-setup.unit.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    );
+
+    await expect(restoreNxIgnoredFiles('/')).resolves.not.toThrow();
+
+    expect(vol.existsSync('/nx.json')).toBe(true);
+    expect(vol.existsSync('/_nx.json')).toBe(false);
+    expect(vol.existsSync('/package.json')).toBe(true);
+    expect(vol.existsSync('/_package.json')).toBe(false);
+    expect(vol.existsSync('/project.json')).toBe(true);
+    expect(vol.existsSync('/_project.json')).toBe(false);
```
matejchalk: How about using `vol.toJSON()` to list everything?

```suggestion
    expect(vol.toJSON()).toEqual({
      '/nx.json': '{}',
      '/package.json': '{}',
      '/project.json': '{}',
    });
```

matejchalk (commented):
I tried replicating the cache bug in `plugin-eslint` as described in #1102, and it now works as I would expect. 🎉 

I left some suggestions for improving the setup tests.

Don't forget to update the PR description, now that you've covered `plugin-eslint`, I believe this PR can close the issue without a follow-up PR (`closes #1102`).

PR #1144, opened 2025-11-13
fix(utils): add silent flag to executeProcess

hanna-skryl:
## Problem

The Axe plugin failed during its first test run in an external repository:

```
Running Axe accessibility checks for 1 URL...
  Testing URL 1/1: https://example.com
  Checking Chromium browser installation...
  Axe execution failed for https://example.com: Internal Logger error - concurrent spinners are not supported
```

**Root cause:** The logger only supports one spinner at a time. When plugins execute concurrently, the Axe plugin's browser installation attempts to start a second spinner while another plugin's spinner is active, triggering the error.

## Solution

Added a `silent` flag to `executeProcess` to bypass the logger spinner for background operations:

```typescript
await executeProcess({
  command: 'npx',
  args: ['playwright-core', 'install', 'chromium'],
  silent: true,
});
```

## Testing

1. Reproduced the error by running multiple Code PushUp plugins in the external repository
2. Applied the fix by modifying `executeProcess` source code
3. Verified the execution succeeded without the concurrent spinner error

### DIFF
--- a/packages/plugin-axe/src/lib/runner/run-axe.ts
+++ b/packages/plugin-axe/src/lib/runner/run-axe.ts
@@ -81,6 +81,7 @@ async function ensureBrowserInstalled(): Promise<void> {
   await executeProcess({
     command: 'npx',
     args: ['playwright-core', 'install', 'chromium'],
+    silent: true,
   });
 
   browserChecked = true;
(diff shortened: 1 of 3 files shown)


matejchalk:
Hi, you've run into a bug I discovered recently. **The plugins are supposed to be run sequentially.** I designed the logger groups and spinners based on this assumption, then I found out that it was (probably mistakenly) refactored to concurrent execution some time ago. :sweat_smile: 

I've opened a PR to prevent concurrent execution. This should enable the Axe plugin to log processes just like any other plugin. The problem isn't specific to Axe, any 2 plugins that execute some shell command would run into the same conflict.
- #1145

PR #1145, opened 2025-11-14
fix(core): run plugins sequentially

matejchalk:
This fix should prevent problems with concurrent spinners, such as in #1144. 

Plugins are intended to be executed sequentially, but it appears they were (unintentionally?) made concurrent in 7a592ebd268cb9b211da29e3473c4170f9ad4525 as part of the large PR #811.

I intend to make more changes to the plugin execution (log groups, etc.), but for now this is just a minimal fix.

### DIFF
--- a/packages/core/src/lib/implementation/execute-plugin.ts
+++ b/packages/core/src/lib/implementation/execute-plugin.ts
@@ -11,11 +11,13 @@ import {
 } from '@code-pushup/models';
 import {
   type ProgressBar,
+  asyncSequential,
   getProgressBar,
   groupByStatus,
   logMultipleResults,
   pluralizeToken,
   scoreAuditsWithTarget,
+  settlePromise,
   stringifyError,
 } from '@code-pushup/utils';
 import {
@@ -176,7 +178,7 @@ export async function executePlugins(
   );
 
   const errorsTransform = ({ reason }: PromiseRejectedResult) => String(reason);
-  const results = await Promise.allSettled(pluginsResult);
+  const results = await asyncSequential(pluginsResult, settlePromise);
 
   progressBar?.endProgress('Done running plugins');
 


hanna-skryl (changes requested):
I tested this change, but the concurrent spinner error still occurs:

1. Applied sequential execution in an external repository
2. Modified source code and verified changes took effect
3. Error still thrown during Axe plugin execution

The spinner conflict appears to be caused by something other than parallel plugin execution.

matejchalk:
> I tested this change, but the concurrent spinner error still occurs:
> 
> 1. Applied sequential execution in an external repository
> 2. Modified source code and verified changes took effect
> 3. Error still thrown during Axe plugin execution
> 
> The spinner conflict appears to be caused by something other than parallel plugin execution.

Weird, I'm not sure why that would be. Concurrent spinners could be caused by `Promise.all`/`Promise.allSettled` or spinner nesting. Running another spinner after the previous one resolved should work fine. I added some extra tests to the logger to check this, just in case (#1147).

The only spinner usage is in `executeProcess` calls at the moment, so I'm having a hard time figuring out how there can be more than 1 running. The only `executeProcess` I can find in `axe-plugin` is the `playwright install` command that runs for each URL, but in your async `reduce` you await the previous value, so I don't see any possible conflict there. :confused: 

If you'd like me to help debug the problem, I'll probably need to reproduce it. Do you think you could share some minimal reproduction?

The Axe plugin is working in this repo, so I'd start by trying to figure out what's different about the configuration in the repo where you're seeing the error (what other plugins are used, etc.).

How are you applying the changes from this PR, BTW? Editing the downloaded `.js` files in `node_modules/@code-pushup/core`? You could also try installing via `npm i https://pkg.pr.new/code-pushup/cli/@code-pushup/axe-plugin@1145`.

matejchalk:
Updating to `0.87.2` fixed the error for me in the portal. :tada: 

Screenshot from 2025-11-14 19-08-15

----

@hanna-skryl All good on your end?

PR #1150, opened 2025-11-20
fix(plugin-axe): resolve pnpm install issue

hanna-skryl:
Use Node's module resolution and npm bin specification to locate and execute `playwright-core` CLI directly, instead of relying on npx, which fails with pnpm's strict dependency isolation.

With pnpm, transitive dependency binaries are not hoisted to the root `node_modules/.bin` directory, causing `npx playwright-core` to fail with `playwright-core: command not found`.

The fix was verified with all major package managers by installing the preview package `https://pkg.pr.new/code-pushup/cli/@code-pushup/axe-plugin@1150` and testing browser installation with Axe accessibility audits:

- **pnpm** (local + GitHub Actions CI)
- **npm** (local)
- **yarn** (local)

### DIFF
--- a/packages/plugin-axe/src/lib/runner/run-axe.ts
+++ b/packages/plugin-axe/src/lib/runner/run-axe.ts
@@ -1,6 +1,6 @@
-import AxeBuilder from '@axe-core/playwright';
+import { AxeBuilder } from '@axe-core/playwright';
 import { createRequire } from 'node:module';
-import { dirname, join } from 'node:path';
+import path from 'node:path';
 import { type Browser, chromium } from 'playwright-core';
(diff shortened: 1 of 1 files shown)


packages/plugin-axe/src/lib/runner/run-axe.ts
```diff
@@ -1,4 +1,6 @@
 import AxeBuilder from '@axe-core/playwright';
+import { createRequire } from 'node:module';
+import { dirname, join } from 'node:path';
```
matejchalk: We use namespace imports `import path from 'node:path'`. I'm surprised this wasn't picked up by ESLint 😕

matejchalk (approved):

PR #1151, opened 2025-11-21
fix: resolve incomplete linting across projects

hanna-skryl:
Fixes incomplete linting configuration that prevented ESLint from checking most TypeScript files in the repository (revealed in [this discussion](https://github.com/code-pushup/cli/pull/1150#discussion_r2545375156)).

The `nx.json` lint target was missing quotes around the glob pattern.

Without quotes, the glob pattern was expanded before reaching ESLint. As a result, instead of matching all TypeScript files recursively, only the top-level files in each project were linted.

Once all files were properly linted, fixed the newly discovered errors across multiple projects.

### DIFF
--- a/packages/utils/src/lib/logger.int.test.ts
+++ b/packages/utils/src/lib/logger.int.test.ts
@@ -275,9 +275,7 @@ ${ansis.cyan('└')} ${ansis.green(`Total line coverage is ${ansis.bold('82%')}`
         .mockReturnValueOnce(0)
         .mockReturnValue(45); // 2nd group duration: 45 ms
       mathRandomSpy
-        // eslint-disable-next-line unicorn/number-literal-case
         .mockReturnValueOnce(0x1a / Math.pow(2, 8)) // 1st group's random ID: "1a"
-        // eslint-disable-next-line unicorn/number-literal-case
         .mockReturnValueOnce(0x1b / Math.pow(2, 8)); // 2nd group's random ID: "1b"
       const logger = new Logger();
 
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -116,6 +116,7 @@ export default tseslint.config(
     files: ['**/*.ts', '**/*.js'],
     rules: {
       'n/file-extension-in-import': ['error', 'always'],
+      'unicorn/number-literal-case': 'off',
     },
   },
   {


packages/utils/src/lib/logger.int.test.ts
```diff
@@ -267,14 +268,16 @@ ${ansis.cyan('└')} ${ansis.green(`Total line coverage is ${ansis.bold('82%')}`
...
       performanceNowSpy
         .mockReturnValueOnce(0)
         .mockReturnValueOnce(123) // 1st group duration: 123 ms
         .mockReturnValueOnce(0)
         .mockReturnValue(45); // 2nd group duration: 45 ms
       mathRandomSpy
+        // eslint-disable-next-line unicorn/number-literal-case
         .mockReturnValueOnce(0x1a / Math.pow(2, 8)) // 1st group's random ID: "1a"
+        // eslint-disable-next-line unicorn/number-literal-case
         .mockReturnValueOnce(0x1b / Math.pow(2, 8)); // 2nd group's random ID: "1b"
```
matejchalk: I'd autofix the number formatting. I don't really mind whether the hexadecimals are in uppercase or lowercase, so I'd stick to the convention.

```suggestion
        .mockReturnValueOnce(0x1A / Math.pow(2, 8)) // 1st group's random ID: "1a"
        .mockReturnValueOnce(0x1B / Math.pow(2, 8)); // 2nd group's random ID: "1b"
```

testing/test-setup-config/project.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "test-setup-config",
   "$schema": "../../node_modules/nx/schemas/project-schema.json",
-  "sourceRoot": "testing/test-setup/src",
+  "sourceRoot": "testing/test-setup-config/src",
```
matejchalk: Good catch 😅

packages/ci/vitest.int.config.ts
matejchalk: The `logger.mock.ts` setup file is already included, so this configuration is unnecessarily complicated. It's a leftover from previous PRs; the vitest configs and logger mocks refactors overlapped somewhat.

```suggestion
export default createIntTestConfig('ci');
```

packages/utils/src/lib/logger.int.test.ts
hanna-skryl: There is a conflict between Prettier and ESLint. Prettier formats hexadecimal literals to lowercase, whereas ESLint flags them as errors for not being uppercase. It seems one should be disabled.

PR #1152, opened 2025-11-24
feat(ci): improve code-pushup command logs

matejchalk:
After reviewing how our [CI logs](https://github.com/code-pushup/cli/actions/runs/19574682793/job/56056764992#step:6:30) look with the new Logger, I decided some further improvements are needed to make them readable.

The new way of logging executed processes nicely formats external tools, but when the CI package logs `code-pushup` commands this way, it introduces too much nesting. This also breaks log group rendering.

Furthermore, I realized a single `verbose` flag for everything isn't the best in this case. The CI doesn't log almost anything without enabling `CP_VERBOSE`, because most of the interesting output is from the `code-pushup` commands (e.g., report summary). However, verbose shouldn't be needed by default, as it introduces too much noise (e.g., all passed audits).

My solution is to disable the default process logging in the CI package, and use some new low-level utilities to print `code-pushup` output as soon as it's received (don't have to wait for everything to complete before seeing progress), with spaces via line breaks instead of indentation and gray color. By default, the output is printed, regardless of `CP_VERBOSE`. To silence the `code-pushup` command output specifically, I've introduced an optional `silent` flag. 

For all other `executeProcess` calls, the logs look the same as before.

### DIFF
--- a/.github/actions/code-pushup/action.yml
+++ b/.github/actions/code-pushup/action.yml
@@ -16,4 +16,3 @@ runs:
       env:
         TSX_TSCONFIG_PATH: .github/actions/code-pushup/tsconfig.json
         GH_TOKEN: ${{ inputs.token }}
-        CP_VERBOSE: true
--- a/.github/actions/code-pushup/src/runner.ts
+++ b/.github/actions/code-pushup/src/runner.ts
@@ -10,7 +10,11 @@ import {
   type SourceFileIssue,
   runInCI,
 } from '@code-pushup/ci';
-import { CODE_PUSHUP_UNICODE_LOGO, stringifyError } from '@code-pushup/utils';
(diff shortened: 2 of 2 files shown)


matejchalk:
> I wonder if there is a specific reason some JSON logs have remained non-collapsible.

I'm guessing you mean the [npm commands](https://github.com/code-pushup/cli/actions/runs/19699122405/job/56431153330?pr=1152#step:6:2282) with large JSON output? That's the raw process output, which is meant to be printed underneath the spinner with indentation, as you see. But it shouldn't be logged by default - only the command name, status, and duration are always logged. However, Michael hard-coded `--verbose` for all our `code-pushup` targets, which I'll be getting rid of in my next PR :sweat_smile:

Issue #1093, opened 2025-08-28
Local action to test `@code-pushup/ci` before release


matejchalk:
## Developer story

Most bugs or feature requests from the user-facing [`code-pushup/github-action`](https://github.com/code-pushup/github-action) or [`code-pushup/gitlab-pipelines-template`](https://gitlab.com/code-pushup/gitlab-pipelines-template) require changes to their shared [`@code-pushup/ci`](https://www.npmjs.com/package/@code-pushup/ci) package. When testing these changes, we're relying on simulated CI environments in integration and end-to-end tests. We can't fully test it in a real CI environment until a new `@code-pushup/ci` version is released and the GitHub Action or GitLab template update their dependencies.

If the `code-pushup/cli` repo would use the current `ci` package source code to run the CI flow on itself, we could catch mistakes at the PR stage, and be confident when releasing that it at least works end-to-end in our own GitHub Actions workflow. To achieve this, we'd need to wrap `@code-pushup/ci` in a local action and use that instead of the latest `code-pushup/github-action` release. The trade off is that some of the `code-pushup/github-action` source code would be duplicated, although we need not copy its whole feature set.

### Local action

GitHub Actions can use locally defined actions when given a relative path to a directory with an `action.yml`.

```diff
  - name: Code PushUp
-   uses: code-pushup/github-action@v0
+   uses: ./.github/actions/code-pushup
```

The `action.yml` may define a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) with arbitrary steps:

```yml
name: Code PushUp
description: Minimalist `code-pushup/github-action` which runs local `@code-pushup/ci` source code

runs:
  using: composite
  steps:
    - name: Run Node script
      run: npx tsx --tsconfig=tools/tsconfig.json tools/github-action.ts
      shell: bash
```

Or it may define a [JavaScript action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-javascript-action):

```yml
name: Code PushUp
description: Minimalist `code-pushup/github-action` which runs local `@code-pushup/ci` source code

runs:
  using: 'node20'
  main: src/runner.js
```

[...]

PR #1154, opened 2025-11-26
ci: run code-pushup on self in monorepo mode

matejchalk:
Related to #888 ... sort of. I want to be able to evaluate our CI logs in a representative environment, which requires completing what @BioPhoton started in #1062.

I split our Code PushUp workflow into 2 distinct jobs:

- **standalone** (see [logs](https://github.com/code-pushup/cli/actions/runs/19770375091/job/56653496086#step:6:119))
  - basic "all in one" setup
  - single config file at repository root
  - no Nx task dependencies or plugin caching
  - includes all plugins
- **monorepo** (see [logs](https://github.com/code-pushup/cli/actions/runs/19770375091/job/56653496280#step:6:39))
  - enterprise-level "mega Nx" setup
  - heavily reliant on Nx task graph and plugin caching
  - each package has its own config
  - skips `print-config` commands by providing `configPatterns`
  - only includes plugins that make sense on project-level
  
Each job posts separate PR comments (I added a special `jobId` to `@code-pushup/ci` for this use case).

I refactored the `code-pushup.preset.ts` file to minimize duplication in `code-pushup.config.ts` files.

### DIFF
--- a/packages/cli/src/lib/implementation/core-config.int.test.ts
+++ b/packages/cli/src/lib/implementation/core-config.int.test.ts
@@ -18,7 +18,7 @@ vi.mock('@code-pushup/core', async () => {
   return {
     ...(core as object),
     readRcByPath: vi.fn().mockImplementation((filepath: string): CoreConfig => {
-      const allPersistOptions = {
+      const allPersistOptions: CoreConfig = {
         ...CORE_CONFIG_MOCK,
         persist: {
           filename: 'rc-filename',
@@ -26,7 +26,7 @@ vi.mock('@code-pushup/core', async () => {
           outputDir: 'rc-outputDir',
         },
       };
-      const persistOnlyFilename = {
+      const persistOnlyFilename: CoreConfig = {
         ...CORE_CONFIG_MOCK,
         persist: {
           filename: 'rc-filename',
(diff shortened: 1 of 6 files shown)


.github/actions/code-pushup/src/runner.ts
```diff
@@ -134,9 +135,32 @@ async function run(): Promise<void> {
...
-    const options: Options = {
-      bin: 'npx nx code-pushup --nx-bail --',
-    };
+    const isMonorepo = core.getBooleanInput('monorepo');
+
+    const options: Options = isMonorepo
+      ? {
+          jobId: 'monorepo-mode',
+          monorepo: 'nx',
+          nxProjectsFilter: '--with-target=code-pushup --exclude=workspace',
```
BioPhoton: Users could have custom target names. Or Nx-Plugin for example supports this with better DX over the [plugin options](https://github.com/code-pushup/cli/blob/main/packages/nx-plugin/src/plugin/README.md#options). Would it be possible to provide this an option with a default value?

BioPhoton (changes requested):
Nice dogfooding! I had a question for custom target names.

.github/actions/code-pushup/src/runner.ts
matejchalk: We already support custom target names via the [`task` option](https://github.com/code-pushup/cli/tree/main/packages/ci#options:~:text=task,project%20in%20monorepo%20mode). But that's not related to this piece of code.

Unlike `@code-pushup/ci`, the local action isn't user-facing. Its sole purpose is to run `code-pushup` on _this and only this repo_ from source. So, for simplicity, we provide the options directly. For the full context, see #1093.

PR #1157, opened 2025-12-01
refactor: use utils in nx-plugin

BioPhoton:
related to #1091 

**This PR includes:**
- alignments of terminal args helper
- add command formatter helper to `utils`
- reuse utils in `nx-plugin` and remove old code
- remove bin logic from config target (left over of #1109)

### DIFF
--- a/packages/utils/src/lib/command.unit.test.ts
+++ b/packages/utils/src/lib/command.unit.test.ts
@@ -1,7 +1,7 @@
 import ansis from 'ansis';
 import path from 'node:path';
 import process from 'node:process';
-import { describe, expect, it, vi } from 'vitest';
+import { describe, expect, it } from 'vitest';
 import { formatCommandStatus } from './command.js';
 
 describe('formatCommandStatus', () => {


packages/utils/src/lib/logger.ts
```diff
@@ -532,3 +532,42 @@ export class Logger {
...
+
+/**
+ * Formats a command string for display with status indicator.
+ *
+ * @param bin Command string with arguments.
+ * @param options Command options (cwd, env).
+ * @param status Command status ('pending' | 'success' | 'failure').
+ * @returns Formatted command string with colored status indicator.
+ */
+export function formatCommand(
```
matejchalk: How about `formatCommandStatus`? The name `formatCommand` is a little ambiguous. Initially, I thought it was the `[command, ...args ?? []].join(' ')` logic.

packages/utils/src/lib/command.unit.test.ts
```diff
@@ -1,19 +1,13 @@
 import ansis from 'ansis';
 import path from 'node:path';
-import { afterEach, describe, expect, it, vi } from 'vitest';
-import { formatCommand } from './logger.js';
+import process from 'node:process';
+import { describe, expect, it, vi } from 'vitest';
+import { formatCommandStatus } from './command.js';
 
 describe('formatCommand', () => {
```
matejchalk:
```suggestion
describe('formatCommandStatus', () => {
```

packages/utils/src/lib/logger.unit.test.ts
```diff
@@ -32,29 +26,20 @@ describe('formatCommand', () => {
...
-    const mockCwd = path.join(originalCwd, 'src');
-    vi.spyOn(path, 'relative').mockReturnValue('src');
-
-    expect(formatCommand('npx -v', { cwd: mockCwd })).toStartWith(
-      `${ansis.blue('src')} `,
-    );
+    expect(
+      formatCommandStatus('npx eslint . --format=json', {}, status),
+    ).toContain(`${color('$')}`);
   });
```
matejchalk: Why did you delete this test case? It only requires a small adjustment to run without mocking `path.relative`. Without it, there's no coverage for the CWD prefix.

```ts
  it('should include cwd prefix when cwd is provided and different from process.cwd()', () => {
    expect(
      formatCommandStatus('npx -v', { cwd: path.join(process.cwd(), 'src') }),
    ).toStartWith(`${ansis.blue('src')} `);
  });
```

BioPhoton: reverted

PR #1159, opened 2025-12-02
refactor(nx-plugin): set env vars if verbose is given

BioPhoton:
This PR transforms the terminal args `--verbose` into a env var `CP_VERBOSE=true`

### DIFF
--- a/packages/nx-plugin/src/executors/cli/executor.int.test.ts
+++ b/packages/nx-plugin/src/executors/cli/executor.int.test.ts
@@ -50,10 +50,6 @@ describe('runAutorunExecutor', () => {
       cwd: process.cwd(),
     });
 
-    expect(process.env).toStrictEqual(
-      expect.objectContaining({
-        CP_VERBOSE: 'true',
-      }),
-    );
+    expect(process.env).toHaveProperty('CP_VERBOSE', 'true');
   });
 });


packages/nx-plugin/src/executors/cli/executor.ts
```diff
@@ -15,44 +15,49 @@ export default async function runAutorunExecutor(
...
         args,
         ...(context.cwd ? { cwd: context.cwd } : {}),
+        ...(verbose
+          ? {
+              env: {
+                ...process.env,
+                ...executorEnvVariables,
+              },
+            }
+          : {}),
```
matejchalk: This logic is a little brittle. What if other properties were later added to `executorEnvVariables` in addition to `verbose`? The condition would have to be rewritten.

Something like this would be more future-proof:

```suggestion
        ...(Object.keys(executorEnvVariables).length > 0
          ? {
              env: {
                ...process.env,
                ...executorEnvVariables,
              },
            }
          : {}),
```

Although, since the `env` option uses `process.env` by default, I believe this simplified version would be equivalent:


```suggestion
        env: { ...process.env, ...executorEnvVariables },
```

packages/nx-plugin/src/executors/cli/executor.ts
```diff
@@ -15,44 +15,49 @@ export default async function runAutorunExecutor(
...
-  if (verbose) {
-    logger.info(`Run CLI executor ${command ?? ''}`);
-    logger.info(`Command: ${commandString}`);
-  }
+
   if (dryRun) {
     logger.warn(`DryRun execution of: ${commandString}`);
   } else {
     try {
+      logger.debug(`With env vars: ${executorEnvVariables}`);
```
matejchalk: The way you're setting `CP_VERBOSE` will only apply to the inner command, but not logs from the executor, like the `logger.debug` statement above.

A simpler alternative would be to call `logger.setVerbose(verbose)`. This method sets `process.env['CP_VERBOSE']`, meaning the verbose setting will apply to the current process as well as child processes. Then you wouldn't need to pass the `env` option to the child process.

BioPhoton: THX

packages/nx-plugin/src/executors/cli/executor.int.test.ts
```diff
@@ -27,6 +27,11 @@ describe('runAutorunExecutor', () => {
   });
 
   it('should normalize context, parse CLI options and execute command', async () => {
+    expect(process.env).toStrictEqual(
+      expect.not.objectContaining({
+        CP_VERBOSE: 'true',
+      }),
+    );
```
matejchalk: The [`toHaveProperty`](https://vitest.dev/api/expect.html#tohaveproperty) matcher would be a good fit here:

```suggestion
    expect(process.env).not.toHaveProperty('CP_VERBOSE', 'true');
```

matejchalk (approved):

PR #1165, opened 2025-12-03
refactor: fix lighthouse text styling

BioPhoton:
This PR fixes a bug that slipped into master.

See: https://github.com/code-pushup/cli/blob/f20b3b879cd3b12727332d8a11a6e0fd05586748/packages/utils/src/lib/text-formats/ascii/link.ts#L4

In the jobs only the main branch fails. https://github.com/code-pushup/cli/actions/runs/19903612406/job/57053902389?pr=1165


matejchalk (approved):
I noticed this as well, opened #1166, then I found that you were already fixing it. 😁 Thanks 👍

PR #1183, opened 2025-12-10
fix: ensure CLI bin directory is included in published package with executable permissions

beaussan:
## Summary

- Fix CLI package publishing by ensuring the `bin` directory is properly included in the npm package
- Update build configuration to preserve executable permissions for `bin/index.js`
- Move bin entry point reference from `index.js` to `./bin/index.js` in package.json

This fix addresses an issue where the CLI binary wasn't being included in the published npm package because the `files` field in package.json only included the `src` directory. Without this fix, users installing the package via npm wouldn't get the executable bin file.

## Impact

- **Users**: Fixes installation issues for CLI users - the `code-pushup` command will now work correctly after npm install
- **Publishing**: Ensures bin directory is included in published package with correct permissions
- **Build System**: Updated Nx build configuration to explicitly include bin directory in assets
- **Risk**: Very low - this is a build/packaging fix with no runtime code changes
- **Breaking Changes**: None - this is a bug fix that makes the package work as intended

## Technical Details

The fix involves:
1. Adding `bin` directory to the build assets in `project.json` to ensure it's copied during build
2. Updating the bin path in `package.json` from `index.js` to `./bin/index.js` to match the actual file location
3. The empty `bin/index.js` file is tracked in git with executable permissions (+x) to ensure they're preserved through version control and build processes

This follows a similar pattern to a previous fix (commit 0fd2eb41) that addressed executable permissions for the old location.

### DIFF
--- a/packages/cli/bin/index.js
+++ b/packages/cli/bin/index.js
@@ -1,2 +1,11 @@
-#! /usr/bin/env node
+#!/usr/bin/env node
+
+/**
+ * This file serves as the CLI entry point.
+ *
+ * We use a separate bin file (instead of pointing directly to src/index.js)
+ * because TypeScript build processes don't preserve file permissions.
+ * By tracking this file in git with executable permissions (+x), we ensure
+ * the CLI remains executable after npm publish without needing post-install scripts.
+ */
 import '../src/index.js';


BioPhoton (approved):
Thx for the update! Looks good to me! 
Maybe a comment in code to explain the why?

PR #1162, opened 2025-12-02
chore: add zod2md-nx-plugin

BioPhoton:
Followup of #1163
Precondition of #1188 

This PR includes:
- create nx plugin under `zod2md-jsdocs`
- logic to add target config if file matches `zod2md.config.ts`
- plugin docs
- remove old config
- remove file reference from production inputs as changes in the config may impact build output

A potential followup PR could implement a nx plugin for our `build` target and add `dependsOn` settings automatically. This is also noted in a clean up issue.

> [!note]
> This is tooling code and it is not planned to get released

Related:
#1168

PR #1188, opened 2025-12-15
refactor: add sync generator for zod2md tooling setup

BioPhoton:
preconditions:
- #1162 
- #1242

This PR implements:
- handle `zod2md` dependencies if `zod2md.config.js` is present

---
## Usage

```bash
nx sync --check # log status
nx sync # sync projets
```

## Example project structure

```txt
Root/
├── libs/
│   └── project-name/
│       ├── zod2md.config.ts        🔎 triggers the sync generator
│       ├── project.json            👈 checks targets + build.dependsOn
│       ├── tsconfig.lib.json       👈 configures zod2md-jsdocs TS plugin
│       └── src/
│           └── index.ts
└── ...
```

---

## Targeted behavior summary

- ✔ No `zod2md.config.ts` → project is ignored
- ✔ Missing TS plugin config → patched automatically
- ✔ Missing Zod2Md target → added automatically
- ✔ Missing `build.dependsOn` entries → updated automatically
- ✔ Fully configured project → no changes, no errors

---

### DIFF
--- a/.env.local
+++ b/.env.local
@@ -1,5 +1,5 @@
 # Local development environment variables
-# Copy this to .env.local (gitignored) and customize for your setup
+# Copy this to .env.local and customize for your setup
 
 # Enable tsx for TypeScript execution in Nx
 # This allows running local generator/executor directly as .ts files without pre-compilation
--- a/packages/plugin-typescript/mocks/fixtures/nx-monorepo/_nx.json
+++ b/packages/plugin-typescript/mocks/fixtures/nx-monorepo/_nx.json
@@ -1,1 +1,8 @@
-{}
+{
+  "$schema": "./node_modules/nx/schemas/nx-schema.json",
+  "targetDefaults": {},
+  "workspaceLayout": {
+    "appsDir": "apps",
+    "libsDir": "packages"
+  }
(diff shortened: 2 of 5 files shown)


BioPhoton:
@matejchalk to be able to use uncompiled Nx tooling code we can't have `.js` extensions present. The reason is, Nx uses SWC under the hood and can't be configured.

Here we see the existing plugin does not have any import to local files: https://github.com/code-pushup/cli/blob/main/tools/zod2md-jsdocs/src/nx-plugin.ts otherwise it fails. 

Any other ideas? 💡 how should we progress here other than avoiding extensions?

matejchalk:
> ... no way to tell Nx to look up a matching `.ts` file for `.js` imports ...

:bulb: Actually, I think I've found a way. We can set a custom Node loader using [environment variables](https://nx.dev/docs/guides/tips-n-tricks/define-environment-variables). I've tested it locally by adding a `constants.ts` that's imported as `./constants.js`, then running `nx graph`. This gave me a `Cannot find module './constants.js'` error, until the `.env.local`, which fixed it. (See #1238) :slightly_smiling_face:

PR #1206, opened 2026-01-07
feat(utils): add performance observer

BioPhoton:
This PR implements:
- performance observer util
- type helper for data flow

Types are used in followup PR #1210

### DIFF
--- a/testing/test-utils/src/lib/utils/perf-hooks.mock.ts
+++ b/testing/test-utils/src/lib/utils/perf-hooks.mock.ts
@@ -1,8 +1,6 @@
-import type { PerformanceEntry } from 'node:perf_hooks';
+import type { EntryType, PerformanceEntry } from 'node:perf_hooks';
 import { vi } from 'vitest';
 import { MockPerformanceObserver } from './performance-observer.mock';
-
-type EntryType = 'mark' | 'measure';
 
 let nowMs = 0;
 let entries: PerformanceEntry[] = [];


testing/test-utils/src/lib/utils/performance-observer.mock.ts
```diff
@@ -0,0 +1,88 @@
...
+    return this.instances.at(-1);
+  }
+
+  buffered = false;
+  private observing = false;
+  bufferedEntries: PerformanceEntry[] = [];
+
+  constructor(cb: PerformanceObserverCallback) {
+    MockPerformanceObserver.instances.push(this);
+  }
```
hanna-skryl: The constructor accepts a callback to match the real `PerformanceObserver` signature, but it's never stored or used. Is this just for API compatibility, or should the mock be invoking it somewhere?

BioPhoton: revamped all mocking

testing/test-utils/src/lib/utils/perf-hooks.mock.ts
```diff
@@ -0,0 +1,75 @@
+import type { PerformanceEntry } from 'node:perf_hooks';
+import { vi } from 'vitest';
+import { MockPerformanceObserver } from './performance-observer.mock';
+
+type EntryType = 'mark' | 'measure';
```
hanna-skryl: Why declare `EntryType` locally instead of importing it from `node:perf_hooks`?

packages/utils/src/lib/sink-source.types.ts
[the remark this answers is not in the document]
BioPhoton: Types are used in followup PR https://github.com/code-pushup/cli/pull/1210

Issue #1197, opened 2025-12-23
Implement performance profiling and trace file generation


BioPhoton:
The goal is to enable Browser functionality in the NodeJS environment, specifically the Extensibility API used for Performance profiling.

_Example - NodeJS code:_

```ts
// Mark used to represent the start of the image processing task
// The start time is defaulted to now
const imageProcessinTimeStart = performance.now();

// ... later in your code

// Track entry representing the completion of image processing
// with additional details and a tooltip
// The start time is a marker from earlier
// The end time is defaulted to now
performance.measure('Image Processing Complete', {
  start: imageProcessinTimeStart,
  detail: {
    devtools: {
      dataType: 'track-entry',
      track: 'Image Processing Tasks',
      trackGroup: 'My Tracks', // Group related tracks together
      color: 'tertiary-dark',
      properties: [
        ['Filter Type', 'Gaussian Blur'],
        ['Resize Dimensions', '500x300'],
      ],
      tooltipText: 'Image processed successfully',
    },
  },
});
```

_Example - ChromeDevTools Performance panel:_

<img width="1015" height="441" alt="Image" src="https://github.com/user-attachments/assets/a22f5c87-2408-4402-84d0-d5e230023478" />

## Problem

ATM NodeJS `--cpu-pro` or similar flags do not support the Extensibility API used by Chrome DevTools for Performance profiling.
Nevertheless, calling the API with additional details does not cause any issues or logs.

## Solution

The solution requires multiple building blocks elaborated below:
- Utils
  - Persist trace data
  - PerformanceEntry to Trace Event
  - PerformanceEntry Details helper
- Profiler
  - API
  - Usage

### Persist trace data

Trace persistence is implemented in 2 steps:
- at runtime add trace events to an in-memory structure
- at the end of the profiling session, serialize the trace events to a file or the terminal.

### PerformanceEntry to Trace Event helper

A helper function can be created to convert `PerformanceEntry` objects into Trace Events that can be consumed by Chrome DevTools.

```ts
const markA: PerformanceMeasureOptions = performance.mark('a-start');
const measureA = performance.measure('Run A', markA);
const traceEvents = [
  convertPerformanceMarkEntryToTraceEvent(markA),
  convertPerformanceMarkEntryToTraceEvent(measureA)
];
```

### PerformanceEntry Details helper

```ts
const Devtools = createDevtoolsSpans({
  analysis: { track: 'Analysis', group: 'Tools', color: 'primary' },
[...]

PR #1212, opened 2026-01-12
feat: add process exit handler

BioPhoton:
related to #1197 

This PR includes:
- adjustments for performance observer buffering
- add exit-process utils

### DIFF
--- a/packages/utils/src/lib/performance-observer.ts
+++ b/packages/utils/src/lib/performance-observer.ts
@@ -51,7 +51,7 @@ export class PerformanceObserverSink<T>
       return;
     }
 
-    // The only used to trigger the flush it is not processing the entries just counting them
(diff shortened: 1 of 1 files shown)


packages/utils/src/lib/exit-process.ts
```diff
@@ -0,0 +1,43 @@
...
+  SIGNALS.forEach(([signal]) => {
+    process.on(signal, () => {
+      options.onClose?.();
+    });
+  });
+
+  // Normal exit
+  process.on('exit', () => {
+    options.onClose?.();
+  });
```
hanna-skryl: If a signal is received, `onClose` is called. Then when the process exits, the `'exit'` event also triggers `onClose`. Should there be a guard to prevent duplicate calls?

BioPhoton:
I will add int tests now that i examined the logger code.☝️

PR #1213, opened 2026-01-12
feat(utils): performance TS ext

BioPhoton:
Related to #1197 

This PR includes:
- `perf-hooks.d.ts` types to extend performance APIs with Chrome DevTools Extensibility API types so we get autocomplete and type checks when using e.g. `performance.mark('mark-name', { details: /*this is typed*/})`


BioPhoton:
Ill drop this feature in favour of our helpers that are fully compatible with all native APIs.

PR #1222, opened 2026-01-19
fix(utils): prevent race condition in parallel plugin runs

beaussan:
## Summary

Fixes a race condition when running the same plugin for multiple projects in parallel (e.g., via Nx targets).

## Problem

When two processes started within the same millisecond, the timestamp-based directory naming in `createRunnerFiles` would collide:

```
node_modules/.code-pushup/{pluginSlug}/{timestamp}/
```

This caused one process to delete the output directory (in `executeRunnerConfig`) while another was still using it, leading to failures.

## Solution

Added `process.pid` and a random suffix to the directory name to ensure uniqueness:

```typescript
const uniqueId = `${Date.now()}-${process.pid}-${Math.random().toString(36).slice(2, 8)}`;
```

This ensures that even when multiple processes run simultaneously, each gets its own unique working directory.

### DIFF
--- a/packages/utils/src/lib/create-runner-files.ts
+++ b/packages/utils/src/lib/create-runner-files.ts
@@ -1,5 +1,6 @@
 import { writeFile } from 'node:fs/promises';
 import path from 'node:path';
+import { threadId } from 'node:worker_threads';
 import type { RunnerFilesPaths } from '@code-pushup/models';
 import { ensureDirectoryExists, pluginWorkDir } from './file-system.js';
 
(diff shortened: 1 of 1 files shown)


packages/utils/src/lib/create-runner-files.ts
```diff
@@ -13,8 +13,11 @@ export async function createRunnerFiles(
   pluginSlug: string,
   configJSON: string,
 ): Promise<RunnerFilesPaths> {
-  const timestamp = Date.now().toString();
-  const runnerWorkDir = path.join(pluginWorkDir(pluginSlug), timestamp);
+  // Use timestamp + process ID + random suffix to ensure uniqueness
+  // This prevents race conditions when running the same plugin for multiple projects in parallel
+  // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+  const uniqueId = `${Date.now()}-${process.pid}-${Math.random().toString(36).slice(2, 8)}`;
```
BioPhoton:
```suggestion
  const uniqueId = `${(performance.timeOrigin+performance.now())*10}-${process.pid}-${threadId}`;
```

- We can get an additional digit from `perf_hooks` which is also monotonic: `(performance.timeOrigin+performance.now())*10`
- Processes can have multiple threads `threadId` is 0 by default

BioPhoton: Let's drop Random numbers. The above is easier to test. We already have PID etc mocked. Also, it's kinda common and the parts of profiler are also inc tid.

PR #1248, opened 2026-02-17
feat(create-cli): setup wizard architecture

### LINKED ISSUE #1240: Architect setup wizard
## Developer story

To build a feature-rich setup wizard described in #1239 and its sub-issues, we need to pick the right tools and architecture. The goal of this task is to research and design how we'll build this tool.

## Our `create-cli` and `nx-plugin` packages

Technically, we do already have a config generator in `@code-pushup/create-cli` and `@code-pushup/nx-plugin`. However, all it does is generate an empty (and therefore invalid) config file, and it's not actually used anywhere. Furthermore, we can't even build on top of it, as it's tightly coupled to Nx - in fact, `@code-pushup/create-cli` executes the `@code-pushup/nx-plugin:configuration` generator. Basing our setup wizard on an Nx generator would severely limit its capabilities. For example, Nx generators only support static prompts, whereas we'll need dynamic prompts that depend on previous prompt answers or information from the user's file system.

We'll need to break `create-cli`'s dependency on `nx-plugin`. Instead of `create-cli` being a thin wrapper around a generator in `nx-plugin`, our core implementation should be entirely Nx-agnostic, and the Nx generator should be a thin wrapper on top of it.

hanna-skryl:
Closes #1240 

Break `create-cli`'s dependency on `nx-plugin` and rebuild the package around a modular setup wizard. Uses a custom class for config generation, `@inquirer/prompts` for interactive input, and a virtual `Tree` for buffered file writes with dry-run support.

### DIFF
--- a/packages/create-cli/src/lib/setup/virtual-fs.unit.test.ts
+++ b/packages/create-cli/src/lib/setup/virtual-fs.unit.test.ts
@@ -4,10 +4,10 @@ import { createTree } from './virtual-fs.js';
 
 function createMockFs(
   files: Record<string, string> = {},
-): FileSystemAdapter & { written: Map<string, string>; dirs: string[] } {
+): FileSystemAdapter & { written: Map<string, string>; dirs: Set<string> } {
   const store = new Map(Object.entries(files));
   const written = new Map<string, string>();
-  const dirs: string[] = [];
+  const dirs = new Set<string>();
 
   return {
     written,
@@ -26,9 +26,8 @@ function createMockFs(
     async exists(path: string) {
       return store.has(toUnixPath(path));
     },
-    async mkdir(_path: string, _options: { recursive: boolean }) {
-      // eslint-disable-next-line functional/immutable-data
-      dirs.push(toUnixPath(_path));
+    async mkdir(path: string): Promise<undefined> {
+      dirs.add(toUnixPath(path));
     },
   };
 }


packages/create-cli/src/lib/setup/types.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export type PluginSetupBinding = {
+  slug: PluginMeta['slug'];
+  title: PluginMeta['title'];
+  packageName: NonNullable<PluginMeta['packageName']>;
+  // TODO: #1244 — add async pre-selection callback (e.g. detect eslint.config.js in repo)
+  prompts?: PluginPromptDescriptor[];
+  codegenConfig: (
+    answers: Record<string, string | string[]>,
+  ) => PluginCodegenResult;
```
matejchalk: The name `codegenConfig` doesn't sound like a method; it's missing a proper verb. How about `generateConfig`?

packages/create-cli/src/lib/setup/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  type: 'select';
+  choices: PromptChoice[];
+  default: string;
+};
+
+type CheckboxPrompt = PromptBase & {
+  type: 'checkbox';
+  choices: PromptChoice[];
+  default: string[];
+};
```
matejchalk: I may be overthinking things at this point, but I suspect it could be useful to support type-safe string unions (e.g., `SelectPrompt<'standalone' | 'monorepo'>`).

```suggestion
type PromptChoice<T extends string> = { name: string; value: T };

type InputPrompt = PromptBase & {
  type: 'input';
  default: string;
};

type SelectPrompt<T extends string = string> = PromptBase & {
  type: 'select';
  choices: PromptChoice<T>[];
  default: T;
};

type CheckboxPrompt<T extends string = string> = PromptBase & {
  type: 'checkbox';
  choices: PromptChoice<T>[];
  default: T[];
};
```

BioPhoton:
As a nit pick question I want to understand if we could choose another library that ts-morph to achieve the same goal.

The library is old and slow and I thought nowadays must be a more performant lib on the market. 

What I can say about ts-morph is, 
ts-morph adds:
	•	heavy TS compiler dependency
	•	performance cost
	•	complexity
	•	unnecessary type-checking


Instead we could write a small typed builder:

```ts
generateConfig({
  imports: [...],
  plugins: [...],
  categories: [...]
})
```

Or use @swc/core AST builder:
	•	Extremely fast
	•	No type-checker
	•	Pure syntax
	•	Lightweight

packages/create-cli/src/lib/setup/wizard.ts
```diff
@@ -0,0 +1,70 @@
...
+  });
+}
+
+function logNextSteps(steps: [string, string][]): void {
+  const colWidth = Math.max(...steps.map(([label]) => label.length));
+  logger.info('Next steps:');
+  steps.forEach(([label, description]) => {
+    logger.info(`  ${label.padEnd(colWidth + COLUMN_GAP)}${description}`);
+  });
+}
```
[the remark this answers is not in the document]
hanna-skryl: Thanks for the pointer!

PR #1250, opened 2026-02-24
feat(create-cli): support config file format selection

### LINKED ISSUE #1243: Generate config file in setup wizard
## User story

Generating `code-pushup.config.{js,mjs,ts}` file is a basic requirement of the setup wizard.

### JavaScript (ESM)

```js
/** @type {import('@code-pushup/models').CoreConfig} */
export default {
  plugins: [
    // ...
  ],
};
```

### TypeScript

```ts
import type { CoreConfig } from '@code-pushup/models';

const config: CoreConfig = {
  plugins: [
    // ...
  ],
};

export default config;
```

## Acceptance criteria

- [ ] The user can select whether to use JavaScript or TypeScript for their config file(s).
- [ ] The default option is inferred from the file system. If a `tsconfig.json` or `tsconfig.*.json` is present in the root directory, then TypeScript is the default, otherwise it's JavaScript.
- [ ] If JavaScript is selected, then the `.js` extension should only be used if the root `package.json` (if any) contains `"type": "module"`. Otherwise, the `.mjs` extension should be used.

hanna-skryl:
Closes #1243 

Config file format selection is added to the setup wizard. The wizard now prompts users to choose between TypeScript and JavaScript, with the default auto-detected from the filesystem based on `tsconfig` presence. A `--config-format` CLI option allows skipping the prompt.

### DIFF
--- a/packages/create-cli/src/lib/setup/config-format.ts
+++ b/packages/create-cli/src/lib/setup/config-format.ts
@@ -3,7 +3,11 @@ import { readdir } from 'node:fs/promises';
 import path from 'node:path';
 import type { PackageJson } from 'type-fest';
 import { fileExists, readJsonFile } from '@code-pushup/utils';
-import type { CliArgs, ConfigFileFormat } from './types.js';
+import {
+  CONFIG_FILE_FORMATS,
+  type CliArgs,
+  type ConfigFileFormat,
+} from './types.js';
 
 const TSCONFIG_PATTERN = /^tsconfig(\..+)?\.json$/;
 
@@ -61,5 +65,6 @@ async function detectDefaultFormat(targetDir: string): Promise<DefaultFormat> {
 function isConfigFileFormat(
   value: string | undefined,
 ): value is ConfigFileFormat {
-  return value === 'ts' || value === 'js' || value === 'mjs';
+  const validValues: readonly string[] = CONFIG_FILE_FORMATS;
+  return value != null && validValues.includes(value);
 }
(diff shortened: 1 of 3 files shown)


packages/create-cli/src/lib/setup/config-format.ts
```diff
@@ -0,0 +1,65 @@
...
+
+async function detectDefaultFormat(targetDir: string): Promise<DefaultFormat> {
+  const files = await readdir(targetDir);
+  return files.some(file => TSCONFIG_PATTERN.test(file)) ? 'ts' : 'js';
+}
+
+function isConfigFileFormat(
+  value: string | undefined,
+): value is ConfigFileFormat {
+  return value === 'ts' || value === 'js' || value === 'mjs';
```
matejchalk: When you need a union with runtime validation, it's usually better to have a runtime constant as the source of truth. That allows you to derive both the type and the runtime validation from the same set of values. Otherwise, you have to keep them in sync without any help from TypeScript.

- source of truth:

  ```ts
  export const CONFIG_FILE_FORMATS = ['ts', 'js', 'mjs'] as const;
  ```

- derived type:

  ```ts
  export type ConfigFileFormat = (typeof CONFIG_FILE_FORMATS)[number];
  ```

- derived runtime validation:

  ```suggestion
    const validValues: readonly string[] = CONFIG_FILE_FORMATS;
    return value != null && validValues.includes(value);
  ```

PR #1283, opened 2026-05-18
refactor: extract wizard infrastructure into utils

hanna-skryl:
`create-cli` will soon integrate `create-eslint-config`'s wizard. Without a shared tree, each would buffer writes separately and we'd maintain two implementations. Sharing one gives us a single buffer and one canonical implementation.

The shared module uses `create-eslint-config`'s implementation rather than `create-cli`'s, because it already handles flush rollback, skips writes when content matches disk, and avoids re-recording identical writes, which `create-cli` was missing.

Additionally, this PR consolidates plugin binding inputs into a single object and extracts the yargs setup, fixing a latent bug where plugin-namespaced CLI flags silently no-opped instead of overriding their prompt defaults.


packages/utils/src/lib/wizard/virtual-fs.unit.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    await tree.write('new.ts', 'content');
+
+    await expect(tree.exists('new.ts')).resolves.toBeTrue();
+  });
+
+  it('should return null from read() for non-existent files', async () => {
+    await expect(
+      createTree('/project', createMockFs()).read('missing.ts'),
+    ).resolves.toBeNull();
+  });
```
matejchalk: How about grouping individual method tests under nested `describe`s? 🤔 

```
describe('createTree', () => {
  describe('exists', () => {
    // ...
  });

  describe('read', () => {
    // ...
  });

  describe('write', () => {
    // ...
  });

  describe('flush', () => {
    // ...
  });
});
```

matejchalk (approved):

matejchalk (approved):

